PR #140, opened 2022-09-30
Electronise

belluzj:
Fixes #100 

@justvanrossum This is a work-in-progress, I'm opening a PR just to signal that my colleague @smugpie is on the case. Also I see there's been work on main since we started this so we'll look into rebasing/merging.

The main point of having an Electron app is that users won't need to first open a terminal, then launch a browser, instead they'll have a regular desktop app with an icon to click on and "that's it". Then, Electron can provide more features in terms of integration with the desktop machine, auto-updating the app, providing native menus, file dialogs, notifications, etc. We can discuss in more details which of these are interesting to have for Fontra.

So far we've produced a Windows installer ([link to Google Drive .exe file](https://drive.google.com/file/d/1iPjXAqw0MgcLUVmnpvMK-sKxpz6oDNaF/view?usp=sharing)) that works as follows:
* as a requirement, you need to have Python 3.10 installed on your system for all users, with the python executable in the PATH. (we're not bundling Python with the Electron app)
* the first time you launch the Fontra app, it will create a Python 3.10 venv in a hidden user folder, install a Fontra Python wheel (bundled with the Electron app) into that venv, also pulling the dependencies of Fontra from pypi
* then it will spawn a Fontra process in the background, and show you the usual UI in the Fontra app window
  * on subsequent runs the venv will be already there so it will jump straight to showing the UI.
* From the Fontra app window you can do File > Open Folder or Ctrl+O to open a folder

This way of installing/launching the app is also up for discussion in case people have ideas for better ways of achieving a similar or better level of user-friendliness.

image

Next week we'll look into making a mac DMG or equivalent, with the same behaviour of making a venv somewhere and setting up Fontra's python side there.

Finally there's also something to discuss in terms of how the HTML/Javascript UI is split from the Python server code. There are several possible approaches for which we made a diagram below, and here we implement option 2. ([see diagram online](https://excalidraw.com/#json=LHX9xc2BQA619B9Aw-Eo1,t-V54nd4geoyKbEW3f5alg))

FontraElectronOptions


justvanrossum:
Ok, first of all: wow! Thank you for this!

Would it be an option to have the electronisation done as a separate project/repo, that imports the Fontra codebase as it is?

Our goal isn't for Fontra to *only* become an Electron app, it will be just one of the ways to deply it. So the more we can separate this the better. Of course, if changes to the main code base are needed, we can discuss, but I'd first like to better understand how this works and what is needed.

belluzj:
> So the more we can separate this the better. 

Yes, certainly. Here are our thoughts in preparation for more discussion tomorrow:

* Option 1A: iframe fontra as it is
  * Least invasive for fontra: can stay exactly the same
  * Small benefit from Electron app: launches fontra automatically in the background; allocates a port number; but little platform integration

* Option 1B: as above + add a bit of communication between Electron and Fontra through Javascript message passing through the iframe
  * Increase the integration with Electron while almost not touching the fontra code
  * Sounds a bit clunky though

* Option 2: (what we were going for) split cleanly in fontra the "frontend" bit and the "API" or backend bit; the two communicate via websocket.
  * Option 2A: invasive option that we were doing = make fontra an electron-only app where the frontend is only available through Electron.
    * BAD, dismissed :)
  * Option 2B: refactor fontra so that the frontend is separate from the API but can be reached both through Electron and other means
    * Would be the best for Electron benefits
    * Practical things to figure out:
       * Code splitting (architecture)
       * Code reuse (in practice how to import the code in the various projects)

## Code splitting: new proposed architecture (can be discussed)

- fontra-ui (Javascript, HTML, CSS = frontend code)
- fontra-api (Python backend)
- fontra-browser (CLI entrypoint, same as fontra now)
- fontra-electron (Electron version)

https://excalidraw.com/#json=k3HS12B9Rxe_DF3ytsfqK,7mAqdxEzuWQHXxXLpwbQhQ
Untitled-2022-10-03-1149

NOTE: doesn't mean that they need to be separate Git repositories, they could be subfolders in 1 git repo (might be much easier for collaboration if all is one place, as new features are highly likely to touch files in all projects)

## Code reuse across projects

- relative folder paths if all in same git repo
- git submodules if different repos
  - fontra-electron and fontra-browser have fontra-ui and fontra-api as submodules
- packages
  - fontra-electron and fontra-browser have
    - npm dependency on fontra-ui
    - pip dependency on fontra-api

justvanrossum:
Thanks for this further explanation! I'm still behind with understanding enough about Electron, but this helps a lot.

The main problem with splitting the frontend from the Python API backend is that it breaks the current plug-in structure: currently, ProjectManager plug-ins (as configured via pyproject.toml) provide bits of Python with bits of web contents, via possible external Python packages.

We currently have two concrete ProjectManager plugins:
- "filesystem", as bundled with Fontra
- "rcjk", which talks to a Django-MySQL DB via yet another web API: [fontra-rcjk](https://github.com/BlackFoundryCom/fontra-rcjk/) with [django-robo-cjk](https://github.com/BlackFoundryCom/django-robo-cjk/)

Fontra's plug-in system is by no means set in stone, but separating frontend from backend will require a pretty radical redesign. (Again, totally discussable, as clean separation of frontend and backend does sound appealing in general.)

About platform integration: in my mind, the main features I'd like to achieve are:
- native menus
- native open/save file dialogs

Native menus will require some interaction between the Electron app and the front end code. Open/save will require some interaction with the backend as well.

I'm curious to learn what you have in mind for platform integration, in addition to those two items.

justvanrossum:
Early-morning-epiphany: it is trivial to extract all web assets from a given fontra install, that takes all plug-in assets into account. Perhaps that's an idea: simply rebundle all web assets in a form that is more useful for Electron?

[scrape_web_assets.py.zip](https://github.com/BlackFoundryCom/fontra/files/9703116/scrape_web_assets.py.zip)

justvanrossum:
The only plug-in type that mixes Python and web content is the "fontra.projectmanagers" entry-point, and we only need one of those in an Electron app.

Btw. the "landing" stuff is really a poor ersatz for a proper file open dialog (when running as a desktop app), so we don't even need that in the end.

smugpie:
Good idea about extracting all web assets. On the JavaScript side, we may also be able to use a bundler like Webpack or esbuild to do the same thing - traverse all the plugin folders for web assets and repurpose them for Electron.

PR #142, opened 2022-10-02
Use hatch instead of setuptools

### LINKED ISSUE #141: CLI: argparse subparsers added multiple times
https://github.com/BlackFoundryCom/fontra/blob/986458b0e4c5f7e1d02d3beb00b4521364f66e56/src/fontra/__main__.py#L20-L21

Multiple entry points cause the `filesystem` subparser to be added, which stops working in Python 3.11 due to https://github.com/python/cpython/issues/83897.

I'm not sure what the intended effect of adding them multiple times is?

justvanrossum:
In an attempt to resolve #141 I played with using hatch instead of setuptools. We should probably move into that direction anyway.

@madig, do I remember correctly that you have some experience with hatch? If so, could you have a look if I didn't mess up anything obvious? Or less obvious.

It should probably have an sdist section.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,8 @@
+[build-system]
+requires = ["hatchling", "setuptools-scm[toml]"]
+build-backend = "hatchling.build"
+
+
 [project]
 name = "fontra"
 description = "A browser-based font editor"
@@ -55,13 +60,12 @@ ttf = "fontra.backends.opentype:OTFBackend"
 otf = "fontra.backends.opentype:OTFBackend"
 
 
-[build-system]
-requires = ["setuptools", "setuptools-scm[toml]"]
-build-backend = "setuptools.build_meta"
+[tool.hatch.build.targets.wheel]
+packages = ["src/fontra"]
 
 
-[tool.setuptools.package-data]
-fontra = [
+[tool.hatch.build]
+include = [
     "*.html",
     "*.ico",
     "*.css",
@@ -70,6 +74,11 @@ fontra = [
     "*.svg",
     "*.woff2",
 ]
+
+
+[tool.hatch.version]
+path = "src/fontra/_version.py"
+pattern = "version = '(?P<version>.+)'"
 
 
 [tool.setuptools_scm]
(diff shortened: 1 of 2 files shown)


justvanrossum:
Hm, I did mess up the setuptools_scm bit apparently.

madig:
Some additional stuff:

```toml
[project.urls]
Documentation = "https://github.com/BlackFoundryCom/fontra#readme"
Issues = "https://github.com/BlackFoundryCom/fontra/issues"
Source = "https://github.com/BlackFoundryCom/fontra"

[tool.hatch.envs.default]
dependencies = [
  "pytest-asyncio",
  "pytest-cov",
  "pytest",
]

[tool.hatch.envs.default.scripts]
pytest-cov = "pytest --cov-report=term-missing --cov-config=pyproject.toml --cov=src/fontra --cov=test-py"

[[tool.hatch.envs.test.matrix]]
python = ["310", "311"]

[tool.coverage.run]
branch = true
parallel = true
omit = [
  "src/fontra/_version.py",
]

[tool.coverage.report]
exclude_lines = [
  "no cov",
  "if __name__ == .__main__.:",
  "if TYPE_CHECKING:",
]
```

You can then do stuff like `hatch run pytest` and `hatch run pytest-cov`. You can also add more commands to run e.g. the JS tests. It's sorta like tox. If you do `hatch run test:pytest` or `hatch run test:pytest-cov`, you'll get a matrix run with 3.10 and 3.11 if available.

madig:
Hm yes, the sdist looks incomplete as is.

justvanrossum:
I think I fixed the sdist issue.

madig:
Hm, it includes the `.vscode/settings.json` file on my disk, even though it's gitignored :grimacing:

madig:
Hmhm, https://hatch.pypa.io/latest/plugins/builder/sdist/#default-file-selection says we should actually not have to do anything for sdists, I think.

Do we actually need `tool.hatch.build.include`? A quick look suggests the sdists and wheels include all the good stuff anyway?

madig:
Yes, leaving that out also includes the icons and github folders, so it's more complete.

justvanrossum:
Cool. Yeah, seems to work better that way, thanks!

PR #144, opened 2022-10-02
Fix one `Component` attribute access

madig:
A little typing could help :)

### DIFF
--- a/src/fontra/backends/designspace.py
+++ b/src/fontra/backends/designspace.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from functools import cached_property
 import logging
 import math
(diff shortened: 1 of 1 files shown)


src/fontra/backends/designspace.py
```diff
@@ -10,7 +10,10 @@
 import watchfiles
 from ..core.classes import VariableGlyph, StaticGlyph, Source, Layer
 from ..core.packedpath import PackedPathPointPen
+from typing import TYPE_CHECKING
 
+if TYPE_CHECKING:
+    from fontTools.ufoLib.glifLib import GlyphSet
```
justvanrossum, line 16: This should probably be an unconditional import, given the test failure.

madig: Oh duh. I forgot the most important statement!

justvanrossum: Why not import GlyphSet unconditionally, and forget about the `__future__` and `typing` imports?

madig: I guess. The future import has performance implications, since it postpones annotation evaluation and is slated to become the default at some point in time.

justvanrossum (approved):
Thanks! Good catch.

PR #145, opened 2022-10-02
Some more typing

madig:
I think the `launchWebBrowser` attribute was mistyped. Fixed some more things Pylance complained about. Why was `projectManager` of type `object` before instead of `FileSystemProjectManager`?

And what is the deal with `self.projectManager.setupWebRoutes(self)`? There seems to be no such method defined in the codebase.

### DIFF
--- a/src/fontra/core/server.py
+++ b/src/fontra/core/server.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
 from datetime import datetime, timezone
@@ -9,6 +11,7 @@ import json
 import logging
 import mimetypes
 import traceback
+from typing import Any
 from urllib.parse import quote
 from aiohttp import WSCloseCode, web
 from .remote import RemoteObjectConnection, RemoteObjectConnectionException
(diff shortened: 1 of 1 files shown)


justvanrossum:
`setupWebRoutes` is for plugins, such as https://github.com/BlackFoundryCom/fontra-rcjk/blob/d44150b6b5bb808421a9bc2d661c4f37c3b356db/src/fontra_rcjk/projectmanager.py#L35-L42

justvanrossum:
Feel free to add a no-op `setupWebRoutes`, if that helps.

justvanrossum:
Wait, this isn't working. For typing to be added, we need an abstract class for ProjectManager. It is *not* the concrete `FileSystemProjectManager` class.

madig:
Ah! :) Well, using the "Any" type would work as well for now. Abstract base classes aren't strictly necessary any more, you can use [Protocols](https://mypy.readthedocs.io/en/stable/protocols.html#simple-user-defined-protocols) as well.

justvanrossum:
+1 for Any now and Protocols in the future!

justvanrossum (approved):

PR #258, opened 2023-02-07
Command-a functionality for select/deselect all points

### LINKED ISSUE #248: Command-A to select all points and components
Involves building a new selection object. Should also update the window location.

nedkamburov:
Fixes #248

### DIFF
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -722,9 +722,9 @@ export class EditorController {
     );
   }
 
-  async doSelectAllNone(selectNone) {
+  doSelectAllNone(selectNone) {
     const positionedGlyph =
-      await this.sceneController.sceneModel.getSelectedPositionedGlyph();
+      this.sceneController.sceneModel.getSelectedPositionedGlyph();
 
     if (!positionedGlyph || !this.sceneController.selectedGlyphIsEditing) {
       return;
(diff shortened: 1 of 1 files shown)


src/fontra/views/editor/editor.js
```diff
@@ -714,6 +722,26 @@ export class EditorController {
     );
   }
 
+  async doSelectAllNone(selectNone) {
+    const positionedGlyph =
+      await this.sceneController.sceneModel.getSelectedPositionedGlyph();
+
+    if (positionedGlyph) {
+      const glyphPath = positionedGlyph.glyph.path;
+      const allPoints = new Set();
```
justvanrossum, line 731: Call this variable `newSelection`

src/fontra/views/editor/editor.js
```diff
@@ -714,6 +722,26 @@ export class EditorController {
...
+
+      for (const [pointIndex] of glyphPath.pointTypes.entries()) {
```
justvanrossum, line 733: Do this only if `!selectNone`

src/fontra/views/editor/editor.js
```diff
@@ -714,6 +722,26 @@ export class EditorController {
...
+        allPoints.add(`point/${pointIndex}`);
+      }
+
+      this.sceneController.selection = allPoints;
+    }
+
+    if (selectNone) {
```
justvanrossum, line 740: ...and get rid of this block, should be handled above.

src/fontra/views/editor/editor.js
justvanrossum: (Good use of `.entries()`! I didn't think of that.)

justvanrossum: In the same `!selectNone` block, also add a loop that adds all `component/${componentIndex}` indices.

nedkamburov:
I made the changes. I hope adding glyphPath points and glyphComponents covers all cases?

justvanrossum (approved):
Perfect, thanks!

PR #270, opened 2023-02-09
Refactor shortcut handling

### LINKED ISSUE #269: [tools] Tool shortcuts need to NOT use command/control modifier key
Command 1, 2, 3 etc (presumably also control on Windows) actually switches tabs in all browsers, so may be confusing.

Also: capturing command 1, 2, 3 does not work properly in Safari: the tab switching has precedence.

justvanrossum:
- replace dumb switch with flexible registry
- this enables a fix for #269: use plain 1, 2, 3 etc. keys to switch tools

Also fixes #269

### DIFF
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -705,6 +705,20 @@ export class EditorController {
   }
 
   registerShortCut(keysOrCodes, modifiers, callback) {
+    //
+    // Register a shortcut handler
+    //
+    // `keysOrCodes` is a list of event codes or a string or list of key strings.
+    // Any item in the list or string will be seen as a trigger for the handler.
+    //
+    // `modifiers` is an object that allows you to match a specific boolean event
+    // property. For example, { shiftKey: false } requires that the shift key must
+    // not be pressed. If shiftKey is undefined, the state of the shift key is not
+    // taken into account when matching the handler.
+    //
+    // `callback` is a callable that will be called with the event as its single
+    // argument.
+    //
     for (const keyOrCode of keysOrCodes) {
       const handlerDef = { ...modifiers, callback };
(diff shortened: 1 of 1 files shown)


src/fontra/views/editor/editor.js
```diff
@@ -665,75 +666,101 @@ export class EditorController {
...
+    this.registerShortCut("-", { metaKey: true, globalOverride: true }, () => {
+      this.zoomOut();
+    });
+    this.registerShortCut("+=", { metaKey: true, globalOverride: true }, () => {
+      this.zoomIn();
+    });
+    this.registerShortCut("0", { metaKey: true, globalOverride: true }, () => {
+      this.zoomFit();
+    });
+    this.registerShortCut("123456789", { metaKey: false }, (event) => {
```
nedkamburov, line 685: Nicee, so now it will work without the metakey and it's very compact! 🚀

src/fontra/views/editor/editor.js
```diff
@@ -665,75 +666,101 @@ export class EditorController {
...
+
+    this.registerShortCut("z", { metaKey: true, shiftKey: false }, () => {
+      this.doUndoRedo(false);
+    });
+    this.registerShortCut("z", { metaKey: true, shiftKey: true }, () => {
+      this.doUndoRedo(true);
+    });
+  }
+
+  registerShortCut(keysOrCodes, modifiers, callback) {
```
nedkamburov, line 707: Great, it seems very flexible and the comment is clear.

nedkamburov (approved):
Nice, it will be much easier to add more shortcuts. But maybe we can split it in a different file later because I can image other 'views' can also use some of these?

justvanrossum:
> maybe we can split it in a different file later because I can image other 'views' can also use some of these?

Good point, but for later.

PR #282, opened 2023-02-14
Add shortcuts labels to the context menu items

### LINKED ISSUE #255: Context menu needs to be able to show short cuts

nedkamburov:
Fixes #255 
It adds shortcuts combinations to the items in the context menus if they exist. I haven't tested it on Windows but the code it's written, it should work properly replacing `⌘` with `Ctrl+` and so on.

Here are all possibilities for shortcut modifiers:
https://apple.stackexchange.com/a/55729/452817

### DIFF
--- a/src/fontra/client/core/context-menu.js
+++ b/src/fontra/client/core/context-menu.js
@@ -21,10 +21,9 @@ export class ContextMenu {
         this.element.appendChild(dividerElement);
       } else {
         const itemElement = document.createElement("div");
-        const itemTitle = this.buildTitle(item);
         itemElement.classList.add("context-menu-item");
         itemElement.classList.toggle("enabled", !!item.enabled());
-        itemElement.append(itemTitle);
+        itemElement.append(this.buildTitle(item));
         itemElement.onmouseenter = (event) => this.selectItem(itemElement);
         itemElement.onmousemove = (event) => {
           if (!itemElement.classList.contains("selected")) {
--- a/src/fontra/client/core/context-menu.js
+++ b/src/fontra/client/core/context-menu.js
@@ -1,4 +1,4 @@
-import { reversed } from "./utils.js";
+import { reversed, buildShortcut } from "./utils.js";
 
 export const MenuItemDivider = { title: "-" };
 
@@ -134,9 +134,9 @@ export class ContextMenu {
     const titleWrapper = document.createElement("div");
     const title = document.createElement("span");
     title.innerText = typeof item.title === "function" ? item.title() : item.title;
+
     const shortCut = document.createElement("span");
-    shortCut.innerText = item.shortCut ? item.shortCut.keysOrCodes : "";
-    // console.log(item.shortCut.keysOrCodes);
+    shortCut.innerHTML = buildShortcut(item.shortCut);
     titleWrapper.append(title, shortCut);
     return titleWrapper;
   }
--- a/src/fontra/client/core/context-menu.js
+++ b/src/fontra/client/core/context-menu.js
@@ -1,4 +1,4 @@
-import { reversed, buildShortcut } from "./utils.js";
+import { reversed } from "./utils.js";
 
 export const MenuItemDivider = { title: "-" };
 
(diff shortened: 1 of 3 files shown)


src/fontra/client/core/utils.js
```diff
@@ -221,3 +221,23 @@ export function makeUPlusStringFromCodePoint(codePoint) {
     ? "U+" + codePoint.toString(16).toUpperCase().padStart(4, "0")
     : "";
 }
+
+export function buildShortcut(shorCutRegister) {
```
justvanrossum, line 225: - Please keep this function in context-menu.js for now.
- I would name the function `buildShortCutString()`
- I would name its argument `shortCutDefinition`

src/fontra/client/core/utils.js
```diff
@@ -221,3 +221,23 @@ export function makeUPlusStringFromCodePoint(codePoint) {
...
+    const isMac = navigator.platform.toLowerCase().indexOf("mac") >= 0;
+
+    if (shorCutRegister.shiftKey) {
+      shorcutCommand += isMac ? "&#8679;" : "Shift+"; // ⇧ or Shift
+    }
+    if (shorCutRegister.metaKey) {
+      shorcutCommand += isMac ? "&#8984;" : "Ctrl+"; // ⌘ or Ctrl
+    }
+    if (shorCutRegister.keysOrCodes) {
+      shorcutCommand += shorCutRegister.keysOrCodes.toUpperCase();
```
justvanrossum, line 238: While this works for the current cases, `keysOrCodes` is meant to accept either a string of multiple single chars (in that case, take the first), or a list of key codes. In the latter case, take the first item of the list, but this will be a string like "ArrowUp", and we need to see how to represent the key code.

The "Delete" use case will be a nice test. I suggest to leave this as it is for now, but count on having to change it for the "Delete" short cut.

justvanrossum (changes requested):

justvanrossum:
Btw. it looks great!

image

I would love to reduce the right hand margin for the short cuts, but not for a long title that doesn't have one. Not sure what the cleanest way to achieve that would be, although the following is effective:
```diff
diff --git a/src/fontra/client/core/context-menu.js b/src/fontra/client/core/context-menu.js
index 5f8393fe..f6d71fce 100644
--- a/src/fontra/client/core/context-menu.js
+++ b/src/fontra/client/core/context-menu.js
@@ -136,7 +136,7 @@ export class ContextMenu {
     title.innerText = typeof item.title === "function" ? item.title() : item.title;
 
     const shortCut = document.createElement("span");
-    shortCut.innerHTML = buildShortcut(item.shortCut);
+    shortCut.innerHTML = buildShortcut(item.shortCut) || "&nbsp;";
     titleWrapper.append(title, shortCut);
     return titleWrapper;
   }
diff --git a/src/fontra/client/css/context-menu.css b/src/fontra/client/css/context-menu.css
index f33eaf81..62a600b3 100644
--- a/src/fontra/client/css/context-menu.css
+++ b/src/fontra/client/css/context-menu.css
@@ -86,7 +86,7 @@
 }
 
 .context-menu-item {
-  padding: 0.2em 1.5em 0.2em 1em; /* top, right, bottom, left */
+  padding: 0.2em 0.8em 0.2em 1em; /* top, right, bottom, left */
   color: #808080a0;
 }
 
@@ -102,5 +102,6 @@
 
 .context-menu-item > div {
   display: flex;
+  gap: 0.5em;
   justify-content: space-between;
 }
```

justvanrossum:
After that patch:
image

src/fontra/client/core/utils.js
nedkamburov: Mhh, maybe we can create some dictionary with definition and representation on different platforms and just look for them when we need it. E.g
```
const shortcutDefinitions = [
  {
    key: "ArrowUp",
    specialChar: "↑", // Used on a Mac or Windows but when it's available
    fallbackChar: "&#8593;", // Used as a fallback
  },
  {
    key: "Delete",
    specialChar: "⌫",
    fallbackChar: "&#9003;",
  },
  {
    key: "Home",
    specialChar: "", // Empty slot when it's not available
    fallbackChar: "Home", // Used as a fallback
  },
];

// To access a key we can browse it like this:
shortcutDefinitions.find(shortcut => shortcut.key === shorCutRegister.keysOrCodes);
```

justvanrossum: Something like that sounds like a good idea.

I think `"↑"` is 100% equivalent to `"&#8593;"`, it's just a different way of spelling it: unicode vs html entity. Modern browsers should all be able to display such chars, and it may be more a matter of following the platform conventions of how short cuts are displayed.

src/fontra/client/core/context-menu.js
```diff
@@ -21,10 +21,10 @@ export class ContextMenu {
         this.element.appendChild(dividerElement);
       } else {
         const itemElement = document.createElement("div");
-        const itemTitle = typeof item.title === "function" ? item.title() : item.title;
+        const itemTitle = this.buildTitle(item);
         itemElement.classList.add("context-menu-item");
         itemElement.classList.toggle("enabled", !!item.enabled());
-        itemElement.innerText = itemTitle;
+        itemElement.append(itemTitle);
```
justvanrossum, line 27:
```suggestion
        itemElement.append(this.buildTitle(item));
```
And remove line 24.

justvanrossum (changes requested):

justvanrossum (approved):

PR #290, opened 2023-02-16
Fix Safari ctrl+click context menu

### LINKED ISSUE #287: Context menu doesn't work on Safari, secondary issue on Chrome/Ff
- [x] Control-click shows the context menu briefly, but it immediately disappears; holding the "mouse" down does keep the menu visible, but the action doesn't work at all. Fixed in #290.
- [ ] If you hold/drag the "mouse" after control-click, choosing an item doesn't work without an additional click. Fixed in #802.

nedkamburov:
Fixes #287

### DIFF
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -188,7 +188,6 @@ export class EditorController {
     this.canvasController.canvas.addEventListener("contextmenu", (event) =>
       this.contextMenuHandler(event)
     );
-
     window.addEventListener("click", (event) => this.dismissContextMenu(event));
     window.addEventListener("blur", (event) => this.dismissContextMenu(event));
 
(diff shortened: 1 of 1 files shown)


src/fontra/views/editor/editor.js
```diff
@@ -188,6 +188,7 @@ export class EditorController {
     this.canvasController.canvas.addEventListener("contextmenu", (event) =>
       this.contextMenuHandler(event)
     );
+
```
justvanrossum, line 191: Can you undo this redundant change, and the one near line 962?

src/fontra/views/editor/editor.js
```diff
@@ -958,6 +958,7 @@ export class EditorController {
         return;
       }
     }
+
```
justvanrossum, line 961:
```suggestion
```

PR #297, opened 2023-02-21
Delete points and components issue295

### LINKED ISSUE #295: Delete key to remove points and components
The menu item, short cut and implementation stub were added in #289.

nedkamburov:
Fixes #295 
WIP 
To-dos:
- [x] delete single points and multiple points on the same segment
- [x] delete multiple points from multiple segments
- [x] delete components


src/fontra/views/editor/scene-controller.js
```diff
@@ -504,6 +504,41 @@ export class SceneController {
...
+      const { point: pointSelection } = parseSelection(this.selection);
+
+      const changes = recordChanges(instance, (instance) => {
+        const path = instance.path;
+
+        let offset = 0;
+        for (const point of pointSelection) {
+          const [contourIndex, contourPointIndex] = path.getContourAndPointIndex(point);
+          const selectedContours = getSelectedContours(path, pointSelection);
+          const contour = path.getUnpackedContour(contourIndex);
```
justvanrossum, line 518: To get at the number of points in a contour, you can do 
```js
const numContourPoints = path.getNumPointsOfContour(contourIndex);
```
This allows you to avoid the `getUnpackedContour()` call.

src/fontra/views/editor/scene-controller.js
```diff
@@ -504,6 +504,41 @@ export class SceneController {
...
 
+  async doDelete() {
+    await this.editInstance((sendIncrementalChange, instance) => {
+      const { point: pointSelection } = parseSelection(this.selection);
+
+      const changes = recordChanges(instance, (instance) => {
+        const path = instance.path;
+
+        let offset = 0;
+        for (const point of pointSelection) {
```
justvanrossum, line 515: Please rename `point` to `pointIndex`

src/fontra/views/editor/scene-controller.js
```diff
@@ -504,6 +504,41 @@ export class SceneController {
...
+          const [contourIndex, contourPointIndex] = path.getContourAndPointIndex(point);
+          const selectedContours = getSelectedContours(path, pointSelection);
```
justvanrossum, line 517: This line is invariant in the loop, so should be moved to before the loop.

src/fontra/views/editor/scene-controller.js
justvanrossum: Also, if you change it to 
```js
for (const pointIndex of reversed(pointSelection)) {
```
You can get rid of the `offset` logic.

(`reversed` lives in utils.js)

(The output of `parseSelection` is guaranteed to be sorted)

src/fontra/views/editor/scene-controller.js
```diff
@@ -504,6 +504,41 @@ export class SceneController {
     return undoInfo !== undefined;
   }
 
+  async doDelete() {
+    await this.editInstance((sendIncrementalChange, instance) => {
+      const { point: pointSelection } = parseSelection(this.selection);
```
justvanrossum, line 509: We'll also need to handle components (`instance.component`). This will be easier.

I think we should move the point logic to a new function in path-functions.js: `deleteSelectedPoints(path, pointIndices)` (everything inside the recordChanges block).

And implement the component logic here. Iterate over the componentIndices in reverse.

justvanrossum: I also think that the doDelete implementation can stay in editor.js, like breakContour and friends.

justvanrossum (changes requested):

nedkamburov:
I think I am done with all tasks. You can test it and merge if there is nothing more to add.

src/fontra/views/editor/editor.js
```diff
@@ -874,11 +874,14 @@ export class EditorController {
...
+    return (
+      this.sceneController.selectedGlyphIsEditing &&
+      this.sceneController.selection.size > 0
+    );
   }
 
-  doDelete() {
-    console.log("delete");
+  async doDelete() {
+    await this.sceneController.doDelete();
```
justvanrossum, line 884: I would love the code to live here, not in SceneController.

src/fontra/client/core/path-functions.js
```diff
@@ -114,3 +114,24 @@ export function connectContours(path, sourcePointIndex, targetPointIndex) {
...
+
+    if (numContourPoints > 1) {
+      path.deletePoint(contourIndex, contourPointIndex);
+    } else {
+      path.deleteContour(contourIndex);
+    }
+  }
+}
+
+export function getSelectedContours(path, pointIndices) {
```
justvanrossum, line 131: Since you didn't end up using this function in deleteSelectedPoints, I would prefer it to stay where it was. I'm not yet convinced it is a super generally useful function.

justvanrossum: This PR should effectively leave scene-controller.js alone.

justvanrossum (changes requested):

src/fontra/client/core/path-functions.js
nedkamburov: No problem, will revert back to the old version.

justvanrossum (approved):

PR #298, opened 2023-02-22
Copy a glyph as svg to copy to Illustrator and Figma

### LINKED ISSUE #260: Possibility to Paste copied contours from external app (Illustrator / RoboFont / Glyphs)
Paste the clipboard content as contours if possible in a Fontra glyph

nedkamburov:
Part of #260


src/fontra/client/core/glyph-svg.js
```diff
@@ -51,3 +51,19 @@ function formatCoordinate(x, y, numerator, denominator, dx, dy) {
...
+
+export function pathToSVG(instance, bounds) {
+  const path = new SVGPath2D();
+  instance.path.drawToPath2d(path);
+  const pathString = path.getPath();
+
+  const width = bounds["xMax"] - bounds["xMin"];
+  const height = bounds["yMax"] - bounds["yMin"];
+
+  const svgString = `<svg xmlns="http://www.w3.org/2000/svg"
```
justvanrossum, line 63: The extra indentation in the SVG data should be removed: it's more important that the SVG looks correctly indented then our code.

src/fontra/client/core/glyph-svg.js
```diff
@@ -51,3 +51,19 @@ function formatCoordinate(x, y, numerator, denominator, dx, dy) {
...
+                      width="${width}"
+                      height="${height}"
+                      viewBox="0 0 ${width} ${height}">
```
justvanrossum, line 66: I think the four fields need to be literally `xMin yMin xMax yMax`. The min values should be passed through `Math.floor()` and the max values through `Math.ceil()`.

src/fontra/client/core/glyph-svg.js
```diff
@@ -51,3 +51,19 @@ function formatCoordinate(x, y, numerator, denominator, dx, dy) {
...
+                      <path d="${pathString}"/></svg>`;
```
justvanrossum, line 67: This needs an y-axis flip transformation, something in this direction:
```
<path d="${pathString}" transform="matrix(1 0 0 -1 0 ${height})" /></svg>`
```

src/fontra/client/core/utils.js
```diff
@@ -221,3 +221,15 @@ export function makeUPlusStringFromCodePoint(codePoint) {
     ? "U+" + codePoint.toString(16).toUpperCase().padStart(4, "0")
     : "";
 }
+
+export async function writeToClipboard(clipboardObject) {
+  if (!clipboardObject) return;
+
+  for (const [key, value] of Object.entries(clipboardObject)) {
+    clipboardObject[key] = new Blob([`${value}`], {
```
justvanrossum, line 229: It is not nice to modify an object we got passed as an argument: it would be better to make a fresh object, something like this (untested):

```js
    const clipboardObject = Object.fromEntries(
      Object.entries(clipboardObject).map((item) => [item[0], new Blob([item[1]])])
    );
```
```

**Review by justvanrossum — CHANGES_REQUESTED** [review]

File: src/fontra/client/core/glyph-svg.js (hunk 2 above)
Comment (justvanrossum) [reply src/fontra/client/core/glyph-svg.js:66]: Perhaps the y values need work after the y-axis flip transformation. To test: save the svg as a file and open in Safari, it nicely shows how the svg is clipped.

Comment (justvanrossum) [reply src/fontra/client/core/glyph-svg.js:66]: To clarify, "0 0" for the first two fields cause any part of the shape that has negative x coordinates to be cropped.

File: src/fontra/client/core/utils.js [hunk 5]
```diff
@@ -221,3 +221,16 @@ export function makeUPlusStringFromCodePoint(codePoint) {
     ? "U+" + codePoint.toString(16).toUpperCase().padStart(4, "0")
     : "";
 }
+
+export async function writeToClipboard(clipboardObject) {
+  if (!clipboardObject) return;
+
+  const clipboardItemObject = {};
+  for (const [key, value] of Object.entries(clipboardObject)) {
+    clipboardItemObject[key] = new Blob([`${value}`], {
```
justvanrossum, line 230: Can this be
```suggestion
    clipboardItemObject[key] = new Blob([value], {
```
?

src/fontra/views/editor/editor.js
```diff
@@ -852,11 +854,27 @@ export class EditorController {
   }
 
   canCopy() {
-    return true;
+    return this.sceneController.selection.size > 0;
```
justvanrossum, line 857: I may want to revisit this: I want an empty selection to mean "copy the whole glyph"
Perhaps I also want copy to work if a glyph is merely only selected, and not only in edit mode.

nedkamburov: We can use selectedGlyph if that makes more sense?

justvanrossum: Yeah, I think so. Funnily enough, the behavior I just described already works :)

src/fontra/client/core/glyph-svg.js
```diff
@@ -51,3 +51,20 @@ function formatCoordinate(x, y, numerator, denominator, dx, dy) {
...
+
+  const width = bounds["xMax"] - bounds["xMin"];
+  const height = bounds["yMax"] - bounds["yMin"];
+
+  const svgString = `<svg xmlns="http://www.w3.org/2000/svg" width="${width}" height="${height}"
+  viewBox="${Math.floor(bounds["xMin"])} ${Math.floor(bounds["yMin"])} ${Math.ceil(
+    bounds["xMax"]
+  )} ${Math.ceil(
+    bounds["yMax"]
+  )}"><path transform="matrix(1 0 0 -1 0 ${height})" d="${pathString}"/></svg>`;
```
justvanrossum, line 68: Please insert newlines before and after the `<path>` element.

src/fontra/client/core/glyph-svg.js
```diff
@@ -51,3 +51,20 @@ function formatCoordinate(x, y, numerator, denominator, dx, dy) {
...
+
+  const width = bounds["xMax"] - bounds["xMin"];
+  const height = bounds["yMax"] - bounds["yMin"];
+
+  const svgString = `<svg xmlns="http://www.w3.org/2000/svg" width="${width}" height="${height}"
+  viewBox="${Math.floor(bounds["xMin"])} ${Math.floor(bounds["yMin"])} ${Math.ceil(
+    bounds["xMax"]
+  )} ${Math.ceil(bounds["yMax"])}">
+  <path transform="matrix(1 0 0 -1 0 ${height})" d="${pathString}"/>
+  </svg>`;
```
justvanrossum, line 68:
```suggestion
</svg>`;
```

src/fontra/client/core/glyph-svg.js
```diff
@@ -51,3 +51,20 @@ function formatCoordinate(x, y, numerator, denominator, dx, dy) {
   y = Math.round(y * numerator + dy) / denominator;
   return `${x},${y}`;
 }
+
+export function pathToSVG(instance, bounds) {
```
justvanrossum, line 55:
```suggestion
export function pathToSVG(path, bounds) {
```

src/fontra/client/core/glyph-svg.js
```diff
@@ -51,3 +51,20 @@ function formatCoordinate(x, y, numerator, denominator, dx, dy) {
...
+  const path = new SVGPath2D();
+  instance.path.drawToPath2d(path);
```
justvanrossum, line 57:
```suggestion
  path.drawToPath2d(path);
```

src/fontra/views/editor/editor.js
```diff
@@ -852,11 +854,26 @@ export class EditorController {
...
   doCopy() {
-    console.log("copy");
+    const positionedGlyph =
+      this.sceneController.sceneModel.getSelectedPositionedGlyph();
+    const glyphController = positionedGlyph?.glyph;
+    if (!glyphController) return;
+
+    const instance = glyphController?.instance;
+    const bounds = glyphController.controlBounds;
+    const svgString = pathToSVG(instance, bounds);
```
justvanrossum, line 868:
```suggestion
    const svgString = pathToSVG(glyphController.flattenedPath, bounds);
```

justvanrossum (commented):
See suggestions (untested)

src/fontra/client/core/glyph-svg.js
```diff
@@ -51,3 +51,20 @@ function formatCoordinate(x, y, numerator, denominator, dx, dy) {
   y = Math.round(y * numerator + dy) / denominator;
   return `${x},${y}`;
 }
+
+export function pathToSVG(path, bounds) {
+  const svgPath = new SVGPath2D();
+  path.drawToPath2d(svgPath);
+  const pathString = svgPath.getPath();
+
+  const width = bounds["xMax"] - bounds["xMin"];
```
justvanrossum, line 60: We need to do a slightly bigger change here. Something along these lines:
```js
  const xMin = Math.floor(bounds.xMin);
  const yMin = Math.floor(bounds.yMin);
  const xMax = Math.floor(bounds.xMax);
  const yMax = Math.floor(bounds.yMax);
  const width = xMax - xMin;
  const height = yMax - yMin;
```
Then, the viewbox will be (bear with me): xMin, 0, xMax, height
And the sixth matrix argument will be yMax.

justvanrossum: I see your recent change, but please also do:
> Then, the viewbox will be (bear with me): xMin, 0, xMax, height
> And the sixth matrix argument will be yMax.

(Please check for unused variables: at least `width` will remain unused I think, but perhaps others.)

nedkamburov: Width is used for width attribute of the svg, so all good.

src/fontra/client/core/glyph-svg.js
```diff
@@ -51,3 +51,22 @@ function formatCoordinate(x, y, numerator, denominator, dx, dy) {
...
 }
+
+export function pathToSVG(path, bounds) {
+  const svgPath = new SVGPath2D();
+  path.drawToPath2d(svgPath);
+  const pathString = svgPath.getPath();
+
+  const xMin = Math.floor(bounds.xMin);
+  const yMin = Math.floor(bounds.yMin);
+  const xMax = Math.floor(bounds.xMax);
```
justvanrossum, line 62:
```suggestion
  const xMax = Math.ceil(bounds.xMax);
```

src/fontra/client/core/glyph-svg.js
```diff
@@ -51,3 +51,22 @@ function formatCoordinate(x, y, numerator, denominator, dx, dy) {
...
+  const yMax = Math.floor(bounds.yMax);
```
justvanrossum, line 63:
```suggestion
  const yMax = Math.ceil(bounds.yMax);
```

src/fontra/client/core/glyph-svg.js
```diff
@@ -51,3 +51,22 @@ function formatCoordinate(x, y, numerator, denominator, dx, dy) {
...
+  const width = xMax - xMin;
+  const height = yMax - yMin;
+
+  const svgString = `<svg xmlns="http://www.w3.org/2000/svg" width="${width}" height="${height}"
+  viewBox="${xMin} 0 ${xMax} ${height}">
```
justvanrossum, line 68:
```suggestion
  viewBox="${xMin} 0 ${width} ${height}">
```
This fixes the display in Safari.

justvanrossum (approved):

PR #318, opened 2023-02-28
Copy paste custom data in Safari and Firefox

### LINKED ISSUE #312: [clipboard] Implement custom clipboard to work around Safari and Firefox limitations

Relates to #260 and #75.

nedkamburov:
Fixes #312 
This manages to:
– copy/paste within Fontra in any browser
– copy/paste between Fontra and RF in any browser

### DIFF
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -235,16 +235,12 @@ export async function writeToClipboard(clipboardObject) {
   try {
     await navigator.clipboard.write([new ClipboardItem(clipboardItemObject)]);
   } catch {
-    // Write at least the plain/text MIME type to the clipboard
-    if (clipboardObject["text/plain"])
-      try {
-        await navigator.clipboard.writeText(clipboardObject["text/plain"]);
-      } catch {
-        // TODO: Safari complains that this needs to be handled. It's probably right.
-      }
-
     // Write to LocalStorage for internal Fontra use
     writeToLocalStorage(clipboardObject);
+
+    // Write at least the plain/text MIME type to the clipboard
+    if (clipboardObject["text/plain"])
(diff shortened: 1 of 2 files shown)


src/fontra/client/core/utils.js
```diff
@@ -232,7 +234,20 @@ export async function writeToClipboard(clipboardObject) {
     });
   }
 
-  navigator.clipboard.write([new ClipboardItem(clipboardItemObject)]);
+  try {
+    await navigator.clipboard.write([new ClipboardItem(clipboardItemObject)]);
+  } catch {
+    // Write at least the plain/text MIME type to the clipboard
+    if (clipboardObject["text/plain"])
+      try {
```
justvanrossum, line 242: I'm not sure this second try/catch is needed.

nedkamburov: Safari throws an error for unhandled promise 
This can be ` await navigator.clipboard.writeText(clipboardObject["text/plain"].catch(error => {})` but the need remains

justvanrossum: You're saying that a plain
```js
await navigator.clipboard.write([new ClipboardItem(clipboardItemObject)])
```
without try/catch doesn't work here?

nedkamburov: if you do it without `await`, It works but Safari says this:
`Unhandled Promise Rejection: NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.`

If I write this: 
```
if (clipboardObject["text/plain"])
      await navigator.clipboard.writeText(clipboardObject["text/plain"]);

writeToLocalStorage(clipboardObject);
```
no errors, but `writeToLocalStorage()` is not executed...

PR #325, opened 2023-03-02
Safari copy/paste plain text to the clipboard

### LINKED ISSUE #312: [clipboard] Implement custom clipboard to work around Safari and Firefox limitations

nedkamburov:
Partially fixes #312

### DIFF
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -226,9 +226,10 @@ export class EditorController {
     this.sceneController.addEventListener("selectedGlyphChanged", () =>
       this.updateWindowLocationAndSelectionInfo()
     );
-    this.sceneController.addEventListener("selectionChanged", () =>
-      this.updateWindowLocationAndSelectionInfo()
-    );
+    this.sceneController.addEventListener("selectionChanged", async () => {
+      this.updateWindowLocationAndSelectionInfo();
+      await this.buildGLIFAndSVGString();
+    });
 
     window.addEventListener("popstate", (event) => {
       this.setupFromWindowLocation();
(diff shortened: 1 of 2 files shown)


src/fontra/views/editor/editor.js
```diff
@@ -692,25 +692,34 @@ export class EditorController {
     }
     this.basicContextMenuItems.push(MenuItemDivider);
 
+    const safariHTTPAgent =
+      navigator.userAgent.indexOf("Safari") > -1 &&
```
justvanrossum, line 696: In Chrome, the userAgent string also contains the substring "Safari".

src/fontra/views/editor/editor.js
```diff
@@ -692,25 +692,34 @@ export class EditorController {
...
+      window.location.protocol === "http:";
+    if (safariHTTPAgent) this.initFallbackClipboardEventListeners();
+    if (!safariHTTPAgent) {
```
justvanrossum, line 699: This can be an else of the previous if.

src/fontra/views/editor/editor.js
```diff
@@ -692,25 +692,34 @@ export class EditorController {
     }
     this.basicContextMenuItems.push(MenuItemDivider);
 
+    const safariHTTPAgent =
+      navigator.userAgent.indexOf("Safari") > -1 &&
+      window.location.protocol === "http:";
+    if (safariHTTPAgent) this.initFallbackClipboardEventListeners();
```
justvanrossum, line 698: Please make this a block with { and }.

src/fontra/views/editor/editor.js
```diff
@@ -910,6 +934,16 @@ export class EditorController {
     await writeToClipboard(clipboardObject);
   }
 
+  _copyToEventClipboard(event) {
+    if (!event) return;
```
justvanrossum, line 938: I don't see how this method gets called with a falsy event. Perhaps this line is redundant?

src/fontra/views/editor/editor.js
```diff
@@ -776,6 +785,30 @@ export class EditorController {
...
+        event.preventDefault();
+        this._copyToEventClipboard(event);
+      }
+    });
+
+    window.addEventListener("cut", (event) => {
+      if (document.activeElement === this.canvasController.canvas) {
+        event.preventDefault();
+        this._copyToEventClipboard(event);
+        this.doCut(event);
```
justvanrossum, line 807: Instead of passing the event, let's just pass a flag. I think the flag argument coud be called "writeToClipboard", which should be `true` by default, so we'd pass `false` here.

src/fontra/views/editor/editor.js
```diff
@@ -858,7 +891,7 @@ export class EditorController {
     return !!this.sceneController.selection.size;
   }
 
-  async doCut() {
+  async doCut(event = null) {
```
justvanrossum, line 894:
```suggestion
  async doCut(writeToClipboard = true) {
```

src/fontra/views/editor/editor.js
```diff
@@ -868,7 +901,7 @@ export class EditorController {
       this.sceneController.selection = new Set();
       return "Cut Selection";
     });
-    if (copyResult) {
+    if (copyResult && !event) {
```
justvanrossum, line 904:
```suggestion
    if (copyResult && writeToClipboard) {
```

justvanrossum (changes requested):

src/fontra/views/editor/editor.js
justvanrossum: https://stackoverflow.com/questions/7944460/detect-safari-browser has some useful answers. I like this one:
```js
var isSafari = navigator.vendor && navigator.vendor.indexOf('Apple') > -1 &&
               navigator.userAgent &&
               navigator.userAgent.indexOf('CriOS') == -1 &&
               navigator.userAgent.indexOf('FxiOS') == -1;
```

nedkamburov: Maybe a combo?
```
const isSafari =
      navigator.vendor &&
      navigator.vendor.indexOf("Apple") > -1 &&
      navigator.userAgent.indexOf("Safari") > -1 &&
      navigator.userAgent.indexOf("CriOS") == -1 &&
      navigator.userAgent.indexOf("FxiOS") == -1 &&
      window.safari !== undefined &&
      window.location.protocol === "http:";
```

justvanrossum: Seems a bit redundant. Maybe only `window.safari !== undefined` and the `http:` check? I understand the former is only for desktop, which is fine for us actually.

nedkamburov: Yes, the `vendor` and` userAgent.indexOf `are probably going to give us some false positives.
Going back to the simple stuff!

justvanrossum (approved):

PR #326, opened 2023-03-02
Fallback Copy/Paste within Fontra via LocalStorage

### LINKED ISSUE #312: [clipboard] Implement custom clipboard to work around Safari and Firefox limitations

nedkamburov:
Helps Firefox to copy/paste 'custom' MIME type
Fixes #312

### DIFF
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -232,7 +232,17 @@ export async function writeToClipboard(clipboardObject) {
     });
   }
 
-  await navigator.clipboard.write([new ClipboardItem(clipboardItemObject)]);
+  try {
+    await navigator.clipboard.write([new ClipboardItem(clipboardItemObject)]);
+  } catch {
+    // Write to LocalStorage for internal Fontra use
+    writeToLocalStorage(clipboardObject);
+
+    // Write at least the plain/text MIME type to the clipboard
+    if (clipboardObject["text/plain"]) {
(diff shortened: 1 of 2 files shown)


src/fontra/views/editor/editor.js
```diff
@@ -1017,25 +1017,25 @@ export class EditorController {
...
     const plainText = await readFromClipboard("text/plain");
-    if (plainText) {
-      pastedGlyph = await this.fontController.parseClipboard(plainText);
+    if (!plainText) {
+      return;
     }
 
-    const isTheSameGlyphPath =
-      JSON.stringify(pastedGlyph) === localStorage.getItem("clipboardSelection.glyph");
+    const isClipboardValid =
```
justvanrossum, line 1024: I prefer to name it more explicitly, perhaps even`isOurLocalClipboardValid`. It is not about the clipboard in general after all.

nedkamburov: Just to clarify: by 'Local Clipboard' you refer to the localStorage values, right?

src/fontra/client/core/utils.js
```diff
@@ -254,3 +264,16 @@ export async function readFromClipboard(type) {
   }
   return undefined;
 }
+
+function writeToLocalStorage(clipboardObject) {
+  const fallbackTypes = {
```
justvanrossum, line 269: If you write this as a list of lists (a length-two list instead of a key/value pair), the for loop below won't need to `Object.entries` bit.

Issue #246, opened 2023-02-06
Figure out how to deploy npm dependencies


justvanrossum:
- http via a npm cdn, such as https://www.jsdelivr.com/
  - disadvantage: user needs to be online even for a local Fontra such as Fontra Pak
  - might be good to start this way, so we're not blocked until we've got a good solution
- use a bundler (artcle: https://snipcart.com/blog/javascript-module-bundler)
  - bundle into our fontra Python package? keeps Fontra nicely self-contained, so we can upload to PyPI

fabiocaccamo:
Should check also: https://vitejs.dev
It's the ultimate alternative and seems very promising.

nedkamburov:
After a bit of research, I can share my thoughts on the choice of technology here. Have in mind, my experience is limited and it is mainly revolving around huge production builds with Webpack.

Main technologies of interest for us seem to be:
– Rollup.js https://rollupjs.org/introduction/#the-why 
– Vite.js https://vitejs.dev/guide/why.html 
– Webpack
– *esbuild (written in Go)
– *Parcel (similar to Vite.js)


TLDR;
Long story short, we will have to make our priorities clear. For me the battle seems to be between Vite.js and Rollup.js
1. Do we want to have a development server with no configuration upfront to see changes right away? (Vite wins)
2. Do we prepare for a switch to Typescript from now? (Vite wins)
3. Do we want to keep it simple and configure only what we need with bigger control? (Rollup wins)
4. Do we want a system of plugins, such as polyfills? (Rollup wins, although Vite also can use it)

--- 
1. Rollup.js is a smart bundler that is using native ESmodules and makes build really fast compared to clunky, cumbersome Webpack. Rollup is established and it has rich library of with tutorials, plugins and configurations.
2. Vite.js is built on top of Rollup.js (it can use it's plugin ecosystem) and it basically provides an extra layer that optimises development by rebuilding only what's needed. it uses its own development server for that. 
In general, I don't see this as a huge problem for Fontra as of now. Vite also does NOT require much configuration but having in mind that we will do this once and just build upon it, it wouldn't matter that much perhaps.
Here is a small list of comparison: https://ruslan.rocks/posts/vite-vs-rollup
5. Webpack provides much more flexibility and customisability but it's really its own world. It is usually used for apps with huge number of dependencies, assets and pages. I don't think it's worth going that deep for our current direction.
[...]

PR #334, opened 2023-03-03
Deploy npm dependencies. #216

fabiocaccamo:
This is a PR for #246

- Add a new Python script `scripts/jsdeps.py` that can be executed also via Node by calling `npm run jsdeps`.

- I named the script `jsdeps`, but any better name suggestion is appreciated.

- Running `npm run jsdeps`, a specified list of files is copied to `src/fontra/client/third-party/` folder.

- The list of files mapping is declared in external `jsdeps.yml` config file.

- `bezier-js` has already been installed as node module and its sources gets copied to the `third-party/` folder when the script is executed.

- The `third-party/` is rebuilt everytime the script is executed, for removing potentially uninstalled modules files.

- The `third-party/` directory has been excluded from `prettier` `pre-commit` hook, and I think that we should probably also add it to `.gitignore`.


justvanrossum:
Thanks! I'm traveling now, so can't review in-depth.

Yes, ideally third-parte should be empty in the repo and listen in .gitignore.

We're using hatch/hatchling to build (pip will use that via pyproject.toml), perhaps we can use a build hook to automate this: https://hatch.pypa.io/latest/config/build/

fabiocaccamo:
> We're using hatch/hatchling to build (pip will use that via pyproject.toml), perhaps we can use a build hook to automate this: https://hatch.pypa.io/latest/config/build/

True, but the command must also be run manually for development (when installing new node modules), not just at build time.

justvanrossum:
My main question currently is: why do we need an config file? Can't the info from scripts/jsdeps.yml be inferred?

fabiocaccamo:
Having a config file allow us to copy only necessary files, alternatively we can avoid the config file at the price of copying the whole module folder to the `third-party/` directory.

justvanrossum:
How do you decide what is a "neccesary file"?

> the whole module folder

Do you mean `node_modules`? But that also includes dev-dependencies, no?

fabiocaccamo:
> How do you decide what is a "neccesary file"?

Inside each node module there are a lot of files that we don't need for sure, the decision has to be taken differently for each third-party module, alternatively we could have automated it.
Let's take `bezier.js` module as an example: we need only the 3 files inside the `src` directory, but the whole module contains many other files (this is not a bad case, but there could be much more files that we don't need at all):
Screenshot 2023-03-09 at 10 02 46

> Do you mean node_modules?

No I mean specific node modules, eg `bezier.js`.

requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
 pre-commit==3.1.0
 pytest==7.2.1
 pytest-asyncio==0.20.3
+PyYAML==6.0
```
justvanrossum, line 4: I believe this dependency is now redundant.

justvanrossum:
I'm coming back to the build hook.

With the jsdeps.py mechanism is place, fontra can only be run from the source when jsdeps.py is executed first.

In other words, `pip install -e .` is no longer enough to get fontra working. Same when building a wheel: the fontra wheel should contain the populated third-party folder.

I think hatch offers a good way to make running jsdeps.py part of the install/build process: https://hatch.pypa.io/latest/plugins/build-hook/reference/ 
We could use the "custom" hook, which ay be lower-effort than writing a true plug-in. The `initialize` hook sounds like it will allow us to run code before hatch does the build.

scripts/jsdeps.py
```diff
@@ -0,0 +1,38 @@
+import json
+import shutil
+from pathlib import Path
+
+DEST_BASE_DIR = "src/fontra/client/third-party/"
```
justvanrossum, line 5: We should probably find the root of the repo explicitly so this script doesn't depend on the repo root being the current working directory.

I tend to use a pattern like this:
```python
repoRoot = pathlib.Path(__file__).resolve().parent.parent
destDir = repoRoot / "src/fontra/client/third-party/"
```

This also avoids calling `Path()` all the time locally.

scripts/jsdeps_build_hook.py
```diff
@@ -0,0 +1,9 @@
+import subprocess
+
+from hatchling.builders.hooks.plugin.interface import BuildHookInterface
+
+
+class JSDepsBuildHook(BuildHookInterface):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        subprocess.check_output("python -I scripts/jsdeps.py", shell=True)
```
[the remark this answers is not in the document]
justvanrossum: https://hatch.pypa.io/latest/plugins/build-hook/reference/#hatchling.builders.hooks.plugin.interface.BuildHookInterface.initialize

scripts/jsdeps_build_hook.py
```diff
@@ -0,0 +1,9 @@
+import subprocess
+
+from hatchling.builders.hooks.plugin.interface import BuildHookInterface
+
+
+class JSDepsBuildHook(BuildHookInterface):
+    def initialize(self, version, build_data):
+        super().initialize(version, build_data)
```
justvanrossum, line 8:
```suggestion
```
I checked, there's no need to call super.

scripts/jsdeps.py
```diff
@@ -0,0 +1,44 @@
+import json
+import shutil
+from pathlib import Path
+
+
+def get_dest_dir():
```
justvanrossum, line 6: Please just use global variables, as I suggested.

justvanrossum:
General remark for jsdeps.py: please use camelCase, as that's the predominant style in this repo.

fabiocaccamo:
> General remark for jsdeps.py: please use camelCase, as that's the predominant style in this repo.

Isn't `snake_case` the standard style?
https://peps.python.org/pep-0008/#function-and-variable-names

justvanrossum:
> Isn't snake_case the standard style?

Yes, but it's not in this project. One reason is to keep sane between JS and Python: there's lots of stuff in common that would be annoying if the names were styled differently.

.github/workflows/test.yml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Install Node dependencies
+        run: |
+          npm install
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install Python dependencies
```
justvanrossum, line 37: The python dependencies should not be needed in the npm job.

.github/workflows/test.yml
```diff
@@ -0,0 +1,86 @@
...
+          pre-commit run --all-files --verbose --show-diff-on-failure
+
+      - name: Run Node tests
+        run: |
+          npm ci
+          npm run jsdeps
+          npm run build --if-present
+          npm test
+
+  run-py-test:
```
justvanrossum, line 55: Please add a "needs" dependency on the run-npm-test job:
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

This can (should?) be done before the runs-on info.

This ensures that we only run pytest after the npm tests succeed. It's mostly important for the pre-commit step: we don't want to start pytest before knowing pre-commit was successful.

scripts/jsdeps.py
```diff
@@ -0,0 +1,39 @@
...
+from pathlib import Path
+
+repoRoot = Path(__file__).resolve().parent.parent
+destDir = repoRoot / "src/fontra/client/third-party/"
+
+
+def loadPackageDependencies():
+    with open("package.json") as packageFile:
+        packageInfo = json.load(packageFile)
+        dependencies = packageInfo.get("dependencies")
```
justvanrossum, line 12: This can be simplified: 
- from this line on it doesn't need to be within the `with` block
- Use `{}` as the fallback in `.get("dependencies")`, which allows you to get rid of the `if`, and the fallback return statement.

scripts/jsdeps.py
```diff
@@ -0,0 +1,39 @@
...
+
+def processDependencies(dependencies):
+    if destDir.is_dir():
+        shutil.rmtree(destDir)
+    for dependency in dependencies:
+        processDependency(name=dependency)
+
+
+def processDependency(name):
+    src = Path("node_modules") / name
```
justvanrossum, line 26: You could add another global const:
`nodeModules = repoRoot /  "node_modules"`
This ensures it will work correctly regardless of the current working dir.

scripts/jsdeps_build_hook.py
```diff
@@ -0,0 +1,9 @@
+import subprocess
+
+from hatchling.builders.hooks.plugin.interface import BuildHookInterface
+
+
+class JSDepsBuildHook(BuildHookInterface):
+    def initialize(self, version, build_data):
```
justvanrossum, line 7: I wonder if hatch guarantees that the current working dir is the repo root at this point. If not, this will do the wrong thing. It would be nice to try something like this:

- make new dir somewhere
- clone fontra repo here
- set up Python venv here
- then `pip install ./fontra/`

If that all works, we're good.

justvanrossum: If not, we may want to specify the repo root dir explicitly through the `cwd` keyword arg of `subprocess.check_output()`.

.github/workflows/test.yml
fabiocaccamo: We run `pre-commit`, but we can run it in the `Python` job.

justvanrossum: Ah ok, let me think. Would it be an idea to indeed run in it the Python job, but reverse the order?

justvanrossum: So that the npm job "needs" the pytest job.

README.md
```diff
@@ -45,6 +45,8 @@ Fontra Pak should also work on Linux and other platforms, but you will have to b
 
   `npm install`
 
+  `npm run jsdeps`
```
justvanrossum, line 48: With the build hook in place, is this even needed?

justvanrossum: Instead, we need to mention that Node needs to be installed, probably right after "Ensure you have Python >= 3.10 installed"

fabiocaccamo: I already removed it, I think you looked an old version of the workflow.

fabiocaccamo: I update the README.

justvanrossum: (My comments are indeed about the README)

fabiocaccamo: Ah ok, sorry again :)

package.json
```diff
@@ -1,10 +1,14 @@
 {
   "scripts": {
+    "jsdeps": "python -I scripts/jsdeps.py",
```
justvanrossum, line 3: I wonder if we should get rid of this. We're not using it anymore, and solely rely on the hatch build hook to do this.

fabiocaccamo: It could be useful for syncing `third-party` folder when installing new node modules.

Even if the same result can be achieved by running directly the `Python` script, having it defined in `package.json` is straightforward for anyone looking what `npm` scripts are available. 
*(personally, I always look inside `package.json` to see what scripts can be run)*

Let's consider the simplicity: 
- `npm install xxx --save`
- `npm run jsdeps`

than:
- `npm install xxx --save`
- `python -I scripts/jsdeps.py`

justvanrossum: Sure ok, we can keep it.

But if we update anything in requirements.txt we also have to re-run `pip install -e .`, and updating the js deps for Fontra is more similar to that in my view.

fabiocaccamo: Yes, it's bidirectional.

README.md
```diff
@@ -39,22 +45,26 @@ Fontra Pak should also work on Linux and other platforms, but you will have to b
 
   `pip install -e .`
 
+#### Testing
+
+The following steps are needed only for contributors:
```
justvanrossum, line 50:
```suggestion
The following steps are needed if you want to run the automated tests and if you want to contribute:
```

scripts/jsdeps_build_hook.py
justvanrossum: I tried this and it seems to work correctly, so no further work is needed here.

justvanrossum (approved):

PR #336, opened 2023-03-07
Add Settings panel to the sidebar

### LINKED ISSUE #316: Make preference panel by enabling and populating the "gear" left-sidebar tab item
`<!-- <div class="sidebar-tab" data-sidebar-name="settings">gear</div> -->`

- [x] Populate it with the theme radio switch from settings.html
- [x] Add a preference item for "clipboard export format". Maybe a popup menu. Options: "SVG", "GLIF (RoboFont)", "JSON (Fontra)".

nedkamburov:
Fixes #316 
Separated settings in their own web component. `<general-settings/>`
My idea is that this can hold all logic which can be further broken down into modules/components. 

*`themeOptions` and `clipboardFormatOptions` are pretty much the same, they can be refactored to use less code but I think it is more convenient for altering them this way while still in the same component.

Screenshot 2023-03-07 at 13 04 04

### DIFF
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -273,6 +273,7 @@ export class EditorController {
     await this.fontController.subscribeChanges(rootSubscriptionPattern, false);
     await this.initGlyphNames();
     await this.initSliders();
+    this.initSettings();
     this.initTools();
     this.initSourcesList();
     await this.setupFromWindowLocation();
@@ -342,6 +343,11 @@ export class EditorController {
         this.autoViewBox = false;
       })
     );
+  }
+
+  initSettings() {
+    const settingsTab = document.querySelector("#settings");
+    settingsTab.innerText = "Hello from editor.js";
   }
 
   initTools() {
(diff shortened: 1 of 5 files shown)


src/fontra/views/editor/editor.js
```diff
@@ -933,8 +941,9 @@ export class EditorController {
       return;
     }
 
-    const preferGLIF = true; // TODO should be user preference
-
+    const preferGLIF =
```
justvanrossum, line 944: Given the multi-state options UI, this should also reflect that. We should probably use a string, and name the variable "clipboardExportFormat". Values: "svg", "glif", "fontra-json". Perhaps the values could be literally the value in localStorage.

src/fontra/views/editor/editor.js
```diff
@@ -344,6 +346,12 @@ export class EditorController {
     );
   }
 
+  initSettings() {
```
justvanrossum, line 349: I suppose this method isn't really needed?

PR #337, opened 2023-03-07
[Sidebar Lists] Add and Remove buttons for the sources list

### LINKED ISSUE #323: [ui] Sources list in left sidebar should only be as tall as needed
It currently extends all the way to the bottom of the left sidebar, but it should only be as tall as needed.

### LINKED ISSUE #324: [ui] Add [ + ] and [ - ] buttons at the bottom the the sources list
These buttons will mean "add source" and "delete source".

Ideally, the latter should only be enabled if there is a selection.

Ideally, typing "delete" in the list will trigger the same action.

I think these items should be separate from the list element. They should probably be implemented as a web component.

This needs #323 to be fixed first.

Do not worry about the actual "add source" and "delete source" implementation, this is just for the UI element(s).

nedkamburov:
Fixes #323
Fixes #324 

Add a `add-remove-buttons` web component that is dynamically inserted when a List is configured. It is an optional include.
– Disabling of the button assumes that the default source cannot be deleted, anything else can be (as long as there is more than one). 
_I see that my logic doesn't cover correctly that case where you open the source list and a source different from the default is selected. I need a bit of clarification on that one._

– The implementation stub is left in the editor.js `initSourcesList() `with a //TODO comment
– The list is also adaptive to its contents now

*The type 'delete' to remove a source part I didn't get fully so it's not implemented yet.

Screenshot 2023-03-07 at 16 34 57
Screenshot 2023-03-07 at 16 36 48

### DIFF
--- a/src/fontra/client/core/ui-list.js
+++ b/src/fontra/client/core/ui-list.js
@@ -62,6 +62,14 @@ export class List {
 
     addRemoveButtonsElement.addButtonCallback = addButtonCallback;
     addRemoveButtonsElement.removeButtonCallback = removeButtonCallback;
+    addRemoveButtonsElement.disableRemoveButton =
+      !this.selectedItemIndex || this.selectedItemIndex === 0;
+
+    this.container.addEventListener("listSelectionChanged", () => {
+      addRemoveButtonsElement.disableRemoveButton =
+        !this.selectedItemIndex || this.selectedItemIndex === 0;
+    });
+
     this.container.appendChild(addRemoveButtonsElement);
   }
 
--- a/src/fontra/client/core/ui-list.js
+++ b/src/fontra/client/core/ui-list.js
@@ -1,3 +1,4 @@
+import { addRemoveButtons } from "../web-components/add-remove-buttons.js";
 const LIST_CHUNK_SIZE = 200; // the amount of items added to the list at a time
 
 export class List {
@@ -26,6 +27,7 @@ export class List {
     this.contents = document.createElement("div");
     this.contents.className = "contents";
     this.container.appendChild(this.contents);
+
     this.contents.addEventListener(
       "click",
       (event) => this._clickHandler(event),
@@ -53,6 +55,14 @@ export class List {
     );
     this.selectedItemIndex = undefined;
     this.container.classList.add("empty");
+  }
+
+  includeAddRemoveButtons(addButtonCallback, removeButtonCallback) {
+    const addRemoveButtonsElement = new addRemoveButtons();
+
+    addRemoveButtonsElement.addButtonCallback = addButtonCallback;
+    addRemoveButtonsElement.removeButtonCallback = removeButtonCallback;
+    this.container.appendChild(addRemoveButtonsElement);
   }
 
   setItems(items) {
--- a/src/fontra/client/core/ui-list.js
+++ b/src/fontra/client/core/ui-list.js
@@ -26,7 +26,6 @@ export class List {
     this.contents = document.createElement("div");
     this.contents.className = "contents";
     this.container.appendChild(this.contents);
-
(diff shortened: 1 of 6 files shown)


src/fontra/views/editor/editor.html
```diff
@@ -268,6 +268,10 @@
       padding-right: 0.8em;
     }
 
+    .ui-list {
```
justvanrossum, line 271: If we change it to `sources-list`, it won't affect the glyph search results.
```suggestion
    .sources-list {
```
I'm not convinced we should set this globally for ui-list.

src/fontra/client/web-components/add-remove-buttons.js
```diff
@@ -0,0 +1,46 @@
...
+    super();
+    this.addButtonCallback = () => {};
+    this.removeButtonCallback = () => {};
+    this.disableRemoveButton = false;
+  }
+
+  render() {
+    return html`
+      <div class="buttons-container">
+        <button name="add-button" @click=${() => this.addButtonCallback()}>+</button>
```
justvanrossum, line 33: Can you show what Lit is adding to this Web Component apart from the `@click` syntax? Because if that's all, we might as well use `onclick=...`

src/fontra/client/core/ui-list.js
```diff
@@ -55,6 +57,21 @@ export class List {
     this.container.classList.add("empty");
   }
 
+  includeAddRemoveButtons(addButtonCallback, removeButtonCallback) {
```
justvanrossum, line 60: I'm not sure this should be part of the ui-list module. Let's discuss live.

nedkamburov: I moved it to the editor.js to be configured all there

src/fontra/client/core/ui-list.js
```diff
@@ -26,6 +26,7 @@ export class List {
     this.contents = document.createElement("div");
     this.contents.className = "contents";
     this.container.appendChild(this.contents);
+
```
justvanrossum, line 29:
```suggestion
```

src/fontra/views/editor/editor.js
```diff
@@ -384,6 +386,20 @@ export class EditorController {
...
     this.sourcesList = new List("sources-list", columnDescriptions);
+    // TODO: relocate those to somewhere more appropriate after implementation
+    const addSourceCallback = () => {
+      console.log("add a source");
+    };
+    const removeSourceCallback = () => {
+      console.log("remove a source");
+    };
+
+    this.includeAddRemoveButtons(
```
justvanrossum, line 397: Suggest to inline the `includeAddRemoveButtons` method, and add its "listSelectionChanged" logic to the existing event listener.

src/fontra/client/web-components/add-remove-buttons.js
```diff
@@ -0,0 +1,47 @@
+import { html, css, LitElement } from "https://cdn.jsdelivr.net/npm/lit@2.6.1/+esm";
+
+export class addRemoveButtons extends LitElement {
```
justvanrossum, line 3: `addRemoveButtons` is a class name and should be capitalized: `AddRemoveButtons`

src/fontra/views/editor/editor.js
nedkamburov: What if we want to use it for more lists though? That was my intent initially.

justvanrossum: It is not a lot of code and currently actually contains some code that is specific to the sources list.

justvanrossum: That method already contains logic that is specific to the sources list. We currently don't want to use it for other lists, so that's a bit of a premature generalization.

PR #338, opened 2023-03-08
'Add Component' popup and context menu option

### LINKED ISSUE #319: Make UI panel to add a new component to a glyph
- [x] Contextual menu "Add Component"
- [ ] Show a panel with from top to bottom:
  - ( search field )
  - selectable list of glyph names, filtered by search field, unless search is empty
  - ( Cancel ) ( Add )

Base the looks/experience on ui-dialog.js, later we can see if things can be generalized to share code.

Don't worry about actually adding the component yet.

nedkamburov:
Fixes  #319
**This is a WIP**. I have implemented the UI but 2 questions remain:
1. Do we keep the glyphList as part of the editor and we pass the data it renders to the component? Or we encapsulate the component so that it works on its own? The need for refactoring is quite big in that case...
2. How is the styling supposed to work? Do we duplicate styles or we import them so they can apply to the component and normal DOM elements?

I opted for using a web component but if this adds a great need for refactoring, we can simplify it. We just need to discuss the aforementioned difficulties.

Visually it looks like this:
Screenshot 2023-03-08 at 16 22 25

### DIFF
--- a/src/fontra/client/core/ui-dialog.js
+++ b/src/fontra/client/core/ui-dialog.js
@@ -76,7 +76,6 @@ export function dialog(headline, message, buttonDefs, autoDismissTimeout) {
 
   /* prevent clicks on the dialog itself to propagate to the container */
   content.onclick = (event) => event.stopImmediatePropagation();
-
   if (autoDismissTimeout) {
     dismissTimeoutID = setTimeout(() => dialogDone(null), autoDismissTimeout);
   }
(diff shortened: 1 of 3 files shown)


justvanrossum:
Please adjust to use ui-dialog.js' `dialog()` function, with the new `messageOrContentFunction` argument. This should allow you to remove all the duplicated code for the dialog itself.

justvanrossum:
Example of how to use this new `dialog()` functionality:
```js
    const addSourceCallback = async () => {
      const el = htmlToElement(`
        <div>
          <div>Heyyy tralala</div>
          <br>
          <input type="checkbox" name="foo" id="foo">
          <label for="foo">Check this<label>
        </div>`);
      const result = await dialog("Testing source", () => el, [
        { title: "Cancel", isCancelButton: true },
        { title: "Ok", getResult: () => "rara", isDefaultButton: true },
      ]);
      console.log("add a source", result);
    };
```

Issue #193, opened 2022-11-24
Need a slider-like control for discrete designspace axes


justvanrossum:
- This will be needed to support DesignSpace-5-style discrete axes (#67).
- Should be _visually_ related to #177. Maybe also technically.
- Instead of a continuous slider, the control should have a set of arbitrary stops along the scale, where the thumb should snap.
- It also needs "numeric" input in the form of a pop-up that shows all choices.
- While the number of stops is arbitrary, it will typically not be a huge number.
- I expect the common case (by far) to be only have states: the min and max value, so it would effectively become an on/off button.
- It should work well with 3, 4 or 5 values (including min and max), but I think compromises are ok if the number gets higher.

justvanrossum:
This was implemented in #972.

justvanrossum:
This was implemented in #972.

PR #340, opened 2023-03-09
Web component range sliders

### LINKED ISSUE #177: Create better slider behavior and styling for designspace axes
This is both for the "viewing" designspace settings, as for variable component axis values.

Requirements:
- Compact (within reason)
- Shows min/default/max values numerically and visually
- Shows the current value numerically
- Current value should be editable numerically, with some digits of floating point precision
- Numeric values *could* be visible only upon hover

Nice to have:
- A button-ish control to make the slider snap back to the default value

Observations:
- horizontal space is somewhat more expensive than vertical space, but compactness is beneficial in both directions.

Initially, let's prototype as a single html file, containing all CSS + JS.

TODO:

- [x] Prototype
- [ ] Integrate into Fontra

@fabiocaccamo

nedkamburov:
Fixes #177 
Another WIP. A web component range slider is added. It is basically configured similarly to native range sliders.
Things to add still:
- [x] add tickmarks (stops) dynamically
- [x] change callback (actually affect the glyph)
- [x] better styling, responding to the global themes

Current designs:
Screenshot 2023-03-09 at 18 23 07
Screenshot 2023-03-09 at 18 23 18


justvanrossum:
We also need to see the min/max/default values, possibly only on hover (or as a tool tip)

A tickmark at the default position may be useful. If you're thinking of the "discrete values slider", then that is a candidate for tickmarks, too.

I really liked that "reset" button you had in your earlier demo.

nedkamburov:
The reset button is there between the value and the slider. Maybe not the best position for now but it's there.
Screenshot 2023-03-14 at 10 05 22

> tickmark at the default position 
you mean a visual exaggeration, e.g. thicker line

> "discrete values slider",
I am not sure what you mean with this

justvanrossum:
> discrete values slider 

https://github.com/googlefonts/fontra/issues/193

> you mean a visual exaggeration, e.g. thicker line

Apart from the "discrete values" use case, I don't think we want tick marks at all, except possibly a single one at the default position.

nedkamburov:
That's optional now. IF you provide a tickmark array, it will show it.

justvanrossum:
I don't think the label should be part of the slider component. How the label sits in the layout should be determined by the parent container.

nedkamburov:
Okay I can take it out, I just imagined it will easier to configure. 
In case you're concerned with space and layout, I can also create several 'layouts' that can be triggered with a class?
Something like 'compact', 'expanded', etc.

src/fontra/client/core/ui-sliders.js
```diff
@@ -24,20 +26,19 @@ export class Sliders {
...
-        slider.max = sliderInfo.maxValue;
-        slider.value = sliderInfo.defaultValue;
-        slider.dataset.name = sliderInfo.name;
-        slider.oninput = (event) => this._dispatchListSelectionChanged();
-        label.appendChild(slider);
-        label.append(sliderInfo.name);
-        this.container.appendChild(label);
+        //TODO: make this dynamic also
+        slider.tickMarksPositions = [0, 100, 200, 300, 500, 700, 900];
+        slider.onChangeCallback = (event) => this._dispatchListSelectionChanged(); // TODO: not sure if that works
```
justvanrossum, line 38: Ah, that method is wrongly named. We should probably rename it to `_dispatchSlidersChangedEvent`

src/fontra/client/web-components/range-slider.js
```diff
@@ -0,0 +1,273 @@
...
+    }
+    this.onChangeCallback(defaultValue);
+  }
+
+  reset() {
+    this.defaultValue = this.midValue;
+    this.onChangeCallback(this.defaultValue);
+  }
+
+  get midValue() {
```
justvanrossum, line 263: "midValue" is not a useful concept. We have an explicit defaultValue, and if it's not given, perhaps it should fall back to the minValue.

nedkamburov: The naming is misleading here: midValue is used to check against the current value of the slider. 
I can rename midValue to defaultValue 
and defaultValue to currentValue?

justvanrossum: It is literally calculating the middle between min and max in the line below.

nedkamburov: Yes, this function is pointless. I meant that it does serve a purpose. But anyways, you can see how it should work after the last commit.

src/fontra/client/web-components/range-slider.js
```diff
@@ -0,0 +1,267 @@
...
+      opacity: 0.65;
+      background: dimgray;
+    }
+  `;
+
+  static properties = {
+    name: { type: String },
+    minValue: { type: Number },
+    maxValue: { type: Number },
+    defaultValue: { state: true },
```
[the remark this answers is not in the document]
nedkamburov: `state: true` I was thinking it's needed in order to make this property reactive but it isn't. I will remove it.

nedkamburov:
After a few iterations of the slider, we have ended up with this:
Screenshot 2023-03-23 at 16 37 36

Actual design that is slightly different:
Screenshot 2023-03-23 at 16 35 44

PR #352, opened 2023-03-21
Third party libraries bundle

### LINKED ISSUE #246: Figure out how to deploy npm dependencies

### LINKED ISSUE #343: Lit should now become a true dependency
TODO:
- [x] Install Lit locally, update package.json etc.
- [ ] Change Lit import in web-components/general-settings.js to the appropriate local path
- [ ] Test that it actually works

nedkamburov:
Fixes #246 and partly fixes #343 
Bundle all third party libraries into separate chunks and use them as standard ES6 Modules

TODO:
- [x] Decide whether the rollup-bundle script should run for all dependencies automatically or they should be listed manually?


src/fontra/client/web-components/general-settings.js
```diff
@@ -1,4 +1,4 @@
-import { html, css, LitElement } from "https://cdn.jsdelivr.net/npm/lit@2.6.1/+esm";
+import {  html, css, LitElement } from "../third-party/lit.js";
```
fabiocaccamo, line 1: Prettier has not formatted the double space correctly?

nedkamburov: I just saw that. Thanks for catching it!
It should work now.

justvanrossum:
Thanks!

The CI fails probably because rollup isn't run as part of the install. This should be done in jsdeps_build_hook.py. At the same time, `jsdeps.py` should be removed, and `jsdeps_build_hook.py` should be renamed to `rollup_build_hook.js`.

nedkamburov:
Ahh, so that's where it is, I will rearrange them. Thanks!

justvanrossum:
You still need to run `subprocess.check_output("npm install", shell=True)` from the build hook.

nedkamburov:
Can I just do this than:
```
class RollupBuildHook(BuildHookInterface):
    def initialize(self, version, build_data):
        subprocess.check_output("npm install", shell=True)
        subprocess.check_output("npm run bundle-rollup", shell=True)
```

justvanrossum:
Very cool. Ready for review? This is now no longer a draft, right?

justvanrossum:
> Decide whether the rollup-bundle script should run for all dependencies automatically or they should be listed manually?

Can you elaborate on what you mean by this?

fabiocaccamo:
> If we want we can automate this and supply it with an array of all non-dev dependencies so we don't have to deal with this every time we add a dependency.

I think that considering how often new dependencies are added, managing it manually is not a pain and give us more explicit control.

---

> Additionally: what happens if we use two different npm packages, but they have a shared dependency? Does the shared dependency gets rolled up twice, or will it remain to be shared?

I think that they will be rolled up twice, I don't know if it is possible, but maybe that it's possible to bundle them with a unique `rollup` call, in this case probably the dependencies would be not rolled up twice.

justvanrossum:
1. According to https://rollupjs.org/command-line-interface/#configuration-files you can be explicit about the output file name

> We can bundle all code inlibraries.js and pull what we need from there

I'm curious about that option. I'm playing around, but I can't find how to configure rolling multiple inputs into a single library.js. Do you know the magic?

nedkamburov:
The output option can allow you to specify a name of the bundled file. You need to provide it with an entry point in the input though. If the input file, e.g. `index.js` has imports it will bundle all of them (with it's node deps) into a `output.js`

That way it work something like this
```
export default [
    {
        input: 'src/third-party/index.js',
        plugins: [commonjs()],
        output: {
            name: 'bundle',
        }
    }
]
```

I used this to get all dependencies but they get bundled up as separate chunks and I couldn't find a way to specify their names easily (not manually)
```
import nodeResolve from "@rollup/plugin-node-resolve";
import commonjs from "@rollup/plugin-commonjs";
const dependencies = require("./package.json").dependencies;

export default {
  input: Object.keys(dependencies).map((dep) => `node_modules/${dep}`),
  output: {
    dir: "src/fontra/client/third-party",
    format: "esm",
    // manualChunks: { // This doesn't work as expected and it is still manual
    // "bezier-js": ["bezier-js"],
      // "lit": ["lit"],
    // },
  },
  plugins: [
    commonjs(), // <-- this handles some parsing of js syntax or something (necessary for `export { init } from "mathjax";`)
    nodeResolve(), // <-- this allows npm modules to be added to bundle
  ],
};
```

justvanrossum:
This does something:

```js
import nodeResolve from "@rollup/plugin-node-resolve";
import commonjs from "@rollup/plugin-commonjs";
import fs from "fs";

const packageInfo = JSON.parse(fs.readFileSync("package.json", "utf8"));
const depNames = [...Object.keys(packageInfo["dependencies"])];

export default depNames.map((depName) => {
  return {
    input: `node_modules/${depName}`,
    output: {
      file: `src/fontra/client/third-party/${depName}.js`,
      format: "cjs",
    },
    plugins: [
      commonjs(), // <-- this handles some parsing of js syntax or something (necessary for `export { init } from "mathjax";`)
      nodeResolve({ browser: true }), // <-- this allows npm modules to be added to bundle
    ],
  };
});
```

@belluzj: thanks, that makes sense! We could perhaps do that dynamically, too.

nedkamburov:
Neat! I haven't thought of that. Let's go with it then.

belluzj:
Usually this happens naturally, because your dependencies are required by your app's scripts, so usually the tree/graph of imports from your app's entry point to the various dependencies from your package.json is there across all the app's files importing each other and ultimately importing third-party code. Rollup and others should be able to follow this tree of imports without you needing to hold their hand by feeding them the dependencies from package.json by hand, if that makes sense?

That's also why Rollup and others advertise "tree shaking". E.g. if you put "lodash" in your package.json, but your code only says import { oneFunction } from "lodash", in theory Rollup should figure out that it only needs to put in your app bundle that one function. You may be giving up on that optimization by listing the full contents of your package.json yourself

belluzj:
Sorry I'm commenting on this without having participated to the effort :) I'm not aware of the particulars

justvanrossum:
> Sorry I'm commenting on this without having participated to the effort :) I'm not aware of the particulars

No problem, I'm happy that you are having a look!

Fontra's JS structure is not quite standard (I think), (and as I think you've noticed during the Electronize attempt), due to the use of a server that turns disconnected subtrees into a connected "virtual" tree that the browser sees. So we don't have a 1-on-1 mapping from file system to asset URL.

justvanrossum:
> So we don't have a 1-on-1 mapping from file system to asset URL.

A future change could be that we add a build step that collects the disconnected trees and copies them to a a temp folder that *does* have a 1-on-1 mapping to the URL scheme. That would probably make things easier in the long run.

justvanrossum (approved):
LGTM, thanks!

Issue #510, opened 2023-06-07
[unittests] Write unittests for utils.js


justvanrossum:
Several function in utils.js that can be tested are not. It would be nice if the ones that are easily testable (without mocking broswer stuff) are acually test.

We should create a file `test-js/test-utils.js`, and follow the patterns used in the other test files, using `chai` for the "expect" assertions.

The (easily) testable function are these (in order from the file, which is rather arbitrary):

- [x] `objectsEqual()`
- [x] `capitalizeFirstLetter()`
- [x] `hyphenatedToCamelCase()`
- [x] `modulo()`
- [ ] ~`sign()`~
- [x] `boolInt()`
- [x] `reversed()`
- [x] `enumerate()`
- [x] `reversedEnumerate()`
- [x] `range()`
- [x] `chain()`
- [x] `makeUPlusStringFromCodePoint()`
- [x] `getCharFromUnicode()`
- [x] `guessCharFromGlyphName()`
- [x] `fileNameExtension()`
- [x] `arrayExtend()`

We should not forget to test exception cases: `expect(...).to.throw(...)`.

Ideally we'd use a data-driven approach, similar to how `test-cmap.js` is written.

fatih-erikli:
Btw there is a `sign` method in JS's `Math`, `(Math.sign)` for a sign function. Is there a reason it's implemented explicitly in utils? @justvanrossum

justvanrossum:
> Btw there is a sign method in JS's Math, (Math.sign) for a sign function. Is there a reason it's implemented explicitly in utils?

Ah, good catch: this is a simple oversight. We should remove it from utils and use Math.sign() instead. (It strangely enough doesn't exist in Python's math module, so maybe that's why I failed to even look for it in JS.)

justvanrossum:
Even worse: there is a second `sign()` implementation in `var-model.js`...

justvanrossum:
This was implemented in #533.

PR #533, opened 2023-06-10
Tests for utils

fatih-erikli:
Initial PR for the test cases of `utils.js`.

### DIFF
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -267,7 +267,7 @@ export function guessCharFromGlyphName(glyphName) {
   // Search for a 4-6 char hex string in the glyph name. If 6, the first digit must
   // be one. Interpret the hex string as a unicode code point and convert to a
   // character. Else, return an empty string.
-  const match = glyphName.match(/(^|[^0-9A-F])(1?[0-9A-F]{4,5})($|[^0-9A-F])/);
+  const match = glyphName.match(/(^|[^0-9A-F])([0-9A-F]{4,5})($|[^0-9A-F])/);
   return match ? String.fromCodePoint(parseInt(match[2], 16)) : "";
 }
 
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -264,7 +264,7 @@ export function getCharFromUnicode(codePoint) {
 }
 
 export function guessCharFromGlyphName(glyphName) {
-  // Search for a 4-6 char hex string in the glyph name.
+  // Search for a 4-5 char hex string in the glyph name.
   // Interpret the hex string as a unicode code point and convert to a
   // character. Else, return an empty string.
   const match = glyphName.match(/(^|[^0-9A-F])([0-9A-F]{4,5})($|[^0-9A-F])/);
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -264,8 +264,8 @@ export function getCharFromUnicode(codePoint) {
 }
 
 export function guessCharFromGlyphName(glyphName) {
-  // Search for a 4-6 char hex string in the glyph name. If 6, the first digit must
-  // be one. Interpret the hex string as a unicode code point and convert to a
+  // Search for a 4-6 char hex string in the glyph name.
+  // Interpret the hex string as a unicode code point and convert to a
   // character. Else, return an empty string.
   const match = glyphName.match(/(^|[^0-9A-F])([0-9A-F]{4,5})($|[^0-9A-F])/);
   return match ? String.fromCodePoint(parseInt(match[2], 16)) : "";
--- a/test-js/test-utils.js
+++ b/test-js/test-utils.js
@@ -1,0 +1,213 @@
+import chai from "chai";
+import {
+  objectsEqual,
+  capitalizeFirstLetter,
+  hyphenatedToCamelCase,
+  modulo,
+  boolInt,
+  reversed,
+  enumerate,
+  reversedEnumerate,
+  range,
+  chain,
+  makeUPlusStringFromCodePoint,
+  getCharFromUnicode,
+  guessCharFromGlyphName,
+  fileNameExtension,
+  arrayExtend,
+} from "../src/fontra/client/core/utils.js";
+const expect = chai.expect;
+
+describe("objectsEquals", () => {
+  it("falsy values", () => {
+    expect(objectsEqual(null, null)).equals(true);
+    expect(objectsEqual(undefined, null)).equals(false);
+    expect(objectsEqual("", "")).equals(true);
+    expect(objectsEqual("", null)).equals(false);
+  });
+  it("with things inside", () => {
+    expect(objectsEqual({}, { a: 1 })).equals(false);
+    expect(objectsEqual({ a: 1 }, {})).equals(false);
+    expect(objectsEqual({ a: 1 }, { a: 2 })).equals(false);
+    expect(objectsEqual({ a: 2 }, { a: 2 })).equals(true);
+  });
+  it("thing in their prototype", () => {
+    // making sure only objects own properties checked
+    class Thing {
+      constructor(value) {
+        this.b = value;
+      }
+    }
+    Thing.prototype.a = 1;
+    expect(objectsEqual(new Thing(1), { b: 1 })).equals(true);
+    expect(objectsEqual(new Thing(1), { b: 1, a: 1 })).equals(false);
+  });
+});
+
+describe("capitalizeFirstLetter", () => {
+  it("basic functionality", () => {
+    expect(capitalizeFirstLetter("sam")).equals("Sam");
+    expect(capitalizeFirstLetter("Sam")).equals("Sam");
+  });
+  it("with spaces prefixed", () => {
+    expect(capitalizeFirstLetter(" sam")).equals(" sam");
+  });
+});
+
+describe("hyphenatedToCamelCase", () => {
+  it("should camelize", () => {
+    expect(hyphenatedToCamelCase("test-case")).equals("testCase");
+  });
+  it("should not delete the hypen when the second part is not a lowercase letter", () => {
+    expect(hyphenatedToCamelCase("test-1")).equals("test-1");
+  });
+});
+
+describe("modulo", () => {
+  it("should return the remaining when divide", () => {
+    expect(modulo(12, 5)).equals(2);
+  });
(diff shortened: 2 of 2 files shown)


fatih-erikli:
#510 @justvanrossum

test-js/test-utils.js
```diff
@@ -0,0 +1,247 @@
...
+    expect(numbersreversed.next().value).equals(2);
+    expect(numbersreversed.next().value).equals(1);
+  });
+});
+
+describe("enumerate", () => {
+  it("enumerate an array, enumeration start with 0 by default", () => {
+    const numbers = [1, 2, 3];
+    const numbersenumerated = enumerate(numbers);
+    expect(numbersenumerated.next().value[0]).equals(0);
```
justvanrossum, line 105: This doesn't test whether enumerate returns the correct _value_, only the index. Unpacking to an array should help, as you can then compare against `[[0, 1], [1, 2], [2, 3]]` using `expect(...).to.deep.equal(...)`.

test-js/test-utils.js
```diff
@@ -0,0 +1,247 @@
...
+    {
+      const iteratorResult = numbers.next();
+      expect(iteratorResult.value).equals(14);
+    }
+  });
+});
+
+describe("chain", () => {
+  it("chain iterators", () => {
+    function* numbers() {
```
justvanrossum, line 190: You could also use `range()` here, as that returns an iterator, too.

test-js/test-utils.js
```diff
@@ -0,0 +1,247 @@
...
+    const chained = chain(numbers(), numbers());
+    expect(chained.next().value).equals(1);
+    expect(chained.next().value).equals(2);
+    expect(chained.next().value).equals(1);
+    expect(chained.next().value).equals(2);
+  });
+});
+
+describe("makeUPlusStringFromCodePoint", () => {
+  it("throw an exception when a not valid parameter given", () => {
```
justvanrossum, line 203:
```suggestion
  it("throws an exception when an invalid parameter is given", () => {
```

test-js/test-utils.js
```diff
@@ -0,0 +1,247 @@
...
+  it("should convert a unicode symbol number to a character", () => {
+    expect(getCharFromUnicode(97)).equals("a");
+  });
+});
+
+describe("guessCharFromGlyphName", () => {
+  it("should guess a character from code points in free text", () => {
+    expect(guessCharFromGlyphName("text 0061 text")).equals("a");
+    expect(guessCharFromGlyphName("text ff0061 text")).equals("a");
+    expect(guessCharFromGlyphName("text ff0041 text")).equals("A");
```
justvanrossum, line 228: Please add tests with strings containing
- `100000`
- `10FFFF`
- `110000`

test-js/test-utils.js
```diff
@@ -0,0 +1,247 @@
...
+  });
+});
+
+describe("fileNameExtension", () => {
```
justvanrossum, line 232: Please add a case for a filename that does not contain any dot.

test-js/test-utils.js
```diff
@@ -0,0 +1,247 @@
...
+  it("should work well when there is a dot in the file name", () => {
+    expect(fileNameExtension("test.utils.js")).equals("js");
+  });
+});
+
+describe("arrayExtend", () => {
+  it("should extend arrays with another one", () => {
+    const array = [1, 2, 3];
+    arrayExtend(array, [4, 5]);
+    expect(array).to.eql([1, 2, 3, 4, 5]);
```
justvanrossum, line 245:
```suggestion
    expect(array).to.deep.equal([1, 2, 3, 4, 5]);
```

justvanrossum: Not sure why/how `.eql` works, to be honest.

justvanrossum (changes requested):
Mostly looks good, thank you! I added some suggestions/change requests.

test-js/test-utils.js
justvanrossum: It seems `.eql()` is almost equivalent to `.deep.equal()`, but perhaps not quite. I don't understand the subtlety yet, if there is one.

test-js/test-utils.js
fatih-erikli: But at this case, if the range function fails (we have different test cases for it), this will fail too. So the chain test will be dependent to range test case. What do you think?

test-js/test-utils.js
fatih-erikli: 110000 throws Invalid code point. I tested whether it throws an exception or not, but it feels like this needs to be handled in guessCharFromGlyphName. It should return an empty string instead.

test-js/test-utils.js
```diff
@@ -0,0 +1,213 @@
...
+    expect(() => makeUPlusStringFromCodePoint("")).to.not.throw();
+  });
+  it("make a number unicode hex", () => {
+    expect(makeUPlusStringFromCodePoint(97)).equals("U+0061"); // a
+    expect(makeUPlusStringFromCodePoint(65)).equals("U+0041"); // A
+  });
+});
+
+describe("getCharFromUnicode", () => {
+  it("should return an empty string if an argument is not passed", () => {
```
justvanrossum, line 168:
```suggestion
  it("should return an empty string if the argument is undefined", () => {
```

test-js/test-utils.js
justvanrossum: Ah yes, I agree this is a bug in `guessCharFromGlyphName`. So at last a new unit test is uncovering a bug :). Feel free to propose a fix.

justvanrossum (changes requested):
Thank you for the changes. 

I would love you propose a fix for the `guessCharFromGlyphName()` bug.

fatih-erikli:
Will guessCharFromGlyphName return non-printable characters too? If so, what should the maximum unicode code point be? @justvanrossum.

According to this source: https://jrgraphix.net/research/unicode_blocks.php
E007F (917631) is the last printable one. In that case 0x100000, 0x10FFFF will be invalid too.

justvanrossum:
> In that case 0x100000, 0x10FFFF will be invalid too

Good point. Maybe we should simply stop at 0xFFFFF, at least that's easy to match with the regular expression.

test-js/test-utils.js
justvanrossum: Btw a test for a valid "high" unicode would be good, for example "👀", `U+1F440`.

fatih-erikli:
Do you think it should be handled in regex? Regex string will be longer. It isn't easy to split it in JS when it grows.

I thought something like this:
```
if (codePoint > 0xFFFFF) {
    return "";
  }
```

fatih-erikli:
Deleting the "1" worked.

src/fontra/client/core/utils.js
```diff
@@ -267,7 +267,7 @@ export function guessCharFromGlyphName(glyphName) {
   // Search for a 4-6 char hex string in the glyph name. If 6, the first digit must
```
justvanrossum, line 267: The comment needs updating, too.

src/fontra/client/core/utils.js
```diff
@@ -264,8 +264,8 @@ export function getCharFromUnicode(codePoint) {
 }
 
 export function guessCharFromGlyphName(glyphName) {
-  // Search for a 4-6 char hex string in the glyph name. If 6, the first digit must
-  // be one. Interpret the hex string as a unicode code point and convert to a
+  // Search for a 4-6 char hex string in the glyph name.
```
justvanrossum, line 267:
```suggestion
  // Search for a 4-5 char hex string in the glyph name.
```

test-js/test-utils.js
justvanrossum: Yes. Hope there's no UTF-16 bug hidden in there...

fatih-erikli: Ok. I will test with naked emoji.

test-js/test-utils.js
```diff
@@ -0,0 +1,206 @@
...
+  it("should work well when there is a dot in the file name", () => {
+    expect(fileNameExtension("test.utils.js")).equals("js");
+  });
+  it("should return the given file name when there is no extension", () => {
+    expect(fileNameExtension("utils")).equals("utils");
+  });
+});
+
+describe("arrayExtend", () => {
+  it("should extend arrays with another one", () => {
```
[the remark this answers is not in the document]
fatih-erikli: I saw it but I didn't understand why. Why do we need to add them chunk by chunk?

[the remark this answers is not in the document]
fatih-erikli: It feels like I am testing node.js, that's why I didn't add a test case.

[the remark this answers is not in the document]
fatih-erikli: I added a test case.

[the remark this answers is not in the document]
justvanrossum: > I saw it but I didn't understand why. Why do we need to add them chunk by chunk?

JS arrays don't have `list.extend(iterable)`, except `array.push(...manyItems)`. The latter explodes the call stack when there are too many items. We could push them one by one, but that is possibly slower than to do it in larger (but not unbounded) chunks. See https://github.com/googlefonts/fontra/pull/508.

But I realize the `arrayExtend()` function deserves a comment to explain its existence.

test-js/test-utils.js
```diff
@@ -0,0 +1,211 @@
...
+    const array = [1, 2, 3];
+    arrayExtend(array, [4, 5]);
+    expect(array).deep.equals([1, 2, 3, 4, 5]);
+  });
+  it("test chunk-by-chunk addition by 1024", () => {
+    const destinationArray = [1, 2, 3];
+    arrayExtend(destinationArray, [...range(1024)]);
```
justvanrossum, line 208: We need to test _more than the chunk size_ to trigger the multiple chunk code path. That's why I suggested to do 3000. 1025 would be enough, actually.

test-js/test-utils.js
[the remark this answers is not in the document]
justvanrossum: https://github.com/googlefonts/fontra/pull/544

test-js/test-utils.js
fatih-erikli: Ups, missed that, sent one more commit.

fatih-erikli: Btw tests passes when I replace the arrayExtend content with `thisArray.push(...itemsArray);`. I'm curious, why it's needed to add them separately on each 1024?

justvanrossum (approved):
Thanks! I'll merge.

PR #550, opened 2023-06-14
Show dialog with unsupported file list message.

### LINKED ISSUE #540: [reference viewer] Show dialog if no accepted font file was dropped
https://github.com/googlefonts/fontra/blob/b21019b91eb6868c645da6034e719ba00dd95d17/src/fontra/client/web-components/reference-font.js#L74-L76

fabiocaccamo:
This PR fix #540 by showing a dialog listing the dropped files that are not supported.

### DIFF
--- a/src/fontra/client/web-components/reference-font.js
+++ b/src/fontra/client/web-components/reference-font.js
@@ -60,7 +60,7 @@ export class ReferenceFont extends UnlitElement {
     filesUIList.columnDescriptions = columnDescriptions;
     filesUIList.minHeight = "6em";
     filesUIList.onFilesDrop = (files) => {
-      let fileItemsInvalid = [];
+      const fileItemsInvalid = [];
       const fileItems = [...files]
         .filter((file) => {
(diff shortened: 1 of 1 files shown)


src/fontra/client/web-components/reference-font.js
```diff
@@ -71,9 +77,24 @@ export class ReferenceFont extends UnlitElement {
             fontName: `ReferenceFont${++this.fontCounter}`,
           };
         });
-      // if (!fileItems.length) {
-      //   // FIXME: add dialog() "Unsupported font file type"
-      // }
+      if (fileItemsInvalid.length) {
+        const dialogTitle = "Unsupported font file type";
```
justvanrossum, line 81:
```suggestion
        const dialogTitle = `The following item${
          fileItemsInvalid.length > 1 ? "s" : ""
        } can't be used as a reference font:`;
```

justvanrossum (changes requested):
Thanks! Looks good. Minor wording suggestion.

justvanrossum (approved):

PR #583, opened 2023-06-23
Make sidebars resizable

### LINKED ISSUE #574: [sidebars] Make sidebars resizable
Currently, the left and right sidebars have fixed widths, set with a CSS variable. The user should be able to change this by dragging the edge of the sidebar.

fatih-erikli:
Fixes #574.

I implemented the resizing in `editor.js`, all the other sidebar related functionality is implemented there, however I feel like the sidebar needs to be converted into a web component.

### DIFF
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -326,6 +326,6 @@ export function rgbaToCSS(rgba) {
   return `rgb(${channels.join(",")})`;
 }
 
-export function clampedNumber(number, min, max) {
+export function clamp(number, min, max) {
   return Math.max(Math.min(number, max), min);
 }
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -325,3 +325,7 @@ export function rgbaToCSS(rgba) {
   }
   return `rgb(${channels.join(",")})`;
 }
+
+export function clampedNumber(number, min, max) {
+  return Math.max(Math.min(number, max), min);
+}
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -58,6 +58,7 @@ import {
 } from "../core/path-functions.js";
 import { staticGlyphToGLIF } from "../core/glyph-glif.js";
 import { pathToSVG } from "../core/glyph-svg.js";
+import { clampedNumber } from "../../core/utils.js";
 
 export class EditorController {
   static async fromWebSocket() {
@@ -477,6 +478,7 @@ export class EditorController {
         } else if (growDirection === "right") {
           width = initialWidth + (event.clientX - initialPointerCoordinateX);
         }
+        width = clampedNumber(width, 200, 500);
         sidebarResizing.style.width = `${width}px`;
       }
     });
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -223,7 +223,6 @@ export class EditorController {
     });
 
     this.updateWithDelay();
-    this.initSidebarGutters();
   }
 
   async updateWithDelay() {
@@ -452,6 +451,7 @@ export class EditorController {
     }
 
     this.initSidebarTextEntry();
+    this.initSidebarGutters();
   }
 
   initSidebarGutters() {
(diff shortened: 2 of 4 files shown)


justvanrossum:
Thank you, this is a great start!

I think the cursor should be col-resize in all cases. the east/west variations suggest it will only move in those directions.

There should be a minimal width. Potentially, going beyond the minimal width could collapse the sidebar, as if clicking the active tab.

There should probably also be a maximum width, but I'm not sure what. Perhaps 1/3rd of the page?

Then we need to look at the various content items, especially in the left sidebar: most don't properly get wider with the sidebar.

src/fontra/client/core/utils.js
```diff
@@ -325,3 +325,7 @@ export function rgbaToCSS(rgba) {
   }
   return `rgb(${channels.join(",")})`;
 }
+
+export function clampedNumber(number, min, max) {
```
justvanrossum, line 329: I think we can call this function `clamp`.

Btw. there is a functionally equavalent function called `minmax` defined elsewhere, and there's some inline use of `Math.min(Math.max(...), ...)` that may be replaced with `clamp()`. I would love you to do that, but after this PR.

fatih-erikli: I see in `cjk-design-frame.js` and 'ui-list.js'.

justvanrossum:
Thanks, it works very well now. Can you fix the text entry field so it follows the width, too? 
image

fatih-erikli:
I should have missed that.

fatih-erikli:
Do you want me to create an issue for deleting minmax

src/fontra/views/editor/editor.js
```diff
@@ -452,6 +454,47 @@ export class EditorController {
     this.initSidebarTextEntry();
   }
 
+  initSidebarGutters() {
+    let initialWidth;
+    let initialPointerCoordinateX;
+    let sidebarResizing;
+    let growDirection;
+    document.body.addEventListener("pointerdown", (event) => {
```
justvanrossum, line 462: I know there are two gutter elements, but I wonder: wouldn't it be better to add the event listeners to the gutter elements instead of the body? Then at least some of the event dispatching is done by the browser, and not by our own JS code.

fatih-erikli: Hmm maybe I could register the event when the pointer is down, and unregister when the pointer is up. The events will be added only when a resize is initiated.

fatih-erikli: I will submit a patch.

fatih-erikli: I did.

src/fontra/views/editor/editor.js
```diff
@@ -452,6 +454,48 @@ export class EditorController {
     this.initSidebarTextEntry();
   }
 
+  initSidebarGutters() {
+    let initialWidth;
+    let initialPointerCoordinateX;
```
fatih-erikli, line 459: There is a few ways to achieve the resize, I chose the easiest I could. @justvanrossum.

Calculating the new width without setting the initial state of the element and cursor in `onpointerdown` is possible, but it requires a calculation of the clicked point on the gutter relatively to the sidebar in onpointermove. Otherwise it flickers couple of pixels (gutter size flickering). Example how it must be done https://stackoverflow.com/a/52910578/17805504, here how it must not be done https://stackoverflow.com/a/34487057/17805504 see the flickering.

fatih-erikli:
Btw, resizing it with css is possible, however it isn't easy to update the browser's resize handle appearance (to make it height 100% for example, or placing it to the left). https://developer.mozilla.org/en-US/docs/Web/CSS/resize. So it's better to do it with JS for now

src/fontra/views/editor/editor.js
```diff
@@ -222,6 +223,7 @@ export class EditorController {
     });
 
     this.updateWithDelay();
+    this.initSidebarGutters();
```
justvanrossum, line 226: Can you move this call to the end of `initSidebars()`, just before
```
    this.initSidebarTextEntry();
```

justvanrossum (approved):
Thank you! I like what you've done, and agree with your solution for the event listeners. I have one last small request, then I'll merge.

fabiocaccamo:
@justvanrossum @fatih-erikli I think that sidebars size should persist, actually if I resize them and I refresh the page, their size is reset.

justvanrossum:
> @justvanrossum @fatih-erikli I think that sidebars size should persist, actually if I resize them and refresh the page, their size is reset.

Yes: #589

fabiocaccamo:
I see now that there is already #589

PR #601, opened 2023-07-03
Various fixes related to sidebar resizing.

### LINKED ISSUE #590: [sidebars] Sidebar misc items, related to resizing
I have some followup feedback on #583:

- [ ] Toggling the sidebar doesn't animate anymore (the "animating" class)
- [ ] The "gutter" element to drag the sidebar width takes up space _in addition_ to the margin. The margin + gutter space should be the same as the "old" margin
- [ ] Sliders in sliders tab don't grow with sidebar

fatih-erikli:
This fixes #590.

- [x] Toggling the sidebar doesn't animate anymore (the "animating" class)
- [x] The "gutter" should not take additional space in sidebar
- [x] Sliders in settings tab should grow with sidebar
- [x] User select should be disabled on the root document when resizing

### DIFF
--- a/src/fontra/client/web-components/designspace-location.js
+++ b/src/fontra/client/web-components/designspace-location.js
@@ -7,7 +7,7 @@ export class DesignspaceLocation extends UnlitElement {
   static styles = `
     :host {
       display: grid;
-      grid-template-columns: 25% auto;
+      grid-template-columns: max-content auto;
       gap: 0.4em;
       overflow: scroll;
     }
--- a/src/fontra/client/web-components/designspace-location.js
+++ b/src/fontra/client/web-components/designspace-location.js
@@ -7,7 +7,7 @@ export class DesignspaceLocation extends UnlitElement {
   static styles = `
     :host {
       display: grid;
-      grid-template-columns: max-content auto;
+      grid-template-columns: 25% auto;
       gap: 0.4em;
       overflow: scroll;
     }
--- a/src/fontra/views/editor/editor.html
+++ b/src/fontra/views/editor/editor.html
@@ -32,6 +32,13 @@
       --sidebar-tab-width: 3em;
       --sidebar-content-width-left: 20em;
       --sidebar-content-width-right: 22em;
+      --resize-cursor: col-resize;
+    }
+
+    :root.--resize {
+      user-select: none;
+      -webkit-user-select: none;
+      cursor: var(--resize-cursor);
     }
 
     .editor-container {
@@ -310,7 +317,7 @@
     .sidebar-resize-gutter {
       height: 100%;
       width: 8px;
-      cursor: col-resize;
+      cursor: var(--resize-cursor);
       position: absolute;
     }
 
--- a/src/fontra/views/editor/editor.html
+++ b/src/fontra/views/editor/editor.html
@@ -316,17 +316,17 @@
 
     .sidebar-resize-gutter {
       height: 100%;
-      width: 6px;
+      width: 4px;
       cursor: var(--resize-cursor);
       position: absolute;
     }
 
     .sidebar-container.left .sidebar-resize-gutter {
-      right: -4px;
+      right: -2px;
     }
 
     .sidebar-container.right .sidebar-resize-gutter {
-      left: -4px;
+      left: -2px;
     }
 
     .sidebar-tab > inline-svg {
--- a/src/fontra/views/editor/editor.html
+++ b/src/fontra/views/editor/editor.html
@@ -47,8 +47,7 @@
       background-color: var(--ui-element-background-color);
       height: 100%;
       width: 0;
-      display: flex;
-      justify-content: space-between;
+      position: relative;
     }
 
     .sidebar-container.animating {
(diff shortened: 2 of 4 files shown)


src/fontra/client/web-components/designspace-location.js
```diff
@@ -7,7 +7,7 @@ export class DesignspaceLocation extends UnlitElement {
   static styles = `
     :host {
       display: grid;
-      grid-template-columns: 25% auto;
+      grid-template-columns: max-content auto;
```
justvanrossum, line 10: Could you try fixing the slider resizing while keeping the first column at 25%? The reason it was 25% is to prevent very long axis names to squish away the sliders too much.

src/fontra/views/editor/editor.html
```diff
@@ -32,6 +32,13 @@
       --sidebar-tab-width: 3em;
       --sidebar-content-width-left: 20em;
       --sidebar-content-width-right: 22em;
+      --resize-cursor: col-resize;
+    }
+
+    :root.--resize {
```
justvanrossum, line 38: What's the reason for this particular naming? Couldn't this class simply be called "resizing"?

src/fontra/views/editor/editor.html
```diff
@@ -315,7 +317,12 @@
     .sidebar-resize-gutter {
       height: 100%;
       width: 8px;
-      cursor: col-resize;
+      cursor: var(--resize-cursor);
+      position: absolute;
+    }
+
+    .sidebar-container.left .sidebar-resize-gutter {
+      right: 0;
```
justvanrossum, line 325: Suggest to make this `-4px`, and add one for `.right.` as well: `left: -4px`.

This would center the gutter around the edge, instead of being completely inside the tab. I feel this is more natural beahvior.

I actually wonder if 8px width for the gutter isn't too generous. Perhaps 6 is enough?

justvanrossum: Or even 4px. It seems enough to grab it.

src/fontra/views/editor/editor.html
```diff
@@ -316,13 +316,17 @@
...
       height: 100%;
-      width: 8px;
+      width: 6px;
       cursor: var(--resize-cursor);
       position: absolute;
     }
 
     .sidebar-container.left .sidebar-resize-gutter {
-      right: 0;
+      right: -4px;
```
justvanrossum, line 325: This should be half of the width. Same for right.

Did you try with a width of 4px? Did you find it too narrow?

fatih-erikli: Ok. I made it 4px with, -2px for offsets. Looks good to me.

src/fontra/views/editor/editor.js
```diff
@@ -458,23 +458,25 @@ export class EditorController {
...
           width = initialWidth + (initialPointerCoordinateX - event.clientX);
-        } else if (growDirection === "right") {
+          cssProperty = "--sidebar-content-width-right";
+        } else {
           width = initialWidth + (event.clientX - initialPointerCoordinateX);
+          cssProperty = "--sidebar-content-width-left";
         }
         width = clamp(width, 200, 500);
-        sidebarResizing.style.width = `${width}px`;
+        document.querySelector(":root").style.setProperty(cssProperty, `${width}px`);
```
justvanrossum, line 470: Can we use `document.documentElement` instead of `document.querySelector(":root")`, or would that be a different element, or a less preferred style?

src/fontra/views/editor/editor.html
```diff
@@ -32,6 +32,13 @@
       --sidebar-tab-width: 3em;
       --sidebar-content-width-left: 20em;
       --sidebar-content-width-right: 22em;
+      --resize-cursor: col-resize;
+    }
+
+    :root.resizing {
```
justvanrossum, line 38: Looking at it now I think this class should be called "sidebar-resizing", to make it more obvious what's goin on here. I also wonder if we can move this entire `:root` block closer to the sidebar CSS, for the same reason.

src/fontra/views/editor/editor.js
fatih-erikli: We may move them to `#editor-container`. Is that what you mean?

justvanrossum: No, I literally meant to ask about `document.querySelector(":root")` vs. `document.documentElement`. Are they the same, and if yes, should we use the former instead?

fatih-erikli: Ok. I think it's ok, we can use documentElement. I will send a patch.

src/fontra/views/editor/editor.html
fatih-erikli: I think class "sidebar-resizing" is ok, more explicit is the better. I will send a patch.

What do you mean by moving entire `:root` to the sidebar css?

fatih-erikli: Lowering by one block? There is only sidebar container between them.

justvanrossum: Ah yes, lowering by one block is good, thanks.

fatih-erikli: Sent a patch.

justvanrossum (approved):

PR #608, opened 2023-07-05
Sidebar widths should be stored in localstorage

### LINKED ISSUE #589: [sidebars] Sidebar widths should be stored in localStorage
Sidebar widths should be stored in localStorage, so they persist when reloading. This should probalby be done similar to how the sidebar on/off/which-tab state is stored.

fatih-erikli:
Fixes #589.


justvanrossum (approved):

justvanrossum:
Thank you!

PR #611, opened 2023-07-06
Find a free localhost port when launched locally

### LINKED ISSUE #353: Fontra 'core' should automatically find a free localhost port when launched in 'filesystem' mode

fatih-erikli:
Fixes #353.

### DIFF
--- a/src/fontra/__main__.py
+++ b/src/fontra/__main__.py
@@ -4,7 +4,9 @@ import secrets
 from importlib.metadata import entry_points
 
 from . import __version__ as fontraVersion
-from .core.server import FontraServer
+from .core.server import FontraServer, findFreeTCPPort
+
+DEFAULT_PORT = 8000
 
 
 def main():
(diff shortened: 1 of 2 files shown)


src/fontra/core/server.py
```diff
@@ -77,6 +78,9 @@ def setup(self):
         self.httpApp.on_shutdown.append(self.closeProjectManager)
         self._activeWebsockets = set()
 
+        if self.httpPort == 8000:
```
justvanrossum, line 81: Suggest to do this in `__main__.py`, while keeping the helper function in `server.py`.

src/fontra/__main__.py
```diff
@@ -45,7 +49,7 @@ def main():
     manager = args.getProjectManager(args)
     server = FontraServer(
         host=host,
-        httpPort=httpPort,
+        httpPort=httpPort or findFreeTCPPort(DEFAULT_PORT),
```
justvanrossum, line 52:
```suggestion
        httpPort=httpPort if httpPort is not None else findFreeTCPPort(DEFAULT_PORT),
```

src/fontra/__main__.py
```diff
@@ -15,7 +17,7 @@ def main():
     )
     parser = argparse.ArgumentParser()
     parser.add_argument("--host", default="localhost")
-    parser.add_argument("--http-port", default=8000, type=int)
+    parser.add_argument("--http-port", type=int, help=f"Default port is {DEFAULT_PORT}")
```
justvanrossum, line 20: Suggest to write

    f"The default port is {DEFAULT_PORT}"

justvanrossum (approved):
Thanks!

PR #613, opened 2023-07-06
Fix textarea grows more than 100%

### LINKED ISSUE #609: [text entry sidebar] Layout doesn't respond correctly when the text gets too long
Enter a really long text, and the alignment buttons disappear, and the text box just gets cropped, making it impossible to reach the text beynd that point.

It seems the sidebar content element grows taller than 100vh.

https://github.com/googlefonts/fontra/assets/4246121/44f277ce-d547-4741-8a3e-ff61a4abc9dd

fatih-erikli:
Fixes #609.

### DIFF
--- a/src/fontra/views/editor/editor.html
+++ b/src/fontra/views/editor/editor.html
@@ -231,10 +231,10 @@
       box-sizing: border-box;
       height: 100%;
       width: 100%;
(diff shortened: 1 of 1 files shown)


justvanrossum:
Thanks!

PR #626, opened 2023-07-10
Can't edit glyph dialog refinement

### LINKED ISSUE #591: [editor] "Can’t edit glyph" dialog should be refined
If the editor is not "at a source", and the user tries to move a point or component, we get a dialog that says:

    Can’t edit glyph “<name-of-glyph>”
    Location is not at a source.
                      (( Okay ))

What it should do instead is:

    Can’t edit glyph “<name-of-glyph>”
    Would you like to create a new source or navigate to the nearest source?
      ( Cancel )  ( Create new source )  (( Go to nearest source ))

fatih-erikli:
Work in progress, it will fix #591. 

- [x] Dialog button labels

Screen Shot 2023-07-10 at 06 40 27

- [x] Go to nearest skip inactive sources

Screen Shot 2023-07-10 at 06 43 21

### DIFF
--- a/src/fontra/client/core/glyph-controller.js
+++ b/src/fontra/client/core/glyph-controller.js
@@ -281,7 +281,7 @@ export class VariableGlyphController {
     });
   }
 
-  findNearestSourceFromGlobalLocation(location) {
+  findNearestSourceFromGlobalLocation(location, skipInactive) {
     const normalizedLocation = normalizeLocation(
       this.mapLocationGlobalToLocal(location),
       this.combinedAxes
@@ -289,7 +289,8 @@ export class VariableGlyphController {
     const indexInfo = findNearestSourceIndexFromLocation(
       this.glyph,
       normalizedLocation,
-      this.combinedAxes
+      this.combinedAxes,
+      skipInactive
     );
     return indexInfo.index;
   }
@@ -733,13 +734,17 @@ function subsetLocation(location, axes) {
   return subsettedLocation;
 }
 
-function findNearestSourceIndexFromLocation(glyph, location, axes) {
+function findNearestSourceIndexFromLocation(glyph, location, axes, skipInactive) {
   const distances = [];
   if (!glyph.sources.length) {
     throw Error("assert -- glyph has no sources");
   }
   for (let i = 0; i < glyph.sources.length; i++) {
-    const sourceLocation = normalizeLocation(glyph.sources[i].location, axes);
+    const source = glyph.sources[i];
+    if (skipInactive && source.inactive) {
+      continue;
+    }
+    const sourceLocation = normalizeLocation(source.location, axes);
     let distanceSquared = 0;
     for (const [axisName, value] of Object.entries(location)) {
       const sourceValue = sourceLocation[axisName];
--- a/src/fontra/client/core/glyph-controller.js
+++ b/src/fontra/client/core/glyph-controller.js
@@ -281,7 +281,7 @@ export class VariableGlyphController {
     });
   }
 
-  findNearestSourceFromGlobalLocation(location, skipInactive) {
+  findNearestSourceFromGlobalLocation(location, skipInactive = false) {
     const normalizedLocation = normalizeLocation(
       this.mapLocationGlobalToLocal(location),
       this.combinedAxes
@@ -734,7 +734,12 @@ function subsetLocation(location, axes) {
   return subsettedLocation;
 }
 
-function findNearestSourceIndexFromLocation(glyph, location, axes, skipInactive) {
+function findNearestSourceIndexFromLocation(
+  glyph,
+  location,
+  axes,
+  skipInactive = false
+) {
   const distances = [];
   if (!glyph.sources.length) {
     throw Error("assert -- glyph has no sources");
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -421,17 +421,36 @@ export class EditorController {
       this.updateWindowLocationAndSelectionInfo();
       this.autoViewBox = false;
     });
-    this.sceneController.addEventListener("createNewSource", (event) => {
-      this.sidebarDesignspace.addSource();
-    });
-    this.sceneController.addEventListener("goToNearestSource", async (event) => {
-      const glyphController =
-        await this.sceneController.sceneModel.getSelectedVariableGlyphController();
-      const nearestSourceIndex = glyphController.findNearestSourceFromGlobalLocation(
-        this.sceneController.getLocation(),
-        true
+    this.sceneController.addEventListener("cantEditGlyphNotAtSource", async () => {
+      const glyphName = this.sceneController.sceneModel.getSelectedGlyphName();
+      const result = await dialog(
+        `Can’t edit glyph “${glyphName}”`,
+        "Location is not at a source.",
+        [
+          { title: "Cancel", resultValue: "cancel", isCancelButton: true },
+          { title: "New source", resultValue: "createNewSource" },
+          {
+            title: "Go to nearest source",
+            resultValue: "goToNearestSource",
+            isDefaultButton: true,
+          },
+        ]
       );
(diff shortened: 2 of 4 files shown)


justvanrossum:
> Dialog buttons shows odd when more than two buttons added. I guess there is work in progress for dialogs. Do you want me to fix the css part of it?

The dialog has a maximum width: `max-width: 32em`. We may want to increae that, but we can also use a shorter label for the second button: `"New source"`.

justvanrossum:
>  I am not sure what the "go to nearest" button should do when the "on" checkbox is checked off in sources list. Should we make the checkbox "on" when clicked to go to nearest?

Very good question. The current behavior is unintentional: it seems you can't even view the disabled source. This needs fixing regardless of the new dialog content.

fatih-erikli:
I corrected the button labels as you suggested. They fit in dialog. @justvanrossum

justvanrossum:
> Apart from the checkbox issue, does the event handling of the buttons look okay to you?

Not sure yet. The facatoring in the code you need is not ideal, so perhaps to use events is a good solution.

I notice two other problems, when testing this with MutatorSans.ttf:
- it seems that "go to nearest" always goes to the default source instead of the actual nearest
- the source gets selected, but the sliders don't get updated

We may have to look at this together.

fatih-erikli:
Ok. I guess my way of reproducing the issue is different than yours @justvanrossum.

src/fontra/views/editor/editor.js
```diff
@@ -421,6 +421,17 @@ export class EditorController {
...
+    this.sceneController.addEventListener("createNewSource", (event) => {
+      this.sidebarDesignspace.addSource();
+    });
+    this.sceneController.addEventListener("goToNearestSource", async (event) => {
+      const glyphController =
+        await this.sceneController.sceneModel.getSelectedVariableGlyphController();
+      const nearestSource = glyphController.findNearestSourceFromGlobalLocation(
+        this.sceneController.getLocation()
+      );
+      this.designspaceLocationController.model.location = nearestSource;
```
justvanrossum, line 433: `nearestSource` is a source index, not a location. So you'll need to find out the location for that source. I think this causes both issues I mentioned.

justvanrossum: Or actually, let me have a look, I think we need to just "select te source" by index. Perhaps we need to add a method to the sidebar code.

(I'm in the process of figuring out how to refactor some if this, especially the messy relationship between scene-controller and scene-model.)

src/fontra/views/editor/scene-controller.js
```diff
@@ -473,9 +473,20 @@ export class SceneController {
         const result = await dialog(
```
justvanrossum, line 473: Please edit the comments right above this line :)

src/fontra/views/editor/editor.js
fatih-erikli: Thanks.

justvanrossum:
I think for now dispatching these events is good. I may change it in the near future, when I've worked out my intended refactoring.

fatih-erikli:
I sent a commit, added a skipInactive parameter for finding the nearest. If the checkbox is off, it will select the closest one. Do you think this behavior is correct? @justvanrossum

justvanrossum:
> I sent a commit, added a skipInactive parameter for finding the nearest. If the checkbox is off, it will select the closest one. 

Thanks, that's a useful parameter for sure.

> Do you think this behavior is correct?

I think so. (I'm not 100% convinced, but this is more convincing so far than the alternative.)

src/fontra/client/core/glyph-controller.js
```diff
@@ -733,13 +734,17 @@ function subsetLocation(location, axes) {
   return subsettedLocation;
 }
 
-function findNearestSourceIndexFromLocation(glyph, location, axes) {
+function findNearestSourceIndexFromLocation(glyph, location, axes, skipInactive) {
```
justvanrossum, line 737:
```suggestion
function findNearestSourceIndexFromLocation(glyph, location, axes, skipInactive = false) {
```

justvanrossum: Just to make it more explicit that this is an optional argument.

src/fontra/views/editor/scene-controller.js
```diff
@@ -466,16 +466,23 @@ export class SceneController {
...
+            { title: "Cancel", resultValue: "cancel", isCancelButton: true },
+            { title: "New source", resultValue: "createNewSource" },
+            {
+              title: "Go to nearest source",
+              resultValue: "goToNearestSource",
+              isDefaultButton: true,
+            },
+          ]
         );
+        if (result === "cancel") {
```
justvanrossum, line 482: Suggest to invert the condition:
```js
    if (result !== "cancel") {
      this._dispatchEvent(result);
    }
```

justvanrossum:
Now that I've played with this code, and thought a bit more about how it's structured, and the events you introduced, I think we should move the entire dialog code to editor.js, and dispatch a single event from scene-controller.js: "cantEditGlyphNotAtSource".

So pretty much
```js
      if (!glyphController.canEdit) {
        this._dispatchEvent("cantEditGlyphNotAtSource");
        return;
      }
```
I think that makes a bit more sense than to show a dialog from scene-controller.js.

What do you think?

fatih-erikli:
Ok. It sounds good to me. I will

src/fontra/views/editor/editor.js
```diff
@@ -421,17 +421,36 @@ export class EditorController {
...
-    this.sceneController.addEventListener("createNewSource", (event) => {
-      this.sidebarDesignspace.addSource();
-    });
-    this.sceneController.addEventListener("goToNearestSource", async (event) => {
-      const glyphController =
-        await this.sceneController.sceneModel.getSelectedVariableGlyphController();
-      const nearestSourceIndex = glyphController.findNearestSourceFromGlobalLocation(
-        this.sceneController.getLocation(),
-        true
+    this.sceneController.addEventListener("cantEditGlyphNotAtSource", async () => {
```
justvanrossum, line 424: Please move the addEventHandler call to right after the addEventHandler call for "doubleClickedComponents", and move the handler implementation to a new method called "handleCantEditGlyphNotAtSource()" (or something else if you can come up with a better name).

fatih-erikli: Ok I did. I created "showGlyphHasNoSourceDialog" method and I added "glyphHasNoSource" event.

justvanrossum: Thanks!

I don't like the name though. How about "glyphEditLocationNotAtSource" + "showGlyphEditLocationNotAtSource". It is not the glyph that doesn't have a source, but that there is no source at the desired edit location.

fatih-erikli: Sounds better.

src/fontra/views/editor/editor.js
```diff
@@ -163,6 +163,10 @@ export class EditorController {
       this.doubleClickedComponentsCallback(event);
     });
 
+    this.sceneController.addEventListener("glyphEditLocationNotAtSource", async () => {
+      this.showGlyphEditLocationNotAtSource();
```
justvanrossum, line 167:
```suggestion
      this.showDialogGlyphEditLocationNotAtSource();
```
(Same below)

src/fontra/views/editor/editor.js
```diff
@@ -358,6 +362,37 @@ export class EditorController {
     userSettings.items = items;
   }
 
+  async showGlyphEditLocationNotAtSource() {
+    const glyphName = this.sceneController.sceneModel.getSelectedGlyphName();
+    const result = await dialog(
+      `Can’t edit glyph “${glyphName}”`,
+      "Location is not at a source.",
```
justvanrossum, line 369:
```suggestion
      "The location is not at a source.",
```

justvanrossum (approved):
Thanks!

PR #639, opened 2023-07-11
Use fontTools.misc.transform.DecomposedTransform

### LINKED ISSUE #245: Use fontTools.misc.transform.DecomposedTransform
Use fontTools.misc.transform.DecomposedTransform and remove `fontra.core.classes.Transformation`

This requires https://github.com/fonttools/fonttools/pull/2958 to be landed in a fonttools release.

TODO:
- [ ] bump fonttools dependencies in requirements.txt and pyproject.toml
- [ ] Use the conversion code from fonttools to replace makeAffineTransform and decomposeTwoByTwo.
- [ ] Review the skewX issue (fonttools flips the sign compared to the glyf-1 spec)

fatih-erikli:
Fixes #245.

- [x] Use decomposed transform for creating affine transformation
- [x] Bump fonttols in pyproject.toml
- [x] Delete decomposeTwoByTwo
- [x] Delete fontra.core.classes.Transformation

Work in progress


fatih-erikli:
I did only unit test checking. It will require a real test

src/fontra/backends/designspace.py
```diff
@@ -682,14 +680,17 @@ def buildUFOLayerGlyph(
         if component.location:
             # It's a variable component
             varCoDict = {"base": component.name, "location": component.location}
-            if component.transformation != Transformation():
+            if isinstance(component.transformation, DecomposedTransform):
```
fatih-erikli, line 683: I would like your attention in this line @justvanrossum. I am not sure if I understand it correctly

src/fontra/backends/designspace.py
```diff
@@ -682,14 +680,17 @@ def buildUFOLayerGlyph(
...
             variableComponents.append(varCoDict)
         else:
             # It's a regular component
+            decomposedTransform = DecomposedTransform(
+                **asdict(component.transformation)
+            )
             pen.addComponent(
                 component.name,
-                cleanAffine(makeAffineTransform(component.transformation)),
+                cleanAffine(decomposedTransform.toTransform()),
```
fatih-erikli, line 693: Should I rename cleanAffine? @justvanrossum

src/fontra/backends/designspace.py
justvanrossum: No, the original check was correct, and should be replaced with 
```python
    if component.transformation != DecomposedTransform():
```
It compares `component.transformation` with the *identity transform* (ie. "no transformation")

src/fontra/backends/designspace.py
```diff
@@ -682,14 +680,17 @@ def buildUFOLayerGlyph(
         if component.location:
             # It's a variable component
             varCoDict = {"base": component.name, "location": component.location}
-            if component.transformation != Transformation():
+            if isinstance(component.transformation, DecomposedTransform):
                 varCoDict["transformation"] = asdict(component.transformation)
             variableComponents.append(varCoDict)
         else:
             # It's a regular component
+            decomposedTransform = DecomposedTransform(
```
justvanrossum, line 688: I don't think this should be necessary: `component.transform` should already be a `DecomposedTransform` instance.

src/fontra/backends/designspace.py
justvanrossum: Why?

src/fontra/core/packedpath.py
```diff
@@ -240,74 +241,20 @@ def endPath(self):
...
 
     def addVarComponent(
         self, glyphName, transformation, location, identifier=None, **kwargs
     ):
-        from .classes import Component, Transformation
+        from .classes import Component
 
-        # TODO: https://github.com/googlefonts/fontra/issues/245
-        transformation = Transformation(**asdict(transformation))
+        transformation = DecomposedTransform(**asdict(transformation))
```
justvanrossum, line 254: Ah, here we were receiving a `DecomposedTransform` and we needed it to be a `Transformation`. So strictly this isn't necessary anymore. It may still be good to make a copy of the object, though, but let's use `copy()` (from the `copy` module) for that.

src/fontra/backends/designspace.py
fatih-erikli: Should I create an DecomposedTransformIdentity? There's misc.transform.Transform.Identity for regular ones.

justvanrossum: I might be a small optimization, but let's not do that. It's also conceptually not quite right: `Transform` instances are immutable, but `DecomposedTransform` instances are not, so there's no guarantee that `DecomposedIdentity` remains the identity (even though no code would be supposed to change it).

src/fontra/backends/designspace.py
fatih-erikli: MakeAffineTransform is deleted. The term `affine` used only in that function. I thought something like that `normalizeFloats` but it is not important.

src/fontra/backends/designspace.py
fatih-erikli: Ok, fixed.

src/fontra/core/packedpath.py
fatih-erikli: I sent a patch but I noticed that unit tests do not cover this function currently. I am not sure if it works correctly.

src/fontra/backends/designspace.py
justvanrossum: Ok yeah, the name is indeed not great. It also converts it to a tuple. Maybe just `cleanupTransform()`?

src/fontra/core/packedpath.py
justvanrossum: Ah yes, I see. Let me think. I can test it manually for now, but indeed we'd need a test case. It would need a special constructed font, though, which we currently don't have in the repo.

src/fontra/backends/designspace.py
justvanrossum: I've renamed it to `cleanupTransform()`.

PR #640, opened 2023-07-12
Rectangle.js unit tests

### LINKED ISSUE #636: rectangle.js needs unit tests
It should use the `parametrize()` helper func as introduced in #638.

fatih-erikli:
Fixes #636

### DIFF
--- a/test-js/test-rectangle.js
+++ b/test-js/test-rectangle.js
@@ -41,6 +41,9 @@ describe("pointInRect", () => {
       true,
       "Should work with floats",
     ],
+    [{ x: 220, y: 220 }, { xMin: 0, yMin: 0, xMax: 200, yMax: 200 }, false],
+    [{ x: 220, y: -1 }, { xMin: 0, yMin: 0, xMax: 200, yMax: 200 }, false],
+    [{ x: -1, y: 200 }, { xMin: 0, yMin: 0, xMax: 200, yMax: 200 }, false],
   ];
   parametrize(
     "is point in rectangle",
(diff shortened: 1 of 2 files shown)


test-js/test-rectangle.js
```diff
@@ -0,0 +1,404 @@
...
+    [
+      { xMin: 50, yMin: 50, xMax: 100, yMax: 100 },
+      { xMin: 75, yMin: 75, xMax: 80, yMax: 80 },
+      { xMin: 75, yMin: 75, xMax: 80, yMax: 80 },
+    ],
+    [
+      { xMin: 50, yMin: 50, xMax: 60, yMax: 60 },
+      { xMin: 60, yMin: 60, xMax: 70, yMax: 70 },
+      { xMin: 60, yMin: 60, xMax: 60, yMax: 60 },
+      "Creates rectangle with sizes of 0 if they intersect at the edge", // should be clarified with @justvanrossum
```
fatih-erikli, line 118: @justvanrossum Should this be fixed in code?

justvanrossum (commented):
Thanks, very good! I left some minor suggestions.

test-js/test-rectangle.js
justvanrossum: It's this line in the code, it should use `>=` instead of `>`:
https://github.com/googlefonts/fontra/blob/952497b65479f6adec8b1d24a93ae59d8bc5b405/src/fontra/client/core/rectangle.js#L47

It comes from fonttools, which indeed uses `>=`. I probably made a mistake when porting this code.
https://github.com/fonttools/fonttools/blob/fb56e7b7c9715895b81708904c840875008adb9c/Lib/fontTools/misc/arrayTools.py#L209

So yes, please fix.

Issue #646, opened 2023-07-13
Fontra scripts won't load


madig:
When launching fontra from a git checkout, I get these errors in Firefox:

```
Loading module from “http://localhost:8000/third-party/lit.04d006ff.js” was blocked because of a disallowed MIME type (“text/plain”).
Loading failed for the module with source “http://localhost:8000/third-party/lit.04d006ff.js”.
Loading module from “http://localhost:8000/third-party/bezier-js.04d006ff.js” was blocked because of a disallowed MIME type (“text/plain”).
Loading failed for the module with source “http://localhost:8000/third-party/bezier-js.04d006ff.js”.
```

Chrome says:

```
Failed to load resource: the server responded with a status of 404 (Not Found) (lit.04d006ff.js:1)
Failed to load resource: the server responded with a status of 404 (Not Found) (bezier-js.04d006ff.js:1)
```

When launching Fontra Pak, I get even more errors and no UI. Not sure what this means... uBlock is disabled in both browsers.

justvanrossum:
Which OS? Maybe there's something wrong with how mime types are determined.

It works for me on macOS 10.15 + recent Firefox.

justvanrossum:
Can you tell me the output of the following command?
```
python -c "import mimetypes; print(mimetypes.guess_type('test.js'))"
```

justvanrossum:
We've had a similar mime type problem before (#575) so now I'm worried that `mimetypes .guess_type()` is even more useless than I thought. It's a fairly trivial mapping for Fontra, though, so it'll be easy to replace.

justvanrossum:
Would you look at that: https://github.com/tensorflow/tensorboard/pull/3128/files I bet this is our issue as well.

justvanrossum:
I've merged a fix that I am fairly confident about: I could reproduce the error messages in Firefox when I deliberately used the wrong mime types. But not your Chrome messages, so that makes me doubt.

Please reopen if the problem persists.

madig:
Can't reopen, but problem persists with same error :(

justvanrossum:
Are you sure you upgraded / installed correctly?

madig:
Uh, I did a git pull and then `hatch run fontra --launch filesystem <...>`. It happens on my Windows work PC (Windows and WSL) and my personal Linux PC :shrug:

justvanrossum:
[...]

PR #651, opened 2023-07-16
[build hook] Add artifact paths to the build data

justvanrossum:
Add artifact paths to the build data, perhaps fixing #646

This is probably how it should have been done in the first place.


justvanrossum:
This is either unnecessary, or not quite the right solution. Closing.

PR #671, opened 2023-07-26
Draft: Fit cubic port to JS

fatih-erikli:


fatih-erikli:
Shall we add `zip` function to the utils? @justvanrossum It's used in some places in python version of the implementation.

src/fontra/client/core/vector.js
```diff
@@ -7,7 +7,11 @@ export function subVectors(vectorA, vectorB) {
 }
 
 export function mulVector(vector, scalar) {
```
justvanrossum, line 9: Instead of modifying this function, let's add a new one vor vector * vector called `mulVectorVector()`.

justvanrossum: Additionally, we could consider renaming the original function to `mulVectorScalar()`.

src/fontra/client/core/vector.js
```diff
@@ -61,3 +65,7 @@ export function intersect(pt1, pt2, pt3, pt4) {
 export function distance(pt1, pt2) {
   return vectorLength(subVectors(pt2, pt1));
 }
+
+export function dotVector(point, point2) {
```
justvanrossum, line 69: Let's rename this `dotProduct()`, and the arguments `vectorA` and `vectorB` to keep the style consistend with the rest of the module.

justvanrossum (commented):
First round of review.

src/fontra/client/core/vector.js
fatih-erikli: Ok. That makes sense

src/fontra/client/core/fit-cubic.js
```diff
@@ -0,0 +1,151 @@
+import { enumerate, range } from "./utils.js";
+import {
+  subVectors,
```
justvanrossum, line 3: Please list imported name in sorted order.

Issue #451, opened 2023-05-05
Deleting points is quite buggy


justvanrossum:
On the one hand there is poor UI behavior, on the other hand it's easily possible to offend the designspace backend:

<img width="584" alt="image" src="https://user-images.githubusercontent.com/4246121/236429666-3735793e-059c-45fb-a6e1-51d38b55c32e.png">

Also relates to #438

The main problem is that deleting points can cause contours to violate the "cubuc segments must have exactly two off-curve points" rule.

So the main task here is to fix `deleteSelectedPoints()` so it doesn't cause invalid cubic contours.

PR #679, opened 2023-07-29
Include smooth curve handles when deleting a point

fatih-erikli:
Draft PR for some part of #451


justvanrossum:
I need to more homework in preparation of this task. We'll have to take a different approach.

PR #685, opened 2023-08-02
Make reference font settings persistent

### LINKED ISSUE #519: [reference viewer] Store uploaded fonts in the "origin private file system"
Currently, reloading the page loses the reference font(s). This could be solved by storing the font data in the "origin private file system".

https://developer.mozilla.org/en-US/docs/Web/API/File_System_Access_API

### LINKED ISSUE #686: [reference font] Visibility setting is not restored after page reload
~This is because the vis layer definition is registered too late.~ _Maybe it isn't_

This code needs some refactoring anyway (there's way too much reference font code in editor.js).

Alternatively, perhaps it can be fixed by using the `synchronizeItemWithLocalStorage()` API.

Also relates to #685.

fabiocaccamo:
Make reference font settings persistent by storing dropped font in the Origin Private File System (OPFS).

This PR fixes #519. Also fixes #686.

### DIFF
--- a/src/fontra/client/web-components/reference-font.js
+++ b/src/fontra/client/web-components/reference-font.js
@@ -344,8 +344,8 @@ function getWriteWorker() {
 }
 
 function writeFontFileToOPFSInWorker(fileName, file) {
-  const worker = getWriteWorker();
   return new Promise((resolve, reject) => {
+    const worker = getWriteWorker();
     worker.onmessage = (event) => {
       if (event.data.error) {
         reject(event.data.error);
--- a/src/fontra/client/web-components/reference-font.js
+++ b/src/fontra/client/web-components/reference-font.js
@@ -210,23 +210,14 @@ function readFileAsync(file) {
 }
 
 async function getOPFSInfo() {
-  try {
-    return navigator.storage.estimate().then((info) => {
-      return {
-        free: info.quota - info.usage,
-        usage: info.usage,
-        percent: Math.round((info.usage / info.quota) * 100),
-        quota: info.quota,
-      };
-    });
-  } catch (error) {
-    dialog(
-      "Unable to retrieve Origin Private File System informations.",
-      error.toString(),
-      [{ title: "OK" }],
-      5000
-    );
-  }
+  return navigator.storage.estimate().then((info) => {
+    return {
+      free: info.quota - info.usage,
+      usage: info.usage,
+      percent: Math.round((info.usage / info.quota) * 100),
+      quota: info.quota,
+    };
+  });
 }
 
 async function getOPFSFontsDir() {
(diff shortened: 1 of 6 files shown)


src/fontra/client/web-components/reference-font.js
```diff
@@ -49,7 +49,7 @@ export class ReferenceFont extends UnlitElement {
     return this.controller.model;
   }
 
-  render() {
+  async render() {
```
justvanrossum, line 52: I think the render() method should stay sync (it should run fast). It can still trigger loading of the fonts, but it can add them to the list UI when it's done. So, either you should use a .then() on the loading result, _or_ add another method that is async to load the fonts, but call it _without_ awaiting. I prefer the latter solution, as it also nicely factors out font loading from the opfs.

src/fontra/client/web-components/reference-font.js
```diff
@@ -183,4 +194,142 @@ function readFileAsync(file) {
   });
 }
 
+async function getOPFSInfo() {
```
justvanrossum, line 197: Instead of computing whether there may be enough space to store the font, I wonder if it isn't simpler to just try to store it, and catch the error when storing fails.

src/fontra/client/web-components/reference-font.js
```diff
@@ -183,4 +194,142 @@ function readFileAsync(file) {
...
+    return navigator.storage.estimate().then((info) => {
+      return {
+        free: info.quota - info.usage,
+        usage: info.usage,
+        percent: Math.round((info.usage / info.quota) * 100),
+        quota: info.quota,
+      };
+    });
+  } catch (error) {
+    dialog(
```
justvanrossum, line 208: Is this actually happening? Why the need for a dialog? Sounds like if this fails we have bigger problems to worry about, and I'm not sure the end user is helped in any way by this message.

justvanrossum (commented):
I left some initial comments.

I see this does not yet take into account duplicate font files. If I drop the same font twice, then reload the page, I only see it once in the list. This also happens if I drop a _different_ font, that happens to have the same file name.

It also does not yet synchronize the list across different tabs/windows.

src/fontra/client/web-components/reference-font.js
fabiocaccamo: This happens only if the browser doesn't support `navigator.storage`, it is nicely supported but there are some exceptions like IE and Safari < 15.2: 
https://caniuse.com/?search=navigator.storage

Showing the dialog may be not useful to the user, but it could be for us. 
In case a user reports us a problem including an error message it would be easier for us to understand what is happening, in all other cases it's just a dead dialog.

fabiocaccamo:
I'm wondering if when dropping a new font, it should be automatically selected.

fabiocaccamo:
I discovered a bug by dropping a variable font with square brackets in the filename.

src/fontra/client/web-components/reference-font.js
justvanrossum: We don't support IE at all, and we also do not support Safari < 15.4 (we use `array.at()`, which is Safari >= 15.4)

I don't like debug dialogs in arbitrary places, because there is literally no good rationaly for when to do or not do it. We need to see console output in case of problems anyway. Later we could perhaps do a more proper up-front browser compatibility check, before we load anything important.

justvanrossum:
> I'm wondering if when new font is dropped, it should be automatically selected.

Perhaps, yes.

justvanrossum:
> I discovered a bug by dropping a variable font with square brackets in the filename.

I think I see the problem, but I don't see an error in the console.

fabiocaccamo:
> I think I see the problem happening, but I don't see an error in the console.

I see it:
Screenshot 2023-08-03 at 18 04 20


The problem is the name passed to the `FontFace` constructor. Replacing incompatible chars with hyphens solves the error, but the font is not rendered in the canvas (should understand the reason).

justvanrossum:
Ah yes. We don't have to use the filename there at all, could be a counter or GUID-ish thing.

justvanrossum:
TODO:
- [ ] make it work on Safari, which doesn't support the `createWritable()` method :(

PR #697, opened 2023-08-05
Add async-utils: memoize

fatih-erikli:


test-js/test-async-utils.js
```diff
@@ -0,0 +1,35 @@
...
+  it("should give the awaiting promise when a function called before the previous execution is done", async () => {
+    let nTimesWorked = 0;
+    const func = memoize(async (n) => {
+      nTimesWorked += 1;
+      await new Promise((resolve) => {
+        setTimeout(resolve, 50);
+      });
+      return n * n;
+    });
+    const pending = func(2);
```
justvanrossum, line 26: You could maybe also check something to the effect of
```js
    expect(func(2)).to.equal(func(2))
    expect(func(2)).to.not.equal(func(4))
```

justvanrossum (commented):
The function as written _also_ works for non-async functions, and I think (at least for now) it is better placed in utils.js, and the tests in test-utils.js.

To prove this is true, you can duplicate you first test case and remove all async/await keywords.

src/fontra/client/core/async-utils.js
```diff
@@ -0,0 +1,12 @@
+export function memoize(func) {
+  const cache = new Map();
+  return async (...args) => {
+    const cacheKey = JSON.stringify(args);
+    if (cache.has(cacheKey)) {
+      return cache.get(cacheKey);
+    }
+    const promise = func(...args);
```
justvanrossum, line 8: So you could rename "promise" to "result".

Issue #687, opened 2023-08-03
Refactor sidebars


justvanrossum:
Construct Sidebars in code rather than hardcoding them in the HTML (to be ready for sidebar plugins)

Actions:
- Add method(s) to EditorController to add a sidebar, modeled after `addEditTool(tool)`:
  - `addSidebarLeft(sidebar)`
  - `addSidebarLeft(sidebar)`
  ...or...
  - `addSidebar(sidebar, side)`
- Add new module in "editor": `sidebar-base.js`, providing a base class for sidebars: `BaseSidebar` (or `SidebarBase`? See `BaseTool`)
- Refactor each sidebar so that it has its own module (this is already the case for most sidebars), with a sublass of `BaseSidebar`

Each sidebar class needs to provide
- An icon path (see the `BaseTool` subclasses)
- An identifier (see the `BaseTool` subclasses)

The constructor should be in the baseclass, and should take a single argument: the editor (EditorController instance). Perhaps it can set up some convenience properties, just like `BaseTool` does.

The baseclass should provide a `toggle(onOff)` method that does nothing, but can be overridden by individual sidebar classes, to replace the `toggleXxxx` functionality in editor.js. Perhaps there should even be some functionality in the baseclass so that subclasses have a common way to know whether they are visible or not. (They should have a chance to not do work when they're invisible, and update when they become visibible.)

The `addSidebar(...)` call should take care of manipulating the dom to add:
- The tabbed sticking-out bit witht the icon
- The sidebar content (by calling a method on the sidebar, say `sidebar.getContentElement()`

PR #705, opened 2023-08-08
Refactor sidebars

fatih-erikli:
Draft PR for #687


src/fontra/views/editor/editor.js
```diff
@@ -83,6 +85,16 @@ export class EditorController {
   }
 
   constructor(font) {
+    const editorContainer = document.querySelector(".editor-container");
+    const sidebarLeft = new SidebarLeft(this);
+    const sidebarRight = new SidebarRight(this);
+    // this.addSidebar(sidebarLeft);
```
fatih-erikli, line 91: please ignore this

fatih-erikli:
I suggest to add a `Panel` class, that will be derived by these:

- TextEntryPanel
- GlyphSearchPanel
- DesignspaceNavigationPanel
- UserSettingsPanel
- ReferenceFontPanel
- SelectionInfo

Each class will render its dom content in their `attach` method.

Sidebar will hold these panels.

fatih-erikli:
Forced-pushed because of rebasing to fix merge conflict with main branch

src/fontra/views/editor/editor.js
```diff
@@ -424,6 +425,27 @@ export class EditorController {
...
+    this.addSidebarPanel(new DesignspaceNavigationPanel(), "left");
+    this.addSidebarPanel(new UserSettingsPanel(), "left");
+    this.addSidebarPanel(new ReferenceFontPanel(), "left");
+    this.addSidebarPanel(new SelectionInfoPanel(), "right");
+
+    for (const sidebar of sidebars) {
+      sidebar.attach(editorContainer);
+    }
+
+    for (const sidebar of sidebars) {
```
justvanrossum, line 443: I think this could/should be handled by `sidebar.addPanel(...)`?

A single call to `this.addSidebarPanel(...)` should be enough to add a new sidebar panel.

src/fontra/views/editor/panel-text-entry.js
```diff
@@ -43,11 +98,13 @@ export class SidebarTextEntry {
...
-  focusTextEntry() {
-    this.textEntryElement.focus();
+  fixTextEntryHeight() {
+    // This adapts the text entry height to its content
+    this.textEntryElement.style.height = "auto";
+    this.textEntryElement.style.height = this.textEntryElement.scrollHeight + 14 + "px";
   }
 
-  setupIntersectionObserver() {
+  setupInteractionObserver() {
```
justvanrossum, line 107: Please revert this rename, the original name is as intended.

src/fontra/views/editor/editor.js
```diff
@@ -463,177 +454,53 @@ export class EditorController {
...
+    sidebar.attach(editorContainer);
+    this.sidebars.push(sidebar);
+  }
 
-    this.initSidebarTextEntry();
-    this.initSidebarGutters();
+  addSidebarPanel(panelInstance, sidebarName) {
+    const sidebar = this.sidebars.find((sidebar) => sidebar.identifier === sidebarName);
+    sidebar.addPanel(panelInstance);
+    panelInstance.attach(this);
```
justvanrossum, line 468: I would prefer if the EditorController instance was passed to the constructor.

justvanrossum (commented):
There is still some sidebar-specific CSS in editor.css. Perhaps Panel should be a subclass from SimpleElement, so we'd effectively make each panel a web component with its own isolated CSS.

justvanrossum:
If Panel becomes a subclass of SimpleElement, then we will indeed need "attach", although I may prefer a different name (but I don't have a better idea right now).

src/fontra/views/editor/editor.js
fatih-erikli: this should be ok

src/fontra/views/editor/panel-text-entry.js
fatih-erikli: this too

src/fontra/client/web-components/ui-form.js
```diff
@@ -1,21 +1,74 @@
...
+    .ui-form-value input[type="range"] {
+      width: 7em;
+    }
+
+    .ui-form-value.text {
+      white-space: normal;
     }
-    this.container.classList.add("ui-form");
-    this.setFieldDescriptions(fieldDescriptions);
+  `;
```
justvanrossum, line 62: Blank line between CSS and constructor:
```suggestion
  `;
 
```

src/fontra/client/web-components/ui-form.js
```diff
@@ -1,21 +1,74 @@
...
+  constructor() {
+    super();
+    this._attachStyles();
```
justvanrossum, line 65: This call should not be needed (it should be done as part of `super()`, but please check.

src/fontra/views/editor/panel.js
```diff
@@ -0,0 +1,19 @@
+import { SimpleElement } from "/core/unlit.js";
+
+export default class Panel extends SimpleElement {
+  constructor(editorController) {
+    super();
+    this.editorController = editorController;
+    this._attachStyles();
```
justvanrossum, line 7: Likewise, the call to `_attachStyles()` should not be needed.

src/fontra/views/editor/panel.js
```diff
@@ -0,0 +1,19 @@
...
+    this.contentElement = this.getContentElement();
+    this.shadowRoot.appendChild(this.contentElement);
+  }
+
+  getContentElement() {}
+
+  attach() {}
+
+  detach() {}
```
justvanrossum, line 16: Let's remove `detach()`: it is currently not needed nor properly implemented.

src/fontra/views/editor/sidebar.js
```diff
@@ -0,0 +1,152 @@
...
+
+    tabOverlayContainer.appendChild(
+      html.div(
+        {
+          "class": "sidebar-tab",
+          "data-sidebarName": panel.name,
+        },
+        [
+          html.createDomElement("inline-svg", {
+            src: panel.icon,
```
justvanrossum, line 44: Please rename the .icon property to `.iconPath` (like in the edit tools)

src/fontra/views/editor/sidebar.js
```diff
@@ -0,0 +1,152 @@
...
+
+    const tabOverlayContainer = this.container.querySelector(
+      `.tab-overlay-container.${this.identifier}`
+    );
+
+    tabOverlayContainer.appendChild(
+      html.div(
+        {
+          "class": "sidebar-tab",
+          "data-sidebarName": panel.name,
```
justvanrossum, line 40: Please rename the .name property to `.identifier` (like in the edit tools)

justvanrossum (commented):
I wonder why we need the `panel.attach()` method at all. If we do, I prefer to rename it `setup()`. But if we don't really need it: I think all that setup should be done as part of the Panel subclass constructor. What do you think?

In editor.js, there are a bunch of init/setup methods, that contain stuff that should be part of the panel implementations:
- `initGlyphsSearch`
- `initUserSettings`
- `initSidebarDesignspace`

Similarly, there are some `toggleXxx` callback methods in editor.js that I think don't belong there anymore:
- `toggleGlyphSearch`
- `toggleTextEntry`
- `toggleSelectionInfo`
- `toggleUserSettings`
The Panel class should have a `toggle(onOff, doFocus)` method, that each subclass can override.

src/fontra/client/web-components/ui-form.js
```diff
@@ -5,6 +5,127 @@ import { SimpleElement } from "../core/unlit.js";
...
+      border-radius: var(--slider-thumb-border-radius);
+      background: var(--slider-thumb-color);
+      cursor: pointer;
+    }
+    input[type="range"]:focus::-ms-fill-lower {
+      background: var(--slider-track-color);
+    }
+    input[type="range"]:focus::-ms-fill-upper {
+      background: var(--slider-track-color);
+    }
```
justvanrossum, line 128: Please add a blank line here.

src/fontra/client/web-components/ui-form.js
```diff
@@ -1,23 +1,22 @@
 import { QueueIterator } from "../core/queue-iterator.js";
 import { hyphenatedToCamelCase, round } from "../core/utils.js";
-import { css } from "../third-party/lit.js";
+import { css, unsafeCSS } from "../third-party/lit.js";
```
justvanrossum, line 3: There is no need to use `css` or `unsafeCSS` with `SimpleElement`. We should not import lit unless we're actually using lit.

Or do you use `css` to trigger syntax highlighting in your editor?

PR #713, opened 2023-08-09
Increase range slider by ten if shift key is pressed

### LINKED ISSUE #706: <range-slider> should support shift-arrow-key increments of 10 units
It currently does steps of one on arrow key. It would be great if it could support steps of 10 by using shift-arrow-key.

Arrow-key nudging works when the focus is on the numeric value box, as well as on the slider itself once it has focus. Shift should work in both cases.

This relates to #467, except the info panel does not yet use `<range-slider>`. But it will.

fatih-erikli:
Fixes #706

### DIFF
--- a/src/fontra/client/web-components/range-slider.js
+++ b/src/fontra/client/web-components/range-slider.js
@@ -211,6 +211,7 @@ export class RangeSlider extends LitElement {
             type="range"
             @input=${this.changeValue}
             @mousedown=${this.handleMouseDown}
+            @keydown=${this.handleKeyDown}
             class="slider ${isAtDefault ? "is-at-default" : ""}"
             min=${this.minValue}
             max=${this.maxValue}
@@ -221,6 +222,29 @@ export class RangeSlider extends LitElement {
         </div>
       </section>
     `;
+  }
+
+  handleKeyDown(event) {
+    event.preventDefault();
+    let increment = event.shiftKey ? 10 : 1;
+    let newValue;
+    switch (event.key) {
+      case "ArrowLeft":
+      case "ArrowDown":
+        newValue = this.value - increment;
+        break;
+      case "ArrowRight":
+      case "ArrowUp":
+        newValue = this.value + increment;
+        break;
+    }
+    if (
+      newValue !== undefined &&
+      newValue > this.minValue &&
+      newValue < this.maxValue
+    ) {
+      this.value = newValue;
+    }
   }
 
   handleMouseDown(event) {


justvanrossum:
Ok, this fixes it _when the slider has focus_ (which is a nice bonus), but it needs to be fixed for when the _numeric input field_ has the focus.

src/fontra/client/web-components/range-slider.js
```diff
@@ -223,6 +225,29 @@ export class RangeSlider extends LitElement {
...
+      case "ArrowLeft":
+      case "ArrowDown":
+        newValue = this.value - increment;
+        break;
+      case "ArrowRight":
+      case "ArrowUp":
+        newValue = this.value + increment;
+        break;
+    }
+    if (
```
justvanrossum, line 242: It should clamp the value rather than ignore the event. Currently, if the max is 100, and I am at 99, arrow up/right doesn't make it go to 100, likewise if the min is 0 and I am at 1, with arrow down/left.

fatih-erikli: ok

justvanrossum:
While the increment/decrement now nicely works, the slider does not emit any change event when using the arrow keys: the scene does not update to reflect the new slider value.

fatih-erikli:
It should be fixed now @justvanrossum

src/fontra/client/web-components/range-slider.js
```diff
@@ -224,6 +226,27 @@ export class RangeSlider extends LitElement {
     `;
   }
 
+  handleKeyDown(event) {
+    event.preventDefault();
```
justvanrossum, line 230: We should _only_ do `preventDefault()` if we will handle the event. I think the switch needs a default branch which will simply return, and the preventDefault should be done after the switch.

justvanrossum: Perhaps we should return even earlier in case the control key, meta key or alt key are pressed.

justvanrossum: Numeric input is currently completely broken.

fatih-erikli: Yes, I missed that @justvanrossum.

justvanrossum: Please implement it as I suggested:
- Return immediately if the control key, meta key or alt key are pressed
- Add a "default" branch in the switch, in which we return

Then, the `newValue !== undefined` condition is no longer needed, too.

PR #714, opened 2023-08-09
Reference font dialog should not overflow

### LINKED ISSUE #699: [reference font] UI does shrink vertically when out of space
If you drop so many fonts on the reference font list that it runs out of space, the UI just falls off the page instead of shrinking the list.

fatih-erikli:
Fixes #699


fatih-erikli:
Upps, ok.

PR #723, opened 2023-08-13
Plugins draft

fatih-erikli:
Draft PR for plugins

I concern about `eval`, something like this this could be used https://github.com/NeilFraser/JS-Interpreter


fatih-erikli:
Plugin source codes
https://github.com/fatih-erikli/fontra-plugin-demo/blob/main/start.js

fatih-erikli:
Deleted the branch due to changes on #705.

PR #761, opened 2023-08-30
Add toggle-fullscreen control with temp icons.

fabiocaccamo:


justvanrossum:
For icons, I think I prefer the "minimize" and "maximize" icons from https://tabler-icons.io/

image

We can use these as-is (filename and all), and put them in the `/tabler-icons` folder. Perhaps we need to tweak the stroke thickness and size to match the other tool icons (with CSS).

justvanrossum:
Can you try adding CSS to tweak the icon stroke width for these? They are a bit heavy compared to the one directly to the left.

fabiocaccamo:
I think that the icon that has a different stroke width is the "bullseye", actually the stroke width of the fullscreen icons is the same of plus and minus icons.

justvanrossum:
It's still too heavy, even if I compare to plus and minus.

src/fontra/views/editor/editor.js
```diff
@@ -1574,6 +1583,46 @@ export class EditorController {
     this.sceneController.autoViewBox = false;
   }
 
+  toggleFullscreen() {
+    if (document.fullscreenElement) {
+      if (document.exitFullscreen) {
```
justvanrossum, line 1588: Is it possible for `document.fullscreenElement` to exist while `document.exitFullscreen` does not?

src/fontra/views/editor/editor.js
```diff
@@ -1574,6 +1583,46 @@ export class EditorController {
...
+        document.exitFullscreen();
+      }
+    } else {
+      const element = document.documentElement;
+      if (element.requestFullscreen) {
```
justvanrossum, line 1593: We can only get here is fullscreen mode is enabled. So, like my comment above, is it possible for `element.requestFullscreen` to not exist here? If so: I would write ` element.requestFullscreen?.()` and leave out the `if` condition.

src/fontra/views/editor/editor.js
fabiocaccamo: Right, they go hand in hand, so the inner condition can be removed.

justvanrossum (approved):

PR #771, opened 2023-09-03
Support building a .designspace system from scratch

justvanrossum:


src/fontra/backends/designspace.py
```diff
@@ -293,7 +326,15 @@ async def putGlyph(self, glyphName, glyph, unicodes):
         usedLayers = set()
         for layerName, layer in glyph.layers.items():
             layerName = revLayerNameMapping.get(layerName, layerName)
-            glyphSet = self.ufoLayers.findItem(fontraLayerName=layerName).glyphSet
+            ufoLayer = self.ufoLayers.findItem(fontraLayerName=layerName)
+
+            if ufoLayer is None:
+                # This layer is not used by any source.
```
justvanrossum, line 332: …and we didn’t see it during load

PR #773, opened 2023-09-04
Beginnings of fontra-copy tool

justvanrossum:
fontra-copy will be able to convert any file Fontra can read to a .designspace system (and later also to .rcjk).


justvanrossum:
I botched the branch, will start from scratch.

PR #777, opened 2023-09-04
Add tool-tip component / fixed botched branch

justvanrossum:
I created a new branch from the proper commits from #763, from before I force-pushed it to pieces.

This should supersede #763 then.


justvanrossum:
Please make the triangle larger (maybe 50% larger than it is now). I tried to fiddle with the CSS but I can't figure it out.

Please set the font size to 1em. (In most cases we don't use px in Fontra)

justvanrossum:
Can you make the triangle point more towards the middle of the button?

image

I will fiddle with the colors.

justvanrossum:
I have tweaked the colors a bit, and the timing.

While the current tooltips look good and work well, I worry about the alignment with different length texts, and how they can be cropped by the window.

I think the tooltips should ideally be centered at their target, but it could be off-center if it otherwise would be cropped by the window. Is somthing like that hard to do?

PR #778, opened 2023-09-04
 Styling + behavior of checkboxes in the source list

### LINKED ISSUE #767: Styling + behavior of checkboxes in the source list
Currently, the source list has two columns that use a checkbox. I would like them to be styled differently. Whether they should even remain an `<input type="checkbox">` is an open question.

image

I would like to use icons from the tabler collection instead.

For the "on" column (reflecting the "inactive" source attribute), I would like to use [circle-dot](https://tabler-icons.io/i/circle-dot) for the "on" state, and [circle-dotted](https://tabler-icons.io/i/circle-dotted) for the "off" state.

Additionally, I'd like this control to be triggered by double-click only, as apparently people sometimes accidentally click it without understanding the consequences.

For the "bg" column ("visible in background"), I would like to use [eye](https://tabler-icons.io/i/eye) for "on", and [eye-closed](https://tabler-icons.io/i/eye-closed) for "off". (This one should remain "single-click")

Finally, I'd like the "bg" header to be replaced by the "eye" icon. This may be a bit more complicated as the list header mechanism currently only supports text. I'm not yet 100% how it should be done, and should be a separate task.

When using tabler icons, we copy them into the [tabler-icons](https://github.com/googlefonts/fontra/tree/main/src/fontra/client/tabler-icons) folder.

fatih-erikli:
Work in progress. It will fix #767.


fatih-erikli:
I implemented that. I creates the whole list from empty div when an item on a list cell has changed, causes flickering when click or doubleclick. `ui-list` should render only the updated cell for such cases. I leave it as it is for further refactoring.

justvanrossum:
I will have a look at the flickering.

Please re-read the issue about which icons to use for the on-off column.

justvanrossum:
https://github.com/googlefonts/fontra/pull/792 fixed the flickering. Please rebase.

fatih-erikli:
I did rebase. Flickering is fixed.

PR #791, opened 2023-09-09
Refactor reference-font into single module

### LINKED ISSUE #747: [reference font] refactor into a single module
There's currently a Panel and a separate UI web component. The latter is redundant: everything from reference-font.js should move to panel-reference-font.js. Possibly still as a separate class, but there's no need for a web component inside the (Panel) web component.

fatih-erikli:
Fixes #747.


fatih-erikli:
I close this pr for now. I will open once #746 is merged.

PR #799, opened 2023-09-13
Work in progress: Implement MIN_CANVAS_SPACE

### LINKED ISSUE #786: Have the width of left and right panels respond to browser size
It'd be great to have the width of the left and right panels respond to the width of the browser.

Currently if I make the left and right panels relatively wide, and then reduce the size of the browser, the panels stay the same width and they overtake the whole browser. They also overlap, so it takes a moment to find the edges to adjust their widths.

Capture d’écran 2023-09-08 à 15 45 06

fatih-erikli:
Fixes #786.

### DIFF
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -32,6 +32,7 @@ import {
   readFromClipboard,
   reversed,
   writeToClipboard,
+  clamp,
 } from "../core/utils.js";
 import { themeController } from "/core/theme-settings.js";
 import { showMenu, MenuItemDivider } from "/web-components/menu-panel.js";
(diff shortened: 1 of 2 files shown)


justvanrossum:
The test failure is something I get regularly, and seems to be a glitch. Re-running the failed job usually succeeds. Do you have a clue what could be causing it?

src/fontra/views/editor/editor.js
```diff
@@ -54,7 +55,7 @@ import {
 import { staticGlyphToGLIF } from "../core/glyph-glif.js";
 import { pathToSVG } from "../core/glyph-svg.js";
 import { parseClipboard } from "../core/server-utils.js";
-import Sidebar from "./sidebar.js";
+import Sidebar, { MAX_SIDEBAR_WIDTH, MIN_SIDEBAR_WIDTH } from "./sidebar.js";
```
justvanrossum, line 58: Can we turn `Sidebar` into a "regular" export, instead of a default export? I would love the import line to read like this:

```suggestion
import { Sidebar, MAX_SIDEBAR_WIDTH, MIN_SIDEBAR_WIDTH } from "./sidebar.js";
```

fatih-erikli:
> The test failure is something I get regularly, and seems to be a glitch. Re-running the failed job usually succeeds. Do you have a clue what could be causing it?

Unit tests? Usually arrayExtend fails. I will try to understand next time it fails.

justvanrossum (approved):
Thanks!

PR #801, opened 2023-09-13
[context-menu] Fix wrong items vertical margins and left padding.

### LINKED ISSUE #800: [context menu] Wrong vertical margins before first item and after last item.
Screenshot 2023-09-13 at 15 43 09

fabiocaccamo:
Fix #800 

Before

Screenshot 2023-09-13 at 15 43 09

---

After

Screenshot 2023-09-13 at 15 56 30

### DIFF
--- a/src/fontra/client/web-components/menu-panel.js
+++ b/src/fontra/client/web-components/menu-panel.js
@@ -62,7 +62,7 @@ class MenuPanel extends SimpleElement {
       display: grid;
       grid-template-columns: 1em auto;
       gap: 0em;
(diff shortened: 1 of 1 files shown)


justvanrossum (commented):
Thanks! While this is indeed better, I find the top and bottom margins a wee bit too much now. I think `0.25em` for both is enough.

justvanrossum:
> I tried 0.25em, but I think the last item is too close to the bottom:

How about top 0.2, bottom 0.3?

fabiocaccamo:
I think 0.5 is better because the last item should have the same margins of an item in the middle.

This is 0.2 top, 0.3 bottom:

Screenshot 2023-09-13 at 17 34 53

PR #802, opened 2023-09-13
[context-menu] Choose item when release mouse after control-click and drag.

### LINKED ISSUE #287: Context menu doesn't work on Safari, secondary issue on Chrome/Ff

fabiocaccamo:
Fix #287


justvanrossum:
Awesome, thanks!

Issue #293, opened 2023-02-20
On Windows, scroll bars are always visible


justvanrossum:
Looks bad. Is this how Windows users expect this to appear, or can we improve our CSS?

![image](https://user-images.githubusercontent.com/4246121/220119232-8eeac351-8eb0-4a33-a12a-305ea4f9f7c3.png)

Issue #300, opened 2023-02-23
Sliders in sidebar-selection-info are too wide so scrollbar always shows


punchcutter:
When viewing rcjk files and looking at deep components the y scrollbar is always on in sidebar-selection-info. 

<img width="285" alt="slider-scrollbar-on" src="https://user-images.githubusercontent.com/11257204/220862802-78f449b8-76fb-49c6-b8d4-be3c36b7e22e.png">

If I turn it off manually it still seems really tight on the right.

<img width="306" alt="slider-scrollbar-off" src="https://user-images.githubusercontent.com/11257204/220862805-2b9bf4d2-d058-4ede-87da-92b639a32826.png">

With max value the sliders are cut off even when I turn off the scrollbar.

<img width="273" alt="slider-scrollbar-maxvalue" src="https://user-images.githubusercontent.com/11257204/220863852-b9b81ab0-2411-4206-9336-889840400298.png">

I also constantly try to expand the sidebar like I would expect possible in many apps so that would be nice. If I want to use the axis sliders to adjust anything I only have a tiny space to work with. Of course that's not a huge deal so could be a feature request, but the cut off and constant scrollbar has been bugging me.

justvanrossum:
May relate to #293.

justvanrossum:
Out of curiosity, on which platform and with which browser are you seeing this?

punchcutter:
Chrome on Mac.

punchcutter:
Checked Safari and Firefox and same thing.

justvanrossum:
What is your global scroll bar setting? (System Prefs -> General)
<img width="487" alt="image" src="https://user-images.githubusercontent.com/4246121/221128672-a8258081-cc64-4b4c-8747-1f05a34b2742.png">

In the past I think I've seen this setting also being influenced by the presence of a Wacom tablet.

justvanrossum:
I think I can reproduce your observations if I switch to "Always".

justvanrossum:
Solutions:

- Make the layout work even when scrollbars are always visible

Or

- Override the OS setting and always hide the scrollbars (if that's even possible with CSS)

punchcutter:
[...]

PR #804, opened 2023-09-13
[ui-list] Improve scrollbars behaviour.

fabiocaccamo:
This fixes part of #293.

We can assume that the ui-list is always a vertically scrollable list, but it doesn't need always to be scrolled, so these are the changes:

- Horizontal scrollbars are always hidden.
- Vertical scrollbars are visible only if necessary.

### DIFF
--- a/src/fontra/client/web-components/ui-list.js
+++ b/src/fontra/client/web-components/ui-list.js
@@ -26,8 +26,7 @@ export class UIList extends UnlitElement {
     }
 
     .container {
-      overflow-x: hidden;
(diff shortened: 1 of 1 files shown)


justvanrossum:
I will try this later, but I’d like to note that the source list is an example of a list that scrolls horizontally, too.

justvanrossum (approved):
This indeed fixes the Windows problem ffor ui-list, but not for other scrollable items, such as the ones mentioned in #293 and #300. Let's follow up in additional PRs.

PR #806, opened 2023-09-14
Show scrollbars only if needed.

### LINKED ISSUE #293: On Windows, scroll bars are always visible

### LINKED ISSUE #300: Sliders in sidebar-selection-info are too wide so scrollbar always shows

fabiocaccamo:
Fix #293
Fix #300

### DIFF
--- a/src/fontra/views/editor/editor.css
+++ b/src/fontra/views/editor/editor.css
@@ -334,7 +334,7 @@ body {
   display: flex;
   flex-direction: column;
   gap: 0.4em;
-  overflow: scroll;
+  overflow: auto;
   padding-bottom: 0.4em;
 }
 
--- a/src/fontra/views/editor/editor.css
+++ b/src/fontra/views/editor/editor.css
@@ -330,14 +330,6 @@ body {
   padding: 0.5em;
 }
 
-.axis-sliders {
-  display: flex;
-  flex-direction: column;
-  gap: 0.4em;
-  overflow: auto;
-  padding-bottom: 0.4em;
-}
-
 .canvas-container {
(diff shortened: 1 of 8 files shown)


src/fontra/views/editor/editor.css
```diff
@@ -334,7 +334,7 @@ body {
   display: flex;
   flex-direction: column;
   gap: 0.4em;
-  overflow: scroll;
```
justvanrossum, line 337: I noticed that this particular group (`.axis-sliders`) is no longer used anywhere. Please delete the whole block now that we're here anyway.

justvanrossum (commented):
There are a bunch of `overflow-y: scroll;` occurences that probably should become "auto", too.

justvanrossum:
Oh, and one `overflow-x: scroll`...

justvanrossum:
Thanks, I will try this on Windows.

justvanrossum (approved):
Excellent, thanks!

PR #841, opened 2023-09-21
[reference font] replace raw "language code" with a popup

### LINKED ISSUE #746: [reference font] replace raw "language code" with a popup

fatih-erikli:
Fixes #746.


justvanrossum:
Some initial feedback:
- lib-font seems to choke on  the GS subset font I sent you. I suspect a bug in lib-font :( It does indeed work with Noto.
- the select has a minimum width that depends on the contents, which breaks the layout of the sidebar:
image
- I would love for the language-mappings.js to become a json file, and that is becomes a dict, with "ot" as keys (I've verified these are unique in the list), and [name, html] as value (length-2 array, first item the name of the language, the second the bcp-47 code). I'm not sure we can use "import" in the browser to load the json, but utils contains a fetchJSON() function. Make sure it doesn't get loaded more than once.
- The parsing of the language systems should be initiated when the font is loaded, and not when it's selected, this will reparse every time you select the font. Maybe the result should simply be cached after the first time it is selected. (That cache does not have to be persistent.)
- Similarly, I'm not sure the "supportedLanguages" should be communicated via the model/controller.
- I'm not quite confident how we should do the last two items, so maybe we should discuss further.

justvanrossum:
Actually, the data file can remain unchanged. Let’s focus on the other items. 

In the meantime I’ll try to make a shareable test font for the lib-font bug.

justvanrossum:
I tried to reproduce the lib-font error with Node, but it seems to work fine. I can share my script later.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -320,6 +322,38 @@ export default class ReferenceFontPanel extends Panel {
...
+      this.model.supportedLanguages = LANGUAGE_MAPPING.filter((l) =>
+        allLangs.has(l.ot)
+      );
+    };
+
+    const reader = new FileReader();
+    reader.addEventListener("load", () => {
+      font.fromDataBuffer(reader.result);
+    });
+    reader.readAsArrayBuffer(await readFontFileFromOPFS(fontItem.fontIdentifier));
```
justvanrossum, line 356: Have you tried to do it without FileReader, like this: `font.fromDataBuffer(await readFontFileFromOPFS(…))`

But either way, we need to do this at load time, so we don’t have to read the data twice.

fatih-erikli: probably FileReader won't be needed if readFontFileFtomOPFS returns an array buffer. I should have missed that.

> But either way, we need to do this at load time, so we don’t have to read the data twice.

It must be loaded when a selected font change.

justvanrossum: > It must be loaded when a selected font change

Why would that be? We can do it at load, keep the result in a dict.

justvanrossum:
Ah, the problem I'm seeing is only on first upload. Once the file is stored in the OPFS it works fine (after reloading the page).

That first time, I suspect `readFontFileFromOPFS()` is called before the write to the OPFS has completed.

This also suggests to me that we should do the language parsing on load.

src/fontra/views/editor/panel-reference-font.js
fatih-erikli: I kept the selected languages in `fontItem` object with the others `uploadedFileName, fontIdentifier`. Is that ok?

justvanrossum:
Please change the first item to say "None", and add the html/bcp-47 code in parentheses to each item, just like in wakamaifondue:
image

The "None" menu is needed to we can always go back to "no language".

justvanrossum:
- Please change the "Language code:" label to "Language:"
- Reloading the page does restore the language code, but the select menu is reset: it should reflect the previously chosen language
- When switching fonts, if the currently chosen language is _also_ supported by the newly selected font, the language setting should not be reset (and the select should show the languages supported by the newly selected font, with the right one selected)

PR #849, opened 2023-09-26
Replace raw language code with select

### LINKED ISSUE #746: [reference font] replace raw "language code" with a popup

fatih-erikli:
Fixes #746.


fatih-erikli:
These are done

- I would love for the language-mappings.js to become a json file, and that is becomes a dict, with "ot" as keys (I've verified these are unique in the list), and [name, html] as value (length-2 array, first item the name of the language, the second the bcp-47 code). I'm not sure we can use "import" in the browser to load the json, but utils contains a fetchJSON() function. Make sure it doesn't get loaded more than once.
- The parsing of the language systems should be initiated when the font is loaded, and not when it's selected, this will reparse every time you select the font. Maybe the result should simply be cached after the first time it is selected. (That cache does not have to be persistent.)
- Similarly, I'm not sure the "supportedLanguages" should be communicated via the model/controller.
> Similarly, I'm not sure the "supportedLanguages" should be communicated via the model/controller.
I was thinking about this and I ended up keeping supported languages in model. When a page reloaded, the reference font panel isn't initialized with a selected font. To restore the previously selected language, we need them to be persisted in localstorage.
- Actually, the data file can remain unchanged. Let’s focus on the other items.
> I did it already. I keep it in the way you suggested. I think it's better to have a json file rather than a javascript module.
- Have you tried to do it without FileReader, like this: font.fromDataBuffer(await readFontFileFromOPFS(…))
> Not like the way you suggested, because readFontFileFromOPFS returns a File instance, lib-font expects an array buffer. To convert a file into an array buffer, I used `FileReader`. I discovered another way to load the font, `fontItem.objectUrl` returns a blob url stored in the browser. I load that fetch and pass it to the File Reader. But I just noticed that there is an easier way, giving this url to lib-font works well.
https://github.com/googlefonts/fontra/pull/849/commits/dc444847ff3c2048f38b41e34a47444334bf8724
> Please change the first item to say "None", and add the html/bcp-47 code in parentheses to each item, just like in wakamaifondue
> Reloading the page does restore the language code, but the select menu is reset: it should reflect the previously chosen language
> When switching fonts, if the currently chosen language is also supported by the newly selected font, the language setting should not be reset (and the select should show the languages supported by the newly selected font, with the right one selected)

justvanrossum:
> Out of the lib-font context, most of the time my browser freezes when I open multiple fontra instances

Please open an issue and describe exactly what you do, and with which browser(s).

justvanrossum:
The select seems to be sorted by bcp-47 code. Perhaps it should be sorted by the "friendly" name instead? What do you think?

fatih-erikli:
Yes, that makes sense

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -378,8 +442,13 @@ export default class ReferenceFontPanel extends Panel {
     this.controller.addKeyListener("fontList", (event) =>
       this._fontListChangedHandler(event)
     );
+    this.controller.addKeyListener("supportedLanguages", () => {
```
justvanrossum, line 445: I don't think we should put supportedLanguages on the model, as this will also make it save to localStorage.

What works equally well is to delete the key listener, and call setSupportedLanguages instead of this line:
```js
 this.model.supportedLanguages = fontItem.supportedLanguages;
```

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -59,7 +61,44 @@ function cleanFontItems(fontItems) {
     return {
       uplodadedFileName: fontItem.uplodadedFileName,
       fontIdentifier: fontItem.fontIdentifier,
+      supportedLanguages: fontItem.supportedLanguages,
+    };
+  });
+}
+
+async function readSupportedLanguages(fontItem, languageMapping) {
```
justvanrossum, line 69:
```suggestion
function readSupportedLanguages(fontItem, languageMapping) {
```
You are returning a promise, so it is async in that way, and there is no need to declare this func as async.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -59,7 +61,44 @@ function cleanFontItems(fontItems) {
     return {
       uplodadedFileName: fontItem.uplodadedFileName,
       fontIdentifier: fontItem.fontIdentifier,
+      supportedLanguages: fontItem.supportedLanguages,
```
justvanrossum, line 64: In the context of my other feedback, this line should be removed:
```suggestion
```

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -366,9 +412,27 @@ export default class ReferenceFontPanel extends Panel {
...
+        )
+      );
+    }
+  }
+
   getContentElement() {
+    this.allLanguages = {};
     this.controller = new ObservableController({
       languageCode: "",
+      supportedLanguages: [],
```
justvanrossum, line 435: See below, this line can be removed.
```suggestion
```

src/fontra/views/editor/editor.html
```diff
@@ -5,6 +5,8 @@
     <title>Fontra</title>
     <link href="/css/core.css" rel="stylesheet" />
     <link href="/editor/editor.css" rel="stylesheet" />
+    <script defer src="/third-party/lib-font/inflate.js"></script>
```
justvanrossum, line 8: Can we import these in panel-reference-font.js? They are quit out of place here.

rollup.config.js
```diff
@@ -2,15 +2,26 @@ import commonjs from "@rollup/plugin-commonjs";
...
+      },
+      plugins: [
+        commonjs(), // <-- this handles some parsing of js syntax or something (necessary for `export { init } from "mathjax";`)
+        nodeResolve({ browser: true }), // <-- this allows npm modules to be added to bundle
+      ],
+    };
+  })
+  .concat({
+    input: [
+      "node_modules/lib-font/lib/inflate.js",
```
justvanrossum, line 21: I'm not really happy with this special case for lib-font here, but I do not yet have a concrete proposal how to do it differently. Once I do, I will suggest a followup task.

justvanrossum:
One more thing: for each reference font I see a lib-font warning in the console:
image
Even though the font obviously loaded correctly. Do you also see that also? Can you try to figure out what causes it?

src/fontra/views/editor/panel-reference-font.js
fatih-erikli: This is strange. The reference font list wasn't loading with the previously selected font. Now it does. This is why I cached it in localstorage.

I will apply your suggestion.

src/fontra/views/editor/panel-reference-font.js
fatih-erikli: Thanks. Previously I was using await inside that function. Looks like I forgot to delete async keyword when it evolved to the current version.

src/fontra/views/editor/panel-reference-font.js
fatih-erikli: I did it because cleanFontItems is called when a font item is deleted in the panel.
I will delete supportedLanguages from fontItem. It's complicating the process. I will probably cache them in panel instance.

fatih-erikli:
About the warning, it's preparing the font to be read as a web font by default. Setting `skipStyleSheet` `true` disabled that, the warning no longer exist.

src/fontra/views/editor/panel-reference-font.js
justvanrossum: Ah I see, I think I misunderstood what you were doing here. Let's keep this field in the "clean" font item. I mixed it up with adding the *current* supported languages to the model.

src/fontra/views/editor/panel-reference-font.js
fatih-erikli: It happens when multiple tabs of fontra is open. I applied your suggestion.

src/fontra/views/editor/panel-reference-font.js
fatih-erikli: I deleted it, it confuses the reader including me after looking at the code a couple of days later. I store the supported languages in panel instance.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -320,6 +360,24 @@ export default class ReferenceFontPanel extends Panel {
       await fontItem.fontFace.load();
     }
     this.model.referenceFontName = fontItem.fontIdentifier;
+
+    if (Object.hasOwn(this.supportedLanguagesMemoized, fontItem.fontIdentifier)) {
```
justvanrossum, line 364: I think this is good enough in this case, and reads better (to a Python reader at least):
```suggestion
    if (fontItem.fontIdentifier in this.supportedLanguagesMemoized) {
```

rollup.config.js
fatih-erikli: This is a special case about how the lib-font is bundled. I am not happy with that, at all.

src/fontra/views/editor/editor.html
fatih-erikli: I did, it looks clearer.

src/fontra/views/editor/panel-reference-font.js
justvanrossum: The font identifier contains a random UUID, so this is a moot point. Please use "in" like I suggested.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -3,11 +3,20 @@ import Panel from "./panel.js";
...
 
 import { ObservableController } from "/core/observable-object.js";
-import { div, input, label } from "/core/unlit.js";
-import { fileNameExtension, withTimeout } from "/core/utils.js";
+import { div, input, label, option, select } from "/core/unlit.js";
+import { fetchJSON, fileNameExtension, withTimeout } from "/core/utils.js";
 import { dialog } from "/web-components/modal-dialog.js";
 import { UIList } from "/web-components/ui-list.js";
 
+// prettier-ignore
```
justvanrossum, line 11: What does prettier do with it?

fatih-erikli: Prettier lift downs these imports to after the line that the "Font" is imported. lib-font wants them to be imported before the main module is imported.

src/fontra/views/editor/panel-reference-font.js
fatih-erikli: Ok, I thought it is the font name.

src/fontra/views/editor/panel-reference-font.js
justvanrossum: What is the status of this? I'm not sure I understand the issue, if there still is one. Can you give concrete steps about what to do and what to observe?

src/fontra/views/editor/panel-reference-font.js
justvanrossum: Ahh, that's nasty. Thanks!

justvanrossum: Please add a comment to that effect.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -3,11 +3,18 @@ import Panel from "./panel.js";
...
-import { fileNameExtension, withTimeout } from "/core/utils.js";
+import { div, input, label, option, select } from "/core/unlit.js";
+import { fetchJSON, fileNameExtension, withTimeout } from "/core/utils.js";
 import { dialog } from "/web-components/modal-dialog.js";
 import { UIList } from "/web-components/ui-list.js";
 
+import "/third-party/lib-font/inflate.js";
+import "/third-party/lib-font/unbrotli.js";
+
+// lib-font expect it's dependencies to be imported. prettier lifts the imports up
```
justvanrossum, line 14:
```suggestion
// lib-font expects its dependencies to be imported first. Prettier moves the imports.
```

Issue #782, opened 2023-09-06
Need proof-of-concept for externally hosted sidebar plug-in


justvanrossum:
Requirements/ideas:
- Starting point is a URL to a .json file hosted on a GitHub repo
- The .json file should reference a .js file in the same repo, using a relative path, or filename
- Via https://www.jsdelivr.com/github we can turn the .js file into an importable module
- Fontra uses dynamic import to import the module (for example `module = await import(moduleURL);`)
- Fontra calls a function in the module (the name of the function could be specified in the .json file)
- The main EditorController instance is passed to the function
- The function can then call `addSidebarPanel()`, and set up a demo sidebar panel.
- The demo sidebar panel doesn't need to do much, just display a bit of static text would be enough

fatih-erikli:
This is ready in plugins-poc branch.

https://github.com/googlefonts/fontra/commit/04911352cc5565a95656f2c33881f2e528c0a94a

External plugin:

https://github.com/fatih-erikli/fontra-plugin-demo

justvanrossum:
This was implemented in #862.

PR #862, opened 2023-09-30
Plugins proof of concept

fatih-erikli:
#782


justvanrossum:
Can you post the link to your demo plugin here?

fatih-erikli:
https://github.com/fatih-erikli/fontra-plugin-demo

justvanrossum:
Next step: develop a workflow how to write and debug plugins locally.

pyproject.toml
```diff
@@ -59,7 +60,7 @@ filesystem = "fontra.filesystem.projectmanager:FileSystemProjectManagerFactory"
 
 [project.entry-points."fontra.webcontent"]
 filesystem = "fontra.filesystem"
-
+localplugins = "fontra.localplugins"
```
justvanrossum, line 63: There were two blank lines after this before: please keep the formatting the same.

src/fontra/client/web-components/plugin-manager.js
```diff
@@ -0,0 +1,167 @@
...
+      this.pluginsEmptyState.style.display = "unset";
+    } else {
+      this.pluginsEmptyState.style.display = "none";
+    }
+  }
+
+  render() {
+    const fragment = document.createDocumentFragment();
+    fragment.appendChild(
+      html.div({ class: "buttons" }, [
```
justvanrossum, line 120: Please use the `add-remove-buttons` web component for this, and place it underneath the list. This is what Fontra does elsewhere (sources list, axes list).

src/fontra/views/editor/editor.html
```diff
@@ -5,6 +5,13 @@
     <title>Fontra</title>
     <link href="/css/core.css" rel="stylesheet" />
     <link href="/editor/editor.css" rel="stylesheet" />
+    <script type="importmap">
```
justvanrossum, line 8: I thought you discovered importmap is redundant. In your remote demo you still do
```js
import { objectsEqual } from "fontra/core/utils.js";
```
So it will work if you would do this instead:
```
import { objectsEqual } from "/core/utils.js";
```
Right? If so, let's remove the importmap.

src/fontra/views/editor/editor.html
```diff
@@ -72,7 +79,6 @@
       </div>
     </div>
   </body>
-
```
justvanrossum, line 75: Please don't edit unrelated whitespace.

src/fontra/views/editor/sidebar.js
```diff
@@ -24,7 +24,7 @@ export class Sidebar {
 
     const panelContent = html.div(
       { "class": "sidebar-content", "data-sidebarName": panel.identifier },
-      [panel]
+      [panel instanceof HTMLElement ? panel : panel.contentElement]
```
justvanrossum, line 27: I'm still not happy with this switch. Can you make the demo plugin work with the original `[panel]` line?

We can also consider change the `addSidebarPanel()` call signature to take three arguments:
1. the config object (containing `iconPath` etc), which may _also_ be the content element
2. the sidebar "name" (left/right)
3. *optionally* a separate content element

How does that sound to you?

fatih-erikli: We discussed about defining the panel in remote plugin declare as a web component in our call. Do you still want that, or you want the second one? Web component in remote panel sounds like a better solution to me. The panels will have their own shadow root in that way. The switch you're not happy with will be deleted.

src/fontra/views/editor/editor.html
fatih-erikli: Yes, ok. I was unsure if we treat it as an alias, to make it clear we import something from fontra. It could be helpful if the plugins are going to have external packages in future.

But I delete it. Better not to have an alias if it's importable by without import map (it's), to avoid possible confusion. Packaging is always confusing.

src/fontra/views/editor/editor.html
fatih-erikli: Ok

src/fontra/views/editor/sidebar.js
justvanrossum: I think it would be nice to support both, which is possible with my suggestion. As your demo shows, it is not always needed nor convenient to write a web component. So yes, please continue to implement my suggestion.

fatih-erikli:
Turns out import maps is necessary. When developing it locally, imports work, when importing it from jsdelivry `import('/core/utils.js')` is trying to be resolved like `https://cdn.jsdelivr.net/core/utils.js`.

fatih-erikli:
I propose to add a proxy endpoint to fontra backend instead of using jsdelivr. The version tag `latest` pointing to the older version, for 30 minutes, the cache still valid. It fetches the latest version when giving the latest version tag manually.

justvanrossum:
> I propose to add a proxy endpoint to fontra backend instead of using jsdelivr. The version tag latest pointing to the older version, for 30 minutes, the cache still valid. It fetches the latest version when giving the latest version tag manually.

Are you sure you are looking at the jsdelivr cache, and not browser cache?

jsdelivr has this tool to manually purge the cache: https://www.jsdelivr.com/tools/purge

fatih-erikli:
> Are you sure you are looking at the jsdelivr cache, and not browser cache?

I did cmd+shift+refresh, added random query parameters to the end. Result was the same. It's not a browser cache.

> jsdelivr has this tool to manually purge the @latest cache: https://www.jsdelivr.com/tools/purge

Do you think it will help us?

src/fontra/views/editor/sidebar.js
fatih-erikli: Creating something like method signature overloading is tricky in Javascript. I still think about how to make it clearly.

fatih-erikli: I create panel component when this method is called. I am not sure if we should do it in that way but I couldn't find a better solution either.

justvanrossum: I don't understand. It's not method signature overloading and it is not tricky. There is simply a third argument that is optional. If it is not given, it is equal to undefined. We separate the panel config (id, icon path) from the content element. If the third argument is undefined, we use the first argument for this instead (so one object is both the config _and_ the contents). This may have to propagate to`sidebar.addPanel()` panel as well. There should be no need for trickeries like creating a subclass on the fly.

Am I missing something? Perhaps we should have a call to figure this out.

justvanrossum:
> I am asking not for myself but for the plugin developers. We can't ask them to use that tool.

I disagree, we totally can. You develop/debug locally, sync with repo, purge cache, test remote version, done.

> Another thing I concern, without an internet connection, remote plugins won't be working. 

Correct. That is a consequence that I am for now willing to live with. Our (black-fourndry-internal) important use cases already require people to be online.

> I guess we will need to download/cache them with the version number. When the internet connection is on again, we will need to check the latest version and download if a newer version is available.

So at least for now, we don't need such complications.

src/fontra/views/editor/sidebar.js
fatih-erikli: Please read the commit, you will see what I'm talking about.

fatih-erikli:
> I disagree, we totally can. You develop/debug locally, sync with repo, purge cache, test remote version, done.

Ok.

fatih-erikli:
Method signature taking three arguments rolled back as we discussed. I kept your suggestions like replacing `panels` with `panelIdentifiers` in sidebar.

fatih-erikli:
Demo plugin is here https://github.com/fatih-erikli/fontra-plugin-demo

fatih-erikli:
Converting identifier and iconPath as static parameters in Panel can help us to declare them in addSidebarPanel if it's not defined with `customElements.define`. In that way plugin will not need to duplicate the identifier in two places.

src/fontra/views/plugins/plugins.html
```diff
@@ -0,0 +1,14 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link href="/css/core.css" rel="stylesheet" />
+    <script type="module" src="/web-components/plugin-manager.js"></script>
+    <title>Plugins</title>
```
justvanrossum, line 8:
```suggestion
    <title>Fontra Plugins</title>
```

src/fontra/client/web-components/plugin-manager.js
```diff
@@ -0,0 +1,158 @@
...
+  constructor() {
+    super();
+    this.contentElement = this.shadowRoot.appendChild(html.div());
+    this.contentElement.classList.add("plugin-manager");
+    const observable = new ObservableController({
+      plugins: [],
+    });
+    observable.synchronizeWithLocalStorage("fontra.plugins");
+    this.observable = observable;
+    this.pluginList = new UIList();
```
justvanrossum, line 34: Adding
```js
    this.pluginList.minHeight = "5em";
```
after this line makes the list not disappear when it is empty. I think that's better in this case.

Perhaps we should add a simple header "Fontra plugins:" just before the list.

I think you could then get rid of the "No plugins added." field.

fatih-erikli: Ok. I did. "No plugins added" deleted. Looks better.

src/fontra/client/web-components/plugin-manager.js
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  renderPlugins() {
+    const plugins = this.observable.model.plugins;
+    this.pluginList.setItems(plugins);
+  }
+
+  render() {
+    const fragment = document.createDocumentFragment();
+    fragment.appendChild(html.div({}, ["Plugin list:"]));
```
justvanrossum, line 116:
```suggestion
    fragment.appendChild(html.div({}, ["Fontra plugins:"]));
```

fatih-erikli:
I reproduce the regression. It's only reproducible in the current branch as you say. It's strange, I investigate.

src/fontra/client/css/plugins.css
```diff
@@ -0,0 +1,9 @@
+.plugins-page {
+  padding: 2rem;
+  font-size: 1.1rem;
```
justvanrossum, line 3: This has the effect of the Dialog _also_ being 10% larger than normal. I think the scope of the font size should be smaller.

src/fontra/client/web-components/plugin-manager.js
```diff
@@ -0,0 +1,149 @@
...
+      "Provide a github handle/repository",
+      "",
+      [
+        { title: "Cancel", resultValue: "no", isCancelButton: true },
+        { title: "Create", resultValue: "ok", isDefaultButton: true },
+      ]
+    );
+    let address = text;
+    newPluginPrompt.setContent(
+      html.input({
```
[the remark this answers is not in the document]
justvanrossum: My comment was not so much about the label (although it is probably an improvement, too), but about the styling of the input field itself. Notice the background color, the font and the focus ring:

image

image

src/fontra/client/web-components/plugin-manager.js
```diff
@@ -0,0 +1,164 @@
...
+        parsePluginBasePath(address)
+      );
+      if (ok) {
+        this.observable.setItem("plugins", [
+          ...this.observable.model.plugins,
+          newPlugin,
+        ]);
+        this.renderPlugins();
+      } else {
+        await dialog("Error", errorMessage, [{ title: "Ok", isDefaultButton: true }]);
```
justvanrossum, line 98:
```suggestion
        await dialog("Error", errorMessage, [{ title: "Okay", isDefaultButton: true }]);
```

PR #929, opened 2023-10-26
Menu bar

### LINKED ISSUE #1027: Add titlebar / menubar
This is work in progress: #929

But we're a bit stuck in defining the exact requirements and contents. This needs to be discussed an determined.

fatih-erikli:
Fixes #1027


justvanrossum:
Great start,  thanks!

I haven't reviewed the code, but playing with it I notice two things:
- the font size for the menu titles is too small, and should probably match those tof the menu-panel items
- hovering the menu title only works once the menu bar has focus, or at least has been clicked once. I think it should always hover.

justvanrossum:
Ok, trying again, and comparing with google docs: the behavior is fine, except the menu title should also highlight when hovering over.

fatih-erikli:
> hovering the menu title only works once the menu bar has focus, or at least has been clicked once. I think it should always hover.

Ok. I copied the top bar behavior of macos.

fatih-erikli:
I did force push to rebase.

justvanrossum:
This looks good, but I would like the `items` list to not be static. I think the `items` field should be a function instead, and be called `getItems`. It is then also important that this `getItems` function is only called when the menu is displayed, so that it really gets populated on-demand.

The same goes for sub-menu: a sub menu should be populated only right before it is displayed.

fatih-erikli:
Not ready for review yet.

justvanrossum (commented):
Good progress, I like the onSelect solution.

One behavioral thing: if I click a menu, and then move around in the menu bar, the appropriate menus show. However, if I go beyond the last item of before the first, I expect any unfolded menu to disappear, yet it stays.

fatih-erikli:
I don't understand the steps. Will you record a video? @justvanrossum

justvanrossum:
We've decided on the next steps for this PR:

Please add a menu bar to editor.js, so we can fine tune the appearance in the editor context. We need to see what color scheme works best, or otherwise how to make sure there is somehow a division between the menu bar and the rest of the window.

Here is an initial set of menus:
```
File       Edit       View          Project        Glyph          Extensions       Help
---------  ---------  ------------  -------------  -------------  ---------------  -------------
New...     Undo       Zoom in       Font info      Add source     Plug-in manager  Documentation
Open...    Redo       Zoom out      Font overview  Add glyph axis
           Etc.       Zoom to fit   
```

fatih-erikli:
I noticed one problem.

https://github.com/googlefonts/fontra/assets/92447202/a84ea8da-b43a-410f-97d3-e14872d38458

The open menu should be dismissed when clicked on the open menu second time.

fatih-erikli:
The defect in the video should be fixed with my latest commit.

justvanrossum:
For now, let's do three links under "Help":
- Homepage: https://fontra.xyz/
- Documentation: https://gferreira.github.io/fontra-docs/
- GitHub: https://github.com/googlefonts/fontra

justvanrossum:
Thanks for the refinements!

Further feedback:
- The "Edit" menu doesn't work when a glyph is in edit mode
- Glyph menu: add "Delete source menu...", "Edit local axes..."
- Select previous/next source and Find glyphs that use ... should go under the Glyph menu
- All menus that show a dialog (such as "Add source") should have titles that end with "..."
- The Add source and Delete source menu's need dymanic enable/disable

justvanrossum:
> Select previous/next source and Find glyphs that use ... should go under the Glyph menu

Or would they be better under "View"?

fatih-erikli:
I noticed one little bug, `Esc` key closes the open menu panel, but the panels keeps opening when hovering on other menu items.

justvanrossum:
> I noticed one little bug, Esc key closes the open menu panel, but the panels keeps opening when hovering on other menu items.

Can you fix it?

fatih-erikli:
Yes I will fix that but there's a UI consistency, I think we should delete the hover states at first. Hovers should be active when a menu is open.

justvanrossum:
> Select previous/next source and Find glyphs that use ... should go under the Glyph menu

Like I commented later, I think they actually should go under "View".

src/fontra/client/web-components/menu-panel.js
```diff
@@ -86,50 +91,97 @@ class MenuPanel extends SimpleElement {
...
     }
+
+    .submenu-icon {
+      width: 10px;
+      height: 14px;
+    }
   `;
 
-  constructor(menuItems, position, positionContainer) {
+  constructor(
```
justvanrossum, line 101: Can we simplify the constructor? I'm thinking all but the first arguments should go into an options object:

```js
  constructor(menuItems, options) {
```
Since many of these are optional this should simplify things. Also, it makes it easier to add more if we need.

fatih-erikli: Yes, you're right.

justvanrossum (approved):
Let's merge this! Looking good to me.

PR #934, opened 2023-10-27
Get rid of attach()

### LINKED ISSUE #920: Small sidebar Panel refactoring: get rid of attach()
All sidebar panel attach() methods can be replaced either with functionality that can be done in the constructor, or, if it does something that needs to be deferred to "attach" time, it can do it in a `connectedCallback()` method, which is a standard web component callback that gets called automatically.

fatih-erikli:
Fixes #920.

### DIFF
--- a/src/fontra/views/editor/panel-designspace-navigation.js
+++ b/src/fontra/views/editor/panel-designspace-navigation.js
@@ -121,7 +121,7 @@ export default class DesignspaceNavigationPanel extends Panel {
     );
   }
 
-  attach() {
+  connectedCallback() {
     this.fontController = this.editorController.fontController;
     this.sceneSettingsController = this.editorController.sceneSettingsController;
     this.sceneSettings = this.editorController.sceneSettingsController.model;
--- a/src/fontra/views/editor/panel-designspace-navigation.js
+++ b/src/fontra/views/editor/panel-designspace-navigation.js
@@ -80,6 +80,19 @@ export default class DesignspaceNavigationPanel extends Panel {
     }
   `;
 
+  constructor(editorController) {
+    super(editorController);
+    this.fontController = this.editorController.fontController;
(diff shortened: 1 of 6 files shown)


src/fontra/views/editor/panel-designspace-navigation.js
```diff
@@ -121,7 +121,7 @@ export default class DesignspaceNavigationPanel extends Panel {
     );
   }
 
-  attach() {
+  connectedCallback() {
```
justvanrossum, line 124: This should be done in the constructor. Please only use connectedCallback() for things that _must_ be done at connect time.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -544,7 +544,7 @@ export default class ReferenceFontPanel extends Panel {
     );
   }
 
-  attach() {
+  connectedCallback() {
```
justvanrossum, line 547: This can also be done in the constructor.

src/fontra/views/editor/panel-selection-info.js
```diff
@@ -51,7 +51,7 @@ export default class SelectionInfoPanel extends Panel {
     }
   }
 
-  attach() {
+  connectedCallback() {
```
justvanrossum, line 54: This can also be done in the constructor.

src/fontra/views/editor/panel-text-entry.js
```diff
@@ -188,7 +188,7 @@ export default class TextEntryPanel extends Panel {
     this.textEntryElement.focus();
   }
 
-  attach() {
+  connectedCallback() {
```
justvanrossum, line 191: This can also be done in the constructor.

fatih-erikli:
I deleted connected callback methods. Moved them to the constructors. All the panels work as expected. I don't remember why it was needed still.

PR #952, opened 2023-11-01
Delete lit dependency on range-slider

### LINKED ISSUE #939: Rewrite range-slider to not use lit

fatih-erikli:
Fixes #939.

### DIFF
--- a/src/fontra/client/web-components/range-slider.js
+++ b/src/fontra/client/web-components/range-slider.js
@@ -158,6 +158,7 @@ export class RangeSlider extends html.UnlitElement {
     // Fallbacks for attributes that are not defined when calling the component
     this.minValue = 0;
     this.maxValue = 100;
+    this.defaultValue = this.minValue;
     this.value = this.defaultValue;
     this.step = 1;
     this.sawMouseDown = false;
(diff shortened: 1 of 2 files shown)


justvanrossum:
To avoid unnecessary rerendering, should we perhaps implement a setter/getter for value, instead of doing the automatic render via the properties list?

fatih-erikli:
Ok. I will try that.

fatih-erikli:
To avoid such rerenders, we may consider using a simple dom diff-patch library in future. This looks like a clean implementation of what react, vue do. https://github.com/tjcchen/simple-virtual-dom/blob/main/src/main.js#L49-L52

fatih-erikli:
I don't know how lit solves this problem.

src/fontra/client/web-components/range-slider.js
```diff
@@ -162,169 +160,172 @@ export class RangeSlider extends LitElement {
...
-          />
-        </div>
-      </section>
-    `;
+    return round(this.value, decimalPlaces);
+  }
+
+  set value(value) {
+    this._value = value;
+    if (this.rangeInput) this.rangeInput.value = value;
```
justvanrossum, line 177: Please no one-line if statements: use a proper { } block. Same for the line below.

(We need to check if there's a prettier option for that)

PR #966, opened 2023-11-06
Wait for plugins to load if the stored panel name is not builtin

### LINKED ISSUE #965: Restore sidebar panel selection earlier
If the stored selected sidebar panel is "built in" and not coming from a plugin, it should be possible to activate it before the plugins are loaded. This would be a refinement for #936, #963.

fatih-erikli:
Fixes #965.

### DIFF
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -183,9 +183,29 @@ export class EditorController {
     this.initContextMenuItems();
     this.initShortCuts();
     this.initMiniConsole();
-    this.initPlugins().then(() => {
-      this.restoreOpenTabs("left");
-      this.restoreOpenTabs("right");
+
+    let waitPluginsToRestoreTabs = [];
+    for (const sidebar of this.sidebars) {
+      const panelName = localStorage.getItem(
+        `fontra-selected-sidebar-${sidebar.identifier}`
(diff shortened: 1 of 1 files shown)


src/fontra/views/editor/editor.js
```diff
@@ -183,9 +183,29 @@ export class EditorController {
     this.initContextMenuItems();
     this.initShortCuts();
     this.initMiniConsole();
-    this.initPlugins().then(() => {
-      this.restoreOpenTabs("left");
-      this.restoreOpenTabs("right");
+
+    let waitPluginsToRestoreTabs = [];
```
fatih-erikli, line 187: I'll need a better variable name

fatih-erikli: This holds sidebar names

justvanrossum: deferredPanels maybe?

Could it be a const?

justvanrossum: Yes, it can be a const.

justvanrossum: Ah, I know see I misunderstood what waitPluginsToRestoreTabs actually stored. Maybe `deferRestoreOpenTabs`.

PR #971, opened 2023-11-08
Delete lit dependency of range slider

fatih-erikli:
Includes fixes for the revert of #952.


fatih-erikli:
I will apply include discrete values functionality in this PR, if you don't have an objection. @justvanrossum

justvanrossum:
Rounding to a limited number of digits doesn't work:
image

PR #972, opened 2023-11-08
Implement discrete values for range-slider

fatih-erikli:
Implements #193.

It's done, but I left test values for testing.

### DIFF
--- a/src/fontra/client/web-components/designspace-location.js
+++ b/src/fontra/client/web-components/designspace-location.js
@@ -111,12 +111,11 @@ export class DesignspaceLocation extends UnlitElement {
         elements.push(html.hr());
         continue;
       }
-      const values = axis.values;
       const modelValue = this.values[axis.name];
       const infoBox = htmlToElement(
         `<div class="info-box">
           ${
-            values && values.length
+            axis.values && axis.values.length > 0
               ? `
           <span>Default: <strong>${axis.defaultValue}</strong></span>&nbsp; |
           <span>Values: <strong style="white-space: break-spaces;">${axis.values.join(
@@ -143,7 +142,7 @@ export class DesignspaceLocation extends UnlitElement {
       const slider = html.createDomElement("range-slider", {
         minValue: axis.minValue,
         maxValue: axis.maxValue,
-        values,
+        values: axis.values,
         defaultValue: axis.defaultValue,
         value: modelValue !== undefined ? modelValue : axis.defaultValue,
         onChangeCallback: (event) =>
--- a/src/fontra/client/web-components/designspace-location.js
+++ b/src/fontra/client/web-components/designspace-location.js
@@ -119,7 +119,7 @@ export class DesignspaceLocation extends UnlitElement {
             values && values.length
               ? `
           <span>Default: <strong>${axis.defaultValue}</strong></span>&nbsp; |
-          <span>Possible values: <strong style="white-space: break-spaces;">${axis.values.join(
+          <span>Values: <strong style="white-space: break-spaces;">${axis.values.join(
             ", "
           )}</strong></span>
           `
--- a/src/fontra/client/web-components/designspace-location.js
+++ b/src/fontra/client/web-components/designspace-location.js
@@ -111,7 +111,7 @@ export class DesignspaceLocation extends UnlitElement {
         elements.push(html.hr());
         continue;
       }
-      const values = axis.values ? axis.values : [];
+      const values = axis.values;
       const modelValue = this.values[axis.name];
       const infoBox = htmlToElement(
         `<div class="info-box">
--- a/src/fontra/client/web-components/designspace-location.js
+++ b/src/fontra/client/web-components/designspace-location.js
@@ -111,13 +111,24 @@ export class DesignspaceLocation extends UnlitElement {
         elements.push(html.hr());
         continue;
       }
+      const values = axis.values ? axis.values : [];
       const modelValue = this.values[axis.name];
       const infoBox = htmlToElement(
         `<div class="info-box">
-          <span>Min: <strong>${axis.minValue}</strong></span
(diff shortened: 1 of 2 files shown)


src/fontra/client/web-components/range-slider.js
```diff
@@ -242,7 +277,24 @@ export class RangeSlider extends html.UnlitElement {
...
+    if (isDiscrete) {
+      minValue = 0;
+      maxValue = this.values.length - 1;
+      step = 1;
+      value = this.valueFormatted;
+    } else {
+      step = this.step;
+      minValue = this.minValue;
+      maxValue = this.maxValue;
+      value = this.getClosestDiscreteValue(this.value);
```
justvanrossum, line 296: Hm, this is happening in the else branch of the `isDiscrete` if. Shouldn't line 291 and 296 be swapped?

src/fontra/client/web-components/range-slider.js
```diff
@@ -164,6 +165,17 @@ export class RangeSlider extends html.UnlitElement {
     this.sawMouseDown = false;
     this.sawMouseUp = false;
     this.onChangeCallback = () => {};
+    this.values = [];
+    this._sortedValuesCached = undefined;
+  }
+
+  get valuesSorted() {
```
justvanrossum, line 172: I'm wondering if the order even matters. For a discrete value slider, minValue and maxValue are meaningless and are ignored. For example, if the values were [50, 0, 100], I see no real reason why they could not be represented directly without sorting, as slider values 0, 1 and 2.

justvanrossum:
Can you add tickmarks? I'm curious what that will look like.

src/fontra/client/web-components/range-slider.js
fatih-erikli: Yes, it's wrong. I sent a fix.

src/fontra/client/web-components/range-slider.js
fatih-erikli: It's for finding the closest value. We start seeking from the lowest one, iterate until the current one is greater than actual value.

justvanrossum: Ok, I can see we need sorting for that. It would be nice if it would find the actual closest, for example for list [0, 100], 200 would find 0, and 60 would find 100.

justvanrossum:
Looking good so far. 

If you click on the label, a field with info gets shown. For a discrete axis it should not show min and max, but the possible values, in addition to the default value:
image

src/fontra/client/web-components/designspace-location.js
```diff
@@ -111,13 +111,24 @@ export class DesignspaceLocation extends UnlitElement {
...
       const infoBox = htmlToElement(
         `<div class="info-box">
-          <span>Min: <strong>${axis.minValue}</strong></span
-          >&nbsp; | <span>Default: <strong>${axis.defaultValue}</strong></span
-          >&nbsp; |
+          ${
+            values && values.length
+              ? `
+          <span>Default: <strong>${axis.defaultValue}</strong></span>&nbsp; |
+          <span>Possible values: <strong style="white-space: break-spaces;">${axis.values.join(
```
justvanrossum, line 122:
```suggestion
          <span>Values: <strong style="white-space: break-spaces;">${axis.values.join(
```

src/fontra/client/web-components/designspace-location.js
```diff
@@ -111,13 +111,24 @@ export class DesignspaceLocation extends UnlitElement {
         elements.push(html.hr());
         continue;
       }
+      const values = axis.values ? axis.values : [];
```
justvanrossum, line 114: Could it work like this, too?
```suggestion
      const values = axis.values;
```
If so, consider not useing an additional variable, but just use `axis.values` directly.

Thit would be my preference.

fatih-erikli: Lets keep it for now, we don't have `values` yet. An empty array is needed for creating range slider.

fatih-erikli: Creating range slider with values `undefined` doesn't work. It must be either an array or not to be existing in initial properties.

justvanrossum: Can you make it to work? I think `values = undefined` or `values = null` should be a valid thing for the range slider.

justvanrossum: Now that it works, please get rid of the `values` variable, and use `axis.values` throughout. (There's one place where you already do that accidentally.)

src/fontra/client/web-components/range-slider.js
```diff
@@ -274,61 +333,78 @@ export class RangeSlider extends html.UnlitElement {
...
+                this.sawMouseDown = false;
+                this.onChangeCallback(callbackEvent);
+              },
+            })),
+            isDiscrete &&
+              html.div(
+                {
+                  class: "tickmarks",
+                },
+                new Array(this.values.length).fill("").map(() =>
```
justvanrossum, line 397: I think this will work just as well, but is more compact:
```js
    this.values.map(() => html.span({ class: "tickmark" }))
```

src/fontra/client/web-components/designspace-location.js
fatih-erikli: I deleted' em

justvanrossum:
I found one more thing: the arrow keys should jump to the next/previous value. Shift-arrow-key should do the same, larger steps don't make sense here.

justvanrossum:
Btw. this is a useful test setting:
```js
axis.values = [axis.minValue, (axis.minValue + axis.maxValue) / 2, axis.maxValue];
```

fatih-erikli:
Test setting makes sense, ok, I will try. I will take care of arrow keys not jumping to the next/prev value.

PR #980, opened 2023-11-13
Rewrite add-remove-buttons with html.UnlitElement

### LINKED ISSUE #940: Rewrite add-remove-buttons to not use lit

fatih-erikli:
Fixes #940.

### DIFF
--- a/src/fontra/client/web-components/add-remove-buttons.js
+++ b/src/fontra/client/web-components/add-remove-buttons.js
@@ -1,5 +1,4 @@
-import { LitElement, css, html, unsafeCSS } from "../third-party/lit.js";
-import { InlineSVG } from "./inline-svg.js";
+import * as html from "../core/html-utils.js";
 import { themeColorCSS } from "./theme-support.js";
 
 const colors = {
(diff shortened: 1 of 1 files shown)


src/fontra/client/web-components/add-remove-buttons.js
```diff
@@ -1,5 +1,4 @@
-import { LitElement, css, html, unsafeCSS } from "../third-party/lit.js";
-import { InlineSVG } from "./inline-svg.js";
```
justvanrossum, line 2: I think we should leave the inline-svg.js import, because we use the web component here, even though we don't use the InlineSVG class directly. While it is imported elsewhere, too, we should not count on that.

Perhaps the import can be written to make it clear why we import it? Or add a comment. And/or do something like
```js
import { InlineSVG as _unused_InlineSVG } from "./inline-svg.js"; /* we need <inline-svg> */
```

Orrrr, you could just use `new InlineSVG("/path/to/image.svg")` instead of `html.createDomElement("inline-svg", ...)`

justvanrossum (commented):
One comment, the rest look good, thanks!

PR #981, opened 2023-11-13
Rotary-like control for angles

### LINKED ISSUE #919: Need a rotary control for angles
We need a simple, small-ish rotary control with which the user can more intuitively enter an angle. The use case initially is the rotation parameter of the component transformation.

fatih-erikli:
Fixes #919.

### DIFF
--- a/src/fontra/client/web-components/rotary-control.js
+++ b/src/fontra/client/web-components/rotary-control.js
@@ -1,0 +1,108 @@
+import * as html from "../core/html-utils.js";
+import { subVectors } from "../core/vector.js";
+
+export class RotaryControl extends html.UnlitElement {
+  static styles = `
+  .knob {
+    width: 5rem;
+    height: 5rem;
+    border-radius: 50%;
+    margin: 0.3rem;
+    background: #e3e3e3;
+    position: relative;
+  }
+
+  .thumb {
+    width: 0.2rem;
+    height: 100%;
+    left: 50%;
+    position: absolute;
+    margin-left: -0.1rem;
+    background: #d7d7d7;
+  }
+
+  .dot {
+    width: 0.2rem;
+    height: 0.2rem;
+    background: black;
+    border-radius: 50%;
+    margin-top: 5px;
+    background: #d92020;
+  }
+  `;
+
+  constructor() {
+    super();
+    this.onChangeCallback = () => {};
+  }
+
+  set value(value) {
+    this._value = value;
+    if (this.thumb) {
+      this.thumb.style.transform = `rotate(${this.value}deg)`;
+    }
+  }
+
+  get value() {
+    return this._value;
+  }
+
+  render() {
+    return (this.knob = html.div(
+      {
+        class: "knob",
+        onmousedown: (event) => {
+          this.startAngle = this.value;
+          const origin = originOfElement(this.knob);
+          const target = { x: event.clientX, y: event.clientY };
+          const sub = subVectors(target, origin);
+          const deg = toDegrees(Math.atan2(sub.y, sub.x));
+          this.angleWhenDragStart = deg;
+        },
+        onmouseup: (event) => {
+          this.startAngle = undefined;
+          this.angleWhenDragStart = undefined;
+        },
+        onmousemove: (event) => {
+          if (this.startAngle === undefined) {
+            return;
+          }
+          const origin = originOfElement(this.knob);
+          const target = { x: event.clientX, y: event.clientY };
+          const sub = subVectors(target, origin);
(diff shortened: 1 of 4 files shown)


src/fontra/client/web-components/rotary-control.js
```diff
@@ -0,0 +1,133 @@
...
+    );
+    this.overlay = overlay;
+    this.shadowRoot.appendChild(overlay);
+  }
+
+  render() {
+    return html.div({ class: "rotary-control" }, [
+      (this.knob = html.div(
+        {
+          onwheel: (event) => {
```
justvanrossum, line 109: Unless we can support dragBegin/dragEnd for wheel events we should drop onwheel.

src/fontra/client/web-components/ui-form.js
```diff
@@ -171,6 +172,37 @@ export class Form extends SimpleElement {
     valueElement.appendChild(inputElement);
   }
 
+  _addEditAngle(valueElement, fieldItem) {
+    const inputElement = document.createElement("input");
```
justvanrossum, line 176: We have an "input" shortcut in html-utils.js. I'm pretty sure you can write this instead:
```js
    const inputElement = html.input({type: "number", "value": ...etc...});
```

src/fontra/client/web-components/ui-form.js
```diff
@@ -171,6 +172,37 @@ export class Form extends SimpleElement {
...
+    inputElement.step = "any";
+
+    inputElement.disabled = fieldItem.disabled;
+    inputElement.onchange = (event) => {
+      let value = parseFloat(inputElement.value);
+      this._fieldChanging(fieldItem.key, value);
+      rotaryControl.value = value;
+    };
+
+    const rotaryControl = html.createDomElement("rotary-control", {
```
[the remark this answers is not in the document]
justvanrossum: Ok, fair enough!

justvanrossum:
Please rebase this.

justvanrossum (commented):
- When setting the angle numerically to (say) -50, the thumb get shown rotated counter-clockwise, but when dragging the thumb, the value changes in the opposite direction
- Dragging the thumb is often jumpy
- Dragging to the right rotates clockwise, and draging up rotates counter-clockwise. I think up should also rotate clockwise.

fatih-erikli:
> When setting the angle numerically to (say) -50, the thumb get shown rotated counter-clockwise, but when dragging the thumb, the value changes in the opposite direction

This is fixed. The rotation of glyph on the canvas is applied counterclockwise. There is conversion to the opposite side happening when dragging the thumb.

fatih-erikli:
Maybe I delete the conversion and keep it as it is? The rotation will be applied to the opposite side when dragging the thumb, is that ok? @justvanrossum.

justvanrossum:
> Maybe I delete the conversion and keep it as it is? The rotation will be applied to the opposite side when dragging the thumb, is that ok?

You mean rotate the rotary control clockwise while the component rotates counter-clockwise? No, they should match, that's the whole point.

Btw. please remove that demo panel, and just test with it with a selected component.

The movement is still very jumpy:

https://github.com/googlefonts/fontra/assets/4246121/2560e266-5b82-4ad2-bcf3-68df7f12358f

fatih-erikli:
Can you test the jumpy behavior with the latest commit?

fatih-erikli:
Hmm no this is a different kind of jumpy behavior. Video has just been loaded.

fatih-erikli:
Hmm, it looks like it's fixed with the latest commit.

justvanrossum:
It still jumps if you move across the NW - SE axis:

image

fatih-erikli:
This is tricky. I tried to fix that by caching the rotation axis, but it expects the drag to be same with when the movement started until the mouse up.

src/fontra/views/editor/editor.js
```diff
@@ -62,6 +62,7 @@ import { themeController } from "/core/theme-settings.js";
 import { MenuItemDivider, showMenu } from "/web-components/menu-panel.js";
 import { dialog, dialogSetup } from "/web-components/modal-dialog.js";
 import { parsePluginBasePath } from "/web-components/plugin-manager.js";
+import { RotaryControl } from "/web-components/rotary-control.js";
```
justvanrossum, line 65: I think this import belongs in ui-form, where it's used via createDomElement.

justvanrossum:
Can you make the rotary control as big as possible without affecting the item spacing? It is currently smaller than the numeric input. Can it be the same size?

fatih-erikli:
Maximum height without affecting the item spacing is the current one. There was a margin, I deleted it, they should seem in same height now.

PR #985, opened 2023-11-13
Delete lit dependency

### LINKED ISSUE #984: Remove Lit from package.json, etc.

fatih-erikli:
Fixes #984.


justvanrossum (commented):
One comment, otherwise looks good to me.

But: please run `pip install -e .`. This implicitly runs `npm install`, and I expect `package-lock.json` to be changed.

Then also do a quick test of the app, after server restart.

PR #989, opened 2023-11-15
Focus only when focus attribute is true when text entry panel is toggled

### LINKED ISSUE #988: [left sidebar] text entry field has focus on page reload
If the text entry tab is active, and you reload the page, the text entry field has the focus. The canvas should have the focus.

fatih-erikli:
Fixes #988.


justvanrossum (approved):
Perfect, thanks!

PR #990, opened 2023-11-15
Delete the third party javascript folder when pip install -e

### LINKED ISSUE #987: rollup script should empty third-party directory first
Currently, the rollup script (which runs as part of `pip install .`) onlly adds and overwrites to the `src/fontra/client/third-party/`. But anything that was there before that it doesn't write stays. This makes it a little harder to test the removal of things (such as recently with lit). The rollup script should empty (or delete) the third-party folder before it writes the new contents.

fatih-erikli:
Fixes #987.

### DIFF
--- a/scripts/rollup_build_hook.py
+++ b/scripts/rollup_build_hook.py
@@ -15,6 +15,6 @@ class RollupBuildHook(BuildHookInterface):
             / "third-party"
         )
         if path.exists():
-            shutil.rmtree(path.resolve())
+            shutil.rmtree(path)
         subprocess.check_output("npm install", shell=True)
         subprocess.check_output("npm run bundle-rollup", shell=True)
--- a/scripts/rollup_build_hook.py
+++ b/scripts/rollup_build_hook.py
@@ -1,3 +1,5 @@
+import pathlib
+import shutil
 import subprocess
 
 from hatchling.builders.hooks.plugin.interface import BuildHookInterface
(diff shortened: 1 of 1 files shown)


justvanrossum:
Sorry, this is the wrong approach (too complicated).

Please do `shutil.rmtree()` in https://github.com/googlefonts/fontra/blob/main/scripts/rollup_build_hook.py, using `pathlib.Path(__file__).resolve().parent` etc. to find the right directory.

scripts/rollup_build_hook.py
```diff
@@ -1,9 +1,21 @@
...
+import shutil
 import subprocess
 
 from hatchling.builders.hooks.plugin.interface import BuildHookInterface
 
 
 class RollupBuildHook(BuildHookInterface):
     def initialize(self, version, build_data):
+        path = (
+            pathlib.Path(__file__).parent
```
justvanrossum, line 11: Adding ".." is the same as adding another `.parent`. Please do `.parent` instead.
Also, the resolve should come first, not last:
`pathlib.Path(__file__).resolve().parent.parent ...`

justvanrossum (commented):
This looks mostly good now.

Btw I closed the PR to indicate that I wanted you to start a new one. We don't need all those earlier commits, so I will have to do a squash commit once it's done.

fatih-erikli:
I deleted earlier commits, no need to squash now.

scripts/rollup_build_hook.py
```diff
@@ -1,9 +1,20 @@
...
     def initialize(self, version, build_data):
+        path = (
+            pathlib.Path(__file__).resolve().parent.parent
+            / "src"
+            / "fontra"
+            / "client"
+            / "third-party"
+        )
+        if path.exists():
+            shutil.rmtree(path.resolve())
```
justvanrossum, line 18: This .resolve() is now redundant.

fatih-erikli: ok

PR #997, opened 2023-11-23
Display current character in reference fonts

### LINKED ISSUE #982: [Reference Font] Add area to the panel showing the current character in all reference fonts
- In the Reference Font panel, add an area below the "Language" field, that can be as tall as the leftover space.
- In the area, show the currently edited glyph/character, for *each* of the reference fonts.
- At the bottom there should be a slider for the font size (min 12, max 300, default 50, as a starting point).
- Perhaps there should be a space between the characters, perhaps not.
- Each character should have its own `<span>`
- Wrapping should be allowed per character
- If the font size is too big for all chars to fit in the area, the area should scroll vertically.
- The "Custom character" should probably override the current glyph here, too.

The fonts should be easily available to CSS via their font id.

fatih-erikli:
Fixes #982.

### DIFF
--- a/src/fontra/views/editor/panel-reference-font.js
+++ b/src/fontra/views/editor/panel-reference-font.js
@@ -319,7 +319,7 @@ export default class ReferenceFontPanel extends Panel {
       this.editorController.sceneSettings.glyphLines
     );
 
-    const letter = this.model.charOverride || selectedGlyphInfo?.glyphName;
+    const letter = this.model.charOverride || selectedGlyphInfo?.character;
 
     if (!letter) {
       return;
--- a/src/fontra/views/editor/panel-reference-font.js
+++ b/src/fontra/views/editor/panel-reference-font.js
@@ -350,23 +350,6 @@ export default class ReferenceFontPanel extends Panel {
       return;
     }
 
-    if (
-      !textToDisplay &&
-      selectedGlyphInfo &&
-      selectedGlyphInfo.glyphName.includes(".")
-    ) {
-      const baseGlyphName = positionedGlyph.glyphName.split(".")[0];
-      const codePoint = (this.editorController.fontController.glyphMap[baseGlyphName] ||
-        [])[0];
-      if (codePoint) {
-        textToDisplay = String.fromCodePoint(codePoint);
-      }
-    }
-
-    if (!textToDisplay) {
-      return;
-    }
-
     const currentCharacter = div(
       {
         class: "current-character",
--- a/src/fontra/views/editor/panel-reference-font.js
+++ b/src/fontra/views/editor/panel-reference-font.js
@@ -369,7 +369,10 @@ export default class ReferenceFontPanel extends Panel {
       return;
     }
 
-    const currentCharacter = div({ class: "current-character" }, []);
+    const currentCharacter = div(
+      { class: "current-character", lang: this.model.languageCode },
+      []
+    );
     const rangeSlider = createDomElement("range-slider", {
       type: "range",
       value: 14,
--- a/src/fontra/views/editor/panel-reference-font.js
+++ b/src/fontra/views/editor/panel-reference-font.js
@@ -300,7 +300,7 @@ export default class ReferenceFontPanel extends Panel {
     referenceFontModel = this.model;
 
     this.editorController.sceneSettingsController.addKeyListener(
-      ["selectedGlyphName"],
+      "selectedGlyphName",
       (event) => {
         this.requestReferenceFontsPreview();
       }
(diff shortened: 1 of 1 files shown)


justvanrossum:
Can you rebase this please?

justvanrossum:
- the font size should be stored
- if I change the language code with the popup, it is not applied to the text

justvanrossum:
Occasionaly, on reload, the text area duplicates:
image

Can you figure how this can be happening?

Second, if the text doesn't fit anymore, it starts overlapping. I think the text should scroll instead:

image

fatih-erikli:
It's because multiple events fired to show the preview, resulting multiple times the preview is shown. It was fixed, but I deleted the fix. If an event fired before the previous event is completed, it should be dismissed. I will include the fix again.

https://github.com/googlefonts/fontra/pull/997/commits/07d3365d796059e9111e3f892e9e30c5412ea586#diff-89f2511fb56caa04c583a1029f729c9022d8a0ec62cebef73c0c99a90e936f2bL308-L318

Something like that should be in the utils I guess, what do you think?

fatih-erikli:
For the overlapping problem, I don't have a quick solution for that, beside applying a constant max-height. I need to dig into css grid a little bit to find a proper solution.

justvanrossum:
The text area should scroll _without_ the size slider.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -1,10 +1,24 @@
...
+  input,
+  label,
+  option,
+  select,
+  span,
+} from "/core/html-utils.js";
 import { ObservableController } from "/core/observable-object.js";
-import { fetchJSON, fileNameExtension, withTimeout } from "/core/utils.js";
+import {
+  consolidateCalls,
```
justvanrossum, line 15: `consolidateCalls` doesn't seem to be used.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -264,15 +292,114 @@ export default class ReferenceFontPanel extends Panel {
...
+        style: `font-size: ${this.model.fontSize}px`,
+        lang: this.model.languageCode,
+      },
+      []
+    );
+    const rangeSlider = createDomElement("range-slider", {
+      type: "range",
+      value: this.model.fontSize,
+      minValue: 10,
+      maxValue: 200,
```
justvanrossum, line 387: Please make the max 300, make the default 100 (also elswhere in the code, maybe it should be a constant if it is currently hardcoded in several places), make sure the slider has a matching `defaultValue` property (it currently goes to 0 when you alt-click the slider).

justvanrossum: Elswhere, the default fo the font size is set to 14. Please make this 100, and use a constant for it.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -264,15 +287,115 @@ export default class ReferenceFontPanel extends Panel {
...
       this.editorController.canvasController.setLangAttribute(this.model.languageCode);
+      this.requestReferenceFontsPreview();
     });
 
     this.editorController.canvasController.setLangAttribute(this.model.languageCode);
 
     referenceFontModel = this.model;
+
+    this.editorController.sceneSettingsController.addKeyListener(
+      ["selectedGlyph", "glyphLines"],
```
justvanrossum, line 303: I think it's enough to listen to "selectedGlyphName". Can you try that?

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -264,15 +287,115 @@ export default class ReferenceFontPanel extends Panel {
...
+      return this.referenceFontsPreviewPromise;
+    }
+    this.referenceFontsPreviewPromise = this.displayCurrentGlyphInReferenceFonts();
+    await this.referenceFontsPreviewPromise;
+    delete this.referenceFontsPreviewPromise;
+  }
+
+  async displayCurrentGlyphInReferenceFonts() {
+    const container = this.contentElement.querySelector(".reference-font-preview");
+    while (container.firstChild) {
```
justvanrossum, line 321: This look could probably be replaced with 
```js
    container.innerHTML = ""
```
(This is a pattern I use elsewhere, too)

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -264,15 +287,115 @@ export default class ReferenceFontPanel extends Panel {
...
+    if (!textToDisplay) {
+      return;
+    }
+
+    if (
+      !textToDisplay &&
+      selectedGlyphInfo &&
+      selectedGlyphInfo.glyphName.includes(".")
+    ) {
+      const baseGlyphName = positionedGlyph.glyphName.split(".")[0];
```
justvanrossum, line 358: This sections is almost identical to the above (at line 336). Why is this needed? It seems redundant.

fatih-erikli: Yes, I forgot this section to delete when copying the behavior to the section above.

src/fontra/views/editor/panel-reference-font.js
fatih-erikli: Yes, it works

PR #1026, opened 2023-12-14
Discrete value range slider should be shorter

### LINKED ISSUE #1002: [designspace panel] Discrete value range slider should be shorter
Currently, the range slider is as wide as there's space, but for a discrete value slider I think it should be shorter, depending on the number of values.

Current situation:

image

Desired layout:

image

(Sketch is not pixel-perfect)

When the number of values becomes to big that the slider would become to wide, it should be made to match the current behavior (fill the available space, bit not more).

fatih-erikli:
Fixes #1002

### DIFF
--- a/src/fontra/client/web-components/range-slider.js
+++ b/src/fontra/client/web-components/range-slider.js
@@ -358,7 +358,12 @@ export class RangeSlider extends html.UnlitElement {
           ]),
         ]),
         html.div(
-          { class: "range-container" },
+          {
+            class: "range-container",
+            style: this.isDiscrete()
+              ? `max-width: ${this.values.length * 40}px; width: 100%;`
+              : "",
+          },
           [
             (this.rangeInput = html.input({
               type: "range",
--- a/src/fontra/client/web-components/range-slider.js
+++ b/src/fontra/client/web-components/range-slider.js
@@ -361,7 +361,13 @@ export class RangeSlider extends html.UnlitElement {
           {
             class: "range-container",
             style: this.isDiscrete()
-              ? `max-width: ${this.values.length * 20}px; width: 100%;`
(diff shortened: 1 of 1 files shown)


justvanrossum:
The item spacing is too wide and not consistent:

image

Please make it match the sketch in #1002

fatih-erikli:
This is a tricky css problem, I will come back to this later.

fatih-erikli:
If you have a suggestion, I'd be happy to hear.

justvanrossum:
The element could have its own css that depends on the number of values.

fatih-erikli:
Can you elaborate? @justvanrossum

justvanrossum:
Something like:
```js
    el = html.div({style: `width: ${some_computed_value_based_on_num_values}em;`);
```
or:
```js
    el.style = `width: ${some_computed_value_based_on_num_values}em;`;
```

fatih-erikli:
`max-width: ${this.values.length * 20}px; width: 100%;`

This is how the total width calculated currently.

They look ok, however not pixel perfect.

Screenshot 2023-12-17 at 00 42 58

fatih-erikli:
Not only the tickmarks, the builtin range slider has that gap too.

Screenshot 2023-12-17 at 00 48 08

Maybe thumb-width should be added to the calculation. I tried, it didn't help.

justvanrossum:
I don't think you are accounting for the track length bore the first and after the last tickmark. I think the computation should be something like:
```js
this.values.length * 20 + margin * 2
```
Where `margin` is the space before the first chckmark.

fatih-erikli:
Space before the checkmark is half of thumb width. I tried that, it doesn't help.

fatih-erikli:
Assigning it to @justvanrossum for the remaining gap issue.

PR #1036, opened 2023-12-21
Delete sidebar init methods in editor.js

### LINKED ISSUE #748: [sidebars] further cleanup and refactoring is needed
- [x] Panels need to be notified when they are shown or hidden, so they can respond to that (Done in #757)
- [ ] There is still some init/setup code of individual panels left in editor.js, that should be moved to the respective panel classes (mostly done in #757, the rest is blocked by a missing feature: #766)
- [x] Rethink the `attach()` method (#934)
- [x] The various panels should ideally more consistent in how their construction/setup is coded (#934)

All this is needed for better future extensibility and use by plugins.

fatih-erikli:
Fixes #748.

### DIFF
--- a/src/fontra/views/editor/panel-designspace-navigation.js
+++ b/src/fontra/views/editor/panel-designspace-navigation.js
@@ -105,213 +105,7 @@ export default class DesignspaceNavigationPanel extends Panel {
     );
 
     this.fontController.ensureInitialized.then(() => {
-      this.designspaceLocation = this.contentElement.querySelector(
-        "#designspace-location"
-      );
-      this.designspaceLocation.values = this.sceneSettings.location;
-
-      this.designspaceLocation.addEventListener(
-        "locationChanged",
-        scheduleCalls(async (event) => {
-          this.sceneController.scrollAdjustBehavior = "pin-glyph-center";
-          this.sceneController.autoViewBox = false;
-
-          this.sceneSettingsController.setItem(
-            "location",
-            { ...this.designspaceLocation.values },
-            { senderID: this }
-          );
-        })
-      );
-
-      this.sceneSettingsController.addKeyListener("selectedGlyphName", (event) => {
-        this._updateAxes();
-        this._updateSources();
-        this._updateEditLocalAxesButtonState();
-        this._updateInterpolationErrorInfo();
-      });
-
-      this.sceneController.addCurrentGlyphChangeListener(
-        scheduleCalls((event) => {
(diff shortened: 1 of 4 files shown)


fatih-erikli:
https://github.com/googlefonts/fontra/pull/1036/commits/f15ac2d8dbe0d413276c7982f65e81607a6edc69. I deleted setup method. Setup sounds like a method name derives from the base panel class, it may confuse the reader, this is why I deleted.

src/fontra/views/editor/panel-designspace-navigation.js
```diff
@@ -103,6 +103,216 @@ export default class DesignspaceNavigationPanel extends Panel {
       () => this._updateResetAllAxesButtonState(),
       100
     );
+
+    this.fontController.ensureInitialized.then(() => {
```
justvanrossum, line 107: This change is too big. Also, the body of this arrow function is too big.

Can you try to do this:
- revert the `setup()` method change
- _call_ `.setup()` from a "then" handler

Since setup is big, I would like to be sure nothing important changes there. By inlining the whole thing in an arrow function I have no way to compare before and after.

src/fontra/views/editor/panel-glyph-search.js
```diff
@@ -37,4 +83,13 @@ export default class GlyphSearchPanel extends Panel {
   }
 }
 
+export function glyphInfoFromGlyphName(glyphName, fontController) {
```
justvanrossum, line 86: Let's make this function a method of FontController instead. I don't like how editor needs to import this from a panel module.

src/fontra/views/editor/panel-designspace-navigation.js
justvanrossum: If the name "setup" is problematic (as you seem to say in https://github.com/googlefonts/fontra/pull/1036#issuecomment-1866883562), then please suggest another name, instead of deleting the entire method.

fatih-erikli: I reverted the commit, kept it as `setup`, I suggest to break the method into several pieces as a next step. As you say, this method is too big. Only updateAxes, updateSources calls at the end requires fontController to be initialized. We can move the event bindings to the constructor. Create another method named `createUIList`. I am not sure if we should do that in this PR.

justvanrossum (commented):
`glyphNameChangedCallback()` should be deleted from editor.js.

PR #1039, opened 2023-12-25
Improve tooltip positioning

### LINKED ISSUE #798: Improve tooltip positioning
The tooltip code currently works well for the "Reset all axes" and "Edit local axes" buttons, but that's only because those labels have similar lengths.

The following improvements should be made:
- the tooltip should be centered
- except when it would be clipped by the window, then it can move

fatih-erikli:
Fixes #798.


justvanrossum:
I noticed that the margin between the edge and the arrow is different between top-left and top-right:

image
image

(I hacked the UI to make the local axes button top-right)

justvanrossum:
(Same for `bottom-right` vs `bottom-left`)

justvanrossum:
Also, please rebase.

PR #1040, opened 2023-12-27
Dropped variable fonts should react to global axis sliders

### LINKED ISSUE #1023: [reference font] Dropped VFs should respond to global axis sliders

fatih-erikli:
Fixes #1023.

### DIFF
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -139,9 +139,13 @@ export class EditorController {
     );
 
     this.sceneSettingsController.addKeyListener("location", () => {
-      const fontVariationSettingsCssString = this.fontController.globalAxes
-        .map((axis) => `'${axis.tag}' ${this.sceneSettings.location[axis.tag]}`)
-        .join(",");
+      const fontVaritionSettings = [];
+      for (const axis of this.fontController.globalAxes) {
+        fontVaritionSettings.push(
+          `'${axis.tag}' ${this.sceneSettings.location[axis.tag]}`
+        );
+      }
+      const fontVariationSettingsCssString = fontVaritionSettings.join(",");
       this.canvasController.canvas.style.fontVariationSettings =
         fontVariationSettingsCssString;
       this.canvasController.requestUpdate();
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -138,22 +138,6 @@ export class EditorController {
       }
     );
 
-    this.sceneSettingsController.addKeyListener("location", () => {
-      const fontVariationSettings = [];
-      for (const axis of this.fontController.globalAxes) {
-        fontVariationSettings.push(
-          `'${axis.tag}' ${this.sceneSettings.location[axis.tag]}`
-        );
-      }
-      const fontVariationSettingsCssString = fontVariationSettings.join(",");
-      this.canvasController.canvas.style.fontVariationSettings =
-        fontVariationSettingsCssString;
-      this.canvasController.requestUpdate();
-      const referenceFontsPanel = this.getSidebarPanel("reference-font");
-      referenceFontsPanel.contentElement.style.fontVariationSettings =
-        fontVariationSettingsCssString;
-    });
-
     this.cjkDesignFrame = new CJKDesignFrame(this);
 
     this.visualizationLayers = new VisualizationLayers(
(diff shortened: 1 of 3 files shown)


src/fontra/views/editor/editor.js
```diff
@@ -138,6 +138,22 @@ export class EditorController {
       }
     );
 
+    this.sceneSettingsController.addKeyListener("location", () => {
```
justvanrossum, line 141: This should be in panel-reference-font.js. The functionality should perhaps be a separate method, so that this can also be called when the reference font panel is initialized, not just when the sliders change.

justvanrossum: Additionally, this setting should also affect the reference glyph drawn in the canvas (see the visualization layer, marked with registerVisualizationLayerDefinition).

fatih-erikli: It should be done already, in line 149.

fatih-erikli: Same canvas is used for drawing the reference text. The only way to set variable glyph is doing by setting by css to the canvas element currently. There is no method for setting font variables in CanvasContext2D.

justvanrossum: Although I don't see any unexpected results in other text drawing, and indeed the cavas API doesn't seem to support `fontVariationSettings` yet. So let's go forwards with this.

justvanrossum: (Sorry, our messages crossed, I confirm what you wrote.)

justvanrossum: Then all that's left to do is move the code to the right module (first comment).

fatih-erikli: Do you want me to separate the event callbacks into two, moving the canvas font setting to the `canvas-controller` and reference font related code to the `panel-reference-font`?

fatih-erikli: I did.

fatih-erikli:
CSS font variable setting on canvas doesn't seem to work on Firefox. Is that ok?

src/fontra/views/editor/scene-controller.js
```diff
@@ -121,6 +121,7 @@ export class SceneController {
 
     // Set up the mutual dependencies between location and selectedSourceIndex
     this.sceneSettingsController.addKeyListener("location", async (event) => {
+      this.setCanvasFontVariables();
```
justvanrossum, line 124: None of this belongs here, it should all be in panel-reference-font.js, like I wrote before.

fatih-erikli: I moved all of them to reference fonts panel.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -305,6 +305,20 @@ export default class ReferenceFontPanel extends Panel {
         this.requestReferenceFontsPreview();
       }
     );
+
+    editorController.sceneSettingsController.addKeyListener("location", () => {
+      const fontVariationSettings = [];
+      for (const axis of this.editorController.fontController.globalAxes) {
+        fontVariationSettings.push(
+          `'${axis.tag}' ${this.editorController.sceneSettings.location[axis.tag]}`
```
justvanrossum, line 313: Did you test this? Because it doesn't work unless you apply this:
```suggestion
          `'${axis.tag}' ${this.editorController.sceneSettings.location[axis.name]}`
```

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -305,6 +305,20 @@ export default class ReferenceFontPanel extends Panel {
...
+        );
+      }
+      const buildFontVariablesCssString = fontVariationSettings.join(",");
```
justvanrossum, line 316: Suggest to use `cssString` as the variable name. The current name makes no sense.

src/fontra/views/editor/panel-reference-font.js
justvanrossum: Ah, it does work when looking at MutatorSans.ttf (where axis.name == axis.tag) but not with MutatorSans.designspace (where axis.name != axis.tag).

src/fontra/views/editor/panel-reference-font.js
fatih-erikli: Ok, it's a renaming mistake.

justvanrossum:
Too bad it doesn't work in Firefox. I don't see a workaround, though.

PR #1047, opened 2024-01-02
Sub-menu implementation in menu-panel

### LINKED ISSUE #1035: [menu-panel] add submenu mechanism
The submenu should be built just before it is shown, so it can be maximally dynamic. So there should probably be a function that will build its contents.

fatih-erikli:
Fixes #1035.


fatih-erikli:
Functionality must be done however the corners visually not looking well yet. @justvanrossum. I will work on it once we make sure it works correct.

justvanrossum:
It works well, and looks quite ok. I would prefer the submenu-icon to look more like in macOS:
image

I don't immediately see a suitable unicode for it, although these come close:
image

But I think the "chevron-right" icon from https://tabler.io/icons is best: at least we wouldn't be depending on a system font's appearance.

image

My [comment](https://github.com/googlefonts/fontra/pull/929#issuecomment-1907680508) about a more dynamic items list (populated on-demand) applies here as well.

justvanrossum:
Some more feedback regarding appearance and behavior.

image

- The red selected items should align: the submenu should be a bit higher

- If I look at macOS, the "parent" items gets a different (gray) highlight when you show a sub-sub-menu:

image



When you go back to the parent-parent, I think the sub-sub-menu should hide again:

image

fatih-erikli:
This is not ready yet, I will rethink the storing open menu instances, currently I read them in the DOM layout. The main PR is ready for review.

PR #1058, opened 2024-01-17
Add message() to modal-dialog

### LINKED ISSUE #1057: Add message() function to modal-dialog.js
- [ ] Create a new function `message()`: It should take two arguments, `headline`, and `message`, and will call `dialog()`, with a single "Okay" button.
- [ ] Find and replace all `dialog()` calls that currently explicitly add a single "Okay" button.

fatih-erikli:
Fixes #1057.

### DIFF
--- a/src/fontra/client/web-components/modal-dialog.js
+++ b/src/fontra/client/web-components/modal-dialog.js
@@ -23,6 +23,10 @@ export async function dialogSetup(headline, message, buttonDefs, autoDismissTime
   return dialogContentElement;
 }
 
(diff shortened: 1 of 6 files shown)


justvanrossum (commented):
Thanks, looks good.

I believe there are a few more instances of dialog() that can be replaced:
- scene-controller.js: "The glyph could not be saved"
- panel-reference-font.js: "Could not store some reference fonts"
- panel-designspace-navigation.js: "The source has an interpolation incompatibility"
- editor.js: in messageFromServer()

PR #1064, opened 2024-01-23
Implementation of X/Y component transformation fields

### LINKED ISSUE #1051: X/Y component transformation fields should be shown on a single line
Current situation:

image

For example, the two rows `translateX` and `translateY` should be merged into a single row, with label `translate`, and two number input fields. Same for `scale`, `skew` and `tCenter`.

fatih-erikli:
Fixes #1051.


justvanrossum (commented):
It looks great, but editing the numeric fields doesn't seem to work.

justvanrossum:
> It looks great, but editing the numeric fields doesn't seem to work.

I see the problem (the individual x y field are not found by key), I will figure out a fix.

PR #1066, opened 2024-01-23
Rotary control should be shown in counterclockwise

### LINKED ISSUE #1065: [selection info] Transform rotation rotary control initially shows wrong direction
A positive angle is a counter clockwise rotation, but the rotary dial shows a clockwise rotation:

image

Once you move the rotary dial, things are ok, but since the initial direction is wrong, you start with a jump.

fatih-erikli:
Fixes #1065.

### DIFF
--- a/src/fontra/client/web-components/ui-form.js
+++ b/src/fontra/client/web-components/ui-form.js
@@ -196,18 +196,18 @@ export class Form extends SimpleElement {
       onchange: () => {
         let value = parseFloat(inputElement.value);
         this._fieldChanging(fieldItem.key, value);
-        rotaryControl.value = value * -1;
+        rotaryControl.value = value;
       },
     });
     const rotaryControl = html.createDomElement("rotary-control", {
-      value: fieldItem.value,
+      value: -1 * fieldItem.value,
     });
     {
       // Rotary change closure
       let valueStream;
 
       rotaryControl.onChangeCallback = (event) => {
-        const value = -event.value;
+        const value = -1 * event.value;
         inputElement.value = value;
         if (event.dragBegin) {
           valueStream = new QueueIterator(5, true);


src/fontra/client/web-components/ui-form.js
```diff
@@ -196,18 +196,18 @@ export class Form extends SimpleElement {
       onchange: () => {
         let value = parseFloat(inputElement.value);
         this._fieldChanging(fieldItem.key, value);
-        rotaryControl.value = value * -1;
+        rotaryControl.value = -1 * value;
```
justvanrossum, line 199:
```suggestion
        rotaryControl.value = -value;
```

src/fontra/client/web-components/ui-form.js
```diff
@@ -196,18 +196,18 @@ export class Form extends SimpleElement {
...
       },
     });
     const rotaryControl = html.createDomElement("rotary-control", {
-      value: fieldItem.value,
+      value: -1 * fieldItem.value,
```
justvanrossum, line 203:
```suggestion
      value: -fieldItem.value,
```

src/fontra/client/web-components/ui-form.js
```diff
@@ -196,18 +196,18 @@ export class Form extends SimpleElement {
...
     });
     {
       // Rotary change closure
       let valueStream;
 
       rotaryControl.onChangeCallback = (event) => {
-        const value = -event.value;
+        const value = -1 * event.value;
```
justvanrossum, line 210:
```suggestion
        const value = -event.value;
```

justvanrossum (commented):
LGTM, except I'd syntactically prefer to just use `-` instead of multiplying by `-1`. This should not be a functional change.

PR #1100, opened 2024-02-19
Issue 1093 olli

### LINKED ISSUE #1093: Created glyph uses different case than the entered glyph name
I wanted to create a new glyph called /h.ss01. I double clicked on the new glyph and fontra created a new glyph called /H.ss01. So, it changed it to uppercase. 
Screenshot 2024-02-16 at 09 57 29
Screenshot 2024-02-16 at 09 58 01

ollimeier:
This fixes #1093.

Screenshot 2024-02-19 at 16 42 42

### DIFF
--- a/src/fontra/client/core/glyph-lines.js
+++ b/src/fontra/client/core/glyph-lines.js
@@ -50,11 +50,23 @@ async function glyphNamesFromText(text, characterMap, glyphMap) {
           // text entry field.
           const [baseGlyphName, extension] = splitGlyphNameExtension(glyphName);
           const baseCharCode = baseGlyphName.codePointAt(0);
+          console.log("baseGlyphName", baseGlyphName);
+          console.log("baseCharCode", baseCharCode);
           const charString = String.fromCodePoint(baseCharCode);
+          console.log("charString", charString);
+          /*
+          if (baseGlyphName && baseGlyphName.normalize("NFD").length === baseGlyphName.length) {
+            glyphName = glyphName;
+            char = charString;
+            console.log("glyphName", glyphName);
+            console.log("char", char);
+          } else
+          */
           if (baseGlyphName === charString) {
             // The base glyph name is a single character, let's see if there's
             // a glyph name associated with that character
             let properBaseGlyphName = characterMap[baseCharCode];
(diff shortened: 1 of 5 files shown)


src/fontra/client/core/glyph-lines.js
```diff
@@ -54,14 +54,23 @@ async function glyphNamesFromText(text, characterMap, glyphMap) {
...
             // a glyph name associated with that character
-            let properBaseGlyphName = characterMap[baseCharCode];
-            if (!properBaseGlyphName) {
-              properBaseGlyphName = await getSuggestedGlyphName(baseCharCode);
-            }
-            if (properBaseGlyphName) {
-              glyphName = properBaseGlyphName + extension;
-              if (!extension) {
-                char = charString;
+            if (baseGlyphName.search(/[^a-zA-Z]+/) === -1) {
```
justvanrossum, line 57: I don't think this is correct. You are searching for any non-A-Z and non-a-z chars, but that is not the right criterium.

I think the preceding `if (baseGlyphName === charString)` condition should be changed to something like:

```js
    if (baseGlyphName === charString && !isPlainLatinLetter(baseGlyphName) ) {
```

Where `isPlainLatinLetter(glyphName)` checks whether the name is exactly one character, matching `[A-Za-z]`. For example (untested):

```js
function isPlainLatinLetter(glyphName) {
  return glyphName.match(/^[A-Za-z]$/)
}
```

Then the rest of the code can remain unchanged.

justvanrossum (commented):
The glyphnames.py change is good, I left a suggestion how to approach .notdef.

src/fontra/client/core/glyph-lines.js
ollimeier: Where should we store the new function `isPlainLatinLetter`?

ollimeier: Within utils?

justvanrossum: Or at the end of the module.

ollimeier: Please review: https://github.com/googlefonts/fontra/pull/1100/commits/28feb5d3fb22425aaec603ea3a398236b7500e42

Issue #756, opened 2023-08-25
[glyph search] Glyph name column should get wider when the sidebar gets wider


justvanrossum:
UIList is unfortunately a bit dumb regarding column widths: they are currently rather inflexible.

UIList should perhaps use CSS Grid, but then we won't have an element for the row as a whole.

Each _cell_ currently has its own width, this should at least be a reference to the column width.

Two options (that are not mutually exclusive):
- Assign one column to be flexible in width, so it can respond to container resizing
- Make columns optionally resizable via the column header

The latter would be good to have in any case.

PR #1101, opened 2024-02-19
Show multiple code points in glyph search 

### LINKED ISSUE #1096: Glyph search list does not show multiple code points
glyph search area does not display all unicodes, only one (even though there are multiple encodings)

Screenshot 2024-02-16 at 09 59 14

ollimeier:
fixes #1096

Screenshot 2024-02-19 at 18 09 14

### DIFF
--- a/src/fontra/client/web-components/glyphs-search.js
+++ b/src/fontra/client/web-components/glyphs-search.js
@@ -6,7 +6,6 @@ import {
   getCharFromUnicode,
   guessCharFromGlyphName,
   makeUPlusStringFromCodePoint,
-  makeUPlusStringFromCodePoints,
 } from "/core/utils.js";
 
 const colors = {
@@ -70,7 +69,7 @@ export class GlyphsSearch extends UnlitElement {
       {
         key: "unicode",
         width: "fit-content",
-        get: (item) => makeUPlusStringFromCodePoints(item.unicodes),
+        get: (item) => item.unicodes?.map(makeUPlusStringFromCodePoint).join(","),
       },
     ];
     this.glyphNamesList = new UIList();
--- a/src/fontra/client/web-components/glyphs-search.js
+++ b/src/fontra/client/web-components/glyphs-search.js
@@ -69,7 +69,7 @@ export class GlyphsSearch extends UnlitElement {
       { key: "glyphName", title: "glyph name", width: "10em", isIdentifierKey: true },
       {
         key: "unicode",
-        width: "5em",
+        width: "fit-content",
         get: (item) => makeUPlusStringFromCodePoints(item.unicodes),
       },
     ];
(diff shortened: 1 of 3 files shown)


src/fontra/client/web-components/glyphs-search.js
```diff
@@ -68,8 +69,8 @@ export class GlyphsSearch extends UnlitElement {
       { key: "glyphName", title: "glyph name", width: "10em", isIdentifierKey: true },
       {
         key: "unicode",
-        width: "5em",
-        get: (item) => makeUPlusStringFromCodePoint(item.unicodes[0]),
+        width: "fit-content",
+        get: (item) => makeUPlusStringFromCodePoints(item.unicodes),
```
justvanrossum, line 73: I would just inline the whole thing:
```suggestion
        get: (item) => item.unicodes?.map(makeUPlusStringFromCodePoints).join(","),
```
Your `makeUPlusStringFromCodePoints()` is too flexible in what it accepts and therefore more complicated than it should be.

src/fontra/client/web-components/glyphs-search.js
```diff
@@ -68,8 +69,8 @@ export class GlyphsSearch extends UnlitElement {
       { key: "glyphName", title: "glyph name", width: "10em", isIdentifierKey: true },
       {
         key: "unicode",
-        width: "5em",
-        get: (item) => makeUPlusStringFromCodePoint(item.unicodes[0]),
+        width: "fit-content",
```
justvanrossum, line 72: This is perhaps fine for now, but relates to #756. This is currently only working correctly because it is the last column...

src/fontra/client/web-components/glyphs-search.js
ollimeier: Thanks for the great feedback. Please see the new commits

ollimeier:
@justvanrossum This time I wanted to go the ideal workflow. First, I created a failing unittest and then – I had the feeling I need another function for that, why I created one. But I am with you, I made things unnecessary complicated. Thanks for your great feedback.

src/fontra/client/web-components/glyphs-search.js
```diff
@@ -68,8 +68,8 @@ export class GlyphsSearch extends UnlitElement {
...
+        get: (item) => item.unicodes?.map(makeUPlusStringFromCodePoint).join(","),
```
justvanrossum, line 72: Let's remove the `?`: I just checked, and `.unicodes` comes straight from the `glyphMap` object, where values _must_ be arrays (but can be empty).
```suggestion
        get: (item) => item.unicodes.map(makeUPlusStringFromCodePoint).join(","),
```

justvanrossum (commented):
Last small suggestion.

PR #1108, opened 2024-02-21
Implement edit field for sidebearings (left/right margin) in sel info panel

### LINKED ISSUE #440: Add info field for left/right sidebearing values
Relates to #113.

justvanrossum:
This fixes #440.


justvanrossum:
Hi @ollimeier, I more or less finished this now.

I think all that needs to be done is to finish the new test code in test-glyph-controller.js.

justvanrossum (commented):
Two comments:
- copyObject is unused
- please use `expect(X).to.equal(Y)` (not `.to.deep.equal`) if Y is a number or a string

PR #1111, opened 2024-02-22
Bug 1109 metrics fields

### LINKED ISSUE #1109: Metrics fields: an empty field causes problems, breaks data

ollimeier:
Fixes #1109 

I am not able to get and set the value before, but I guess it would require a more. For now, this is a fix with an info message and after that you can undo you change. What do you think @justvanrossum

### DIFF
--- a/src/fontra/client/web-components/ui-form.js
+++ b/src/fontra/client/web-components/ui-form.js
@@ -183,7 +183,7 @@ export class Form extends SimpleElement {
     inputElement.onchange = (event) => {
       let value = parseFloat(inputElement.value);
       if (isNaN(value)) {
-        message("Wrong input type", `Entered value is not a number: ${value}`);
+        message("Invalid number input", `Entered value is not a number: ${value}.`);
         return;
       }
 
--- a/src/fontra/client/web-components/ui-form.js
+++ b/src/fontra/client/web-components/ui-form.js
@@ -2,7 +2,6 @@ import * as html from "../core/html-utils.js";
 import { SimpleElement } from "../core/html-utils.js";
 import { QueueIterator } from "../core/queue-iterator.js";
 import { hyphenatedToCamelCase } from "../core/utils.js";
-import { dialog, dialogSetup, message } from "/web-components/modal-dialog.js";
 import { RangeSlider } from "/web-components/range-slider.js";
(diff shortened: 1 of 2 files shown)


justvanrossum:
Quick thought: the initialValue should at least be optional. It makes no sense in the non-editing field types. But let me think a bit longer, I think we should just store the previous value so we can access it.

src/fontra/views/editor/panel-selection-info.js
```diff
@@ -463,6 +482,7 @@ export default class SelectionInfoPanel extends Panel {
     }
 
     this.infoForm.onFieldChange = async (fieldItem, value, valueStream) => {
+      fieldItem.initialValue = value;
```
ollimeier, line 485: @justvanrossum  This is the line of code, where I store the latest value.

justvanrossum: That is not correct: we shouldn't use the field item that way.

justvanrossum:
Also: we should never just pop up a dialog in such a case.

src/fontra/client/web-components/ui-form.js
```diff
@@ -181,6 +183,11 @@ export class Form extends SimpleElement {
     };
     inputElement.onchange = (event) => {
       let value = parseFloat(inputElement.value);
+      if (isNaN(value)) {
```
justvanrossum, line 186: Let's move this bit to just before the `this._lastValidFieldValues[fieldItem.key] = value;` line, that way it is more similar to the rotation case.

ollimeier: Sorry, but I don't know where you want to move it.

ollimeier: Isn't it pretty much the same? `if (isNaN(value)) {` is right after `let value = parseFloat(inputElement.value);` in both cases.

justvanrossum: Rereading the code: never mind, the current place is best after all.

(I was hoping the code would look more similar to the rotation case, but I think it indeed needs to be before the "reportValidity" section.

ollimeier: 'Code Design' ;)

PR #1063, opened 2024-01-22
Implement Font Info view

justvanrossum:
This will fix #907, part of #78, and will fix #54. Work in progress.

PR #1112, opened 2024-02-22
Added tooltips attributes

### LINKED ISSUE #1110: Add tooltips to several UI elements
as discussed. Don't use 'title', use the own designed tooltip with 'positioning'

ollimeier:
Fixes #1110 

I added the tooltip attributes, but they actually breaks the current functionality of the UI. 

@fatih-erikli @fabiocaccamo Do you have an idea, why this is the case?

### DIFF
--- a/src/fontra/client/css/tooltip.css
+++ b/src/fontra/client/css/tooltip.css
@@ -24,7 +24,6 @@
 }
 
 [data-tooltip]:hover::before {
-  text-transform: capitalize;
   content: attr(data-tooltip);
   width: max-content;
   max-width: var(--tooltip-max-width);
--- a/src/fontra/client/css/tooltip.css
+++ b/src/fontra/client/css/tooltip.css
@@ -1,4 +1,4 @@
-:host {
+:root {
   --tooltip-background-color-light: var(--fontra-theme-marker) black;
   --tooltip-background-color-dark: white;
   --tooltip-background-color: var(
--- a/src/fontra/client/css/tooltip.css
+++ b/src/fontra/client/css/tooltip.css
@@ -21,10 +21,10 @@
 
 [data-tooltip] {
   position: relative;
-  cursor: default;
 }
 
 [data-tooltip]:hover::before {
+  text-transform: capitalize;
   content: attr(data-tooltip);
   width: max-content;
   max-width: var(--tooltip-max-width);
--- a/src/fontra/views/editor/editor.html
+++ b/src/fontra/views/editor/editor.html
@@ -58,6 +58,7 @@
                 </div>
                 <div
                   class="tool-button"
+                  data-tool="zoom-in"
                   data-tooltip="Zoom in"
                   data-tooltipposition="bottom"
                 >
--- a/src/fontra/views/editor/editor.html
+++ b/src/fontra/views/editor/editor.html
@@ -5,6 +5,7 @@
     <title>Fontra</title>
     <link href="/css/core.css" rel="stylesheet" />
     <link href="/editor/editor.css" rel="stylesheet" />
+    <link href="/css/tooltip.css" rel="stylesheet" />
     <script type="importmap">
       {
         "imports": {
@@ -47,16 +48,35 @@
                 <!-- content will be filled in dynamically -->
               </div>
               <div id="zoom-tools" class="tools-item">
-                <div class="tool-button" data-tool="zoom-out">
+                <div
+                  class="tool-button"
+                  data-tool="zoom-out"
+                  data-tooltip="zoom-out"
+                  data-tooltipposition="bottom"
+                >
                   <inline-svg class="tool-icon" src="/images/minus.svg"></inline-svg>
                 </div>
-                <div class="tool-button" data-tool="zoom-in">
+                <div
+                  class="tool-button"
+                  data-tooltip="zoom-in"
+                  data-tooltipposition="bottom"
+                >
                   <inline-svg class="tool-icon" src="/images/plus.svg"></inline-svg>
                 </div>
-                <div class="tool-button" data-tool="zoom-fit-selection">
+                <div
+                  class="tool-button"
+                  data-tool="zoom-fit-selection"
+                  data-tooltip="zoom-fit-selection"
+                  data-tooltipposition="bottom"
+                >
                   <inline-svg class="tool-icon" src="/images/bullseye.svg"></inline-svg>
                 </div>
-                <div class="tool-button" data-tool="toggle-fullscreen">
+                <div
+                  class="tool-button"
+                  data-tool="toggle-fullscreen"
+                  data-tooltip="toggle-fullscreen"
+                  data-tooltipposition="bottom"
+                >
                   <inline-svg
                     class="tool-icon tool-icon--fullscreen-enter tool-icon--hidden"
                     src="/tabler-icons/maximize.svg"
--- a/src/fontra/views/editor/editor.html
+++ b/src/fontra/views/editor/editor.html
@@ -67,7 +67,7 @@
                 <div
                   class="tool-button"
                   data-tool="zoom-fit-selection"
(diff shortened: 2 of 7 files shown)


justvanrossum:
Yes, I realized this after we spoke: there is a bug in tooltips.css, that's already fixed in the font-info branch, but should be applied to the main branch: https://github.com/googlefonts/fontra/pull/1063/commits/5338eb8695f2b2b761efcd6c7d6055a0846cd56b The first change only. Please see whether this helps.

src/fontra/views/editor/editor.html
```diff
@@ -47,16 +48,35 @@
                 <!-- content will be filled in dynamically -->
               </div>
               <div id="zoom-tools" class="tools-item">
-                <div class="tool-button" data-tool="zoom-out">
+                <div
+                  class="tool-button"
+                  data-tool="zoom-out"
+                  data-tooltip="zoom-out"
```
justvanrossum, line 54: General comment: I would make the tooltip "normal language", in this case "Zoom out". The `"zoom-out"` string is an identifier, good for the machine, less so for humans.

ollimeier:
Hi @fatih-erikli @fabiocaccamo @justvanrossum There is still an issue with the tooltip when it comes to the „top centered tools“. The CSS overflow: hidden; has a negativ impact to the way the tooltip gets displayed. Any idea how to fix that? Please see the video demo:


https://github.com/googlefonts/fontra/assets/21055547/899e3c9e-68bf-42ef-9e5e-714a599e5c91

fabiocaccamo:
@ollimeier why don't just removing the overflow from the tools as you're doing in the video? 
I don't think it has a specific functionality there, or is there any drawback?

ollimeier:
> @ollimeier why don't just removing the overflow from the tools like you did in the video? I don't think it has a specific functionality there, or is there any drawback?

It would change the design. We would lose the round corners.

fabiocaccamo:
@ollimeier I think you can achieve the same result by applying a border radius to the first and last button of each group instead of to the whole group.

justvanrossum:
> It now works, but the design differs a tiny bit, in compare to previous version. What do you think?

Please post a screenshot.

ollimeier:
@justvanrossum Please see:
Screenshot 2024-02-22 at 20 36 17

Screenshot 2024-02-22 at 20 35 40

ollimeier:
The tooltips work:
Screenshot 2024-02-22 at 20 38 17

justvanrossum:
I think we should be able to solve this without changing the design.

FWIW: this `overflow: hidden` code is from relatively early days Fontra CSS design, where I barely knew what I was doing.

ollimeier:
Fixed the style-issue, please see the following screenrecording for reference:

https://github.com/googlefonts/fontra/assets/21055547/fdda43df-c085-4ef2-9a48-9a54a51a689c

Please review @justvanrossum

src/fontra/client/css/tooltip.css
```diff
@@ -21,10 +21,10 @@
 
 [data-tooltip] {
   position: relative;
-  cursor: default;
```
justvanrossum, line 24: Do you remember why you changerd this?

src/fontra/client/css/tooltip.css
```diff
@@ -21,10 +21,10 @@
...
 }
 
 [data-tooltip]:hover::before {
+  text-transform: capitalize;
```
justvanrossum, line 27: I don't think I like this. I prefer the text to start with a cap, but not each word. For example, I prefer "Edit local axes" over "Edit Local Axes".

src/fontra/views/editor/editor.html
```diff
@@ -47,16 +48,36 @@
...
+                  data-tooltip="Zoom in"
+                  data-tooltipposition="bottom"
+                >
                   <inline-svg class="tool-icon" src="/images/plus.svg"></inline-svg>
                 </div>
-                <div class="tool-button" data-tool="zoom-fit-selection">
+                <div
+                  class="tool-button"
+                  data-tool="zoom-fit-selection"
+                  data-tooltip="Zoom fit selection"
```
justvanrossum, line 70: I wonder if a better title would bo "Zoom to fit", or even "Zoom to fit selection"?

src/fontra/views/editor/editor.js
```diff
@@ -446,7 +446,12 @@ export class EditorController {
 
     const editToolsElement = document.querySelector("#edit-tools");
     const toolButton = html.div(
-      { "class": "tool-button selected", "data-tool": tool.identifier },
+      {
+        "class": "tool-button selected",
+        "data-tool": tool.identifier,
+        "data-tooltip": tool.identifier.replace(/-/g, " "),
```
justvanrossum, line 452:
```suggestion
        "data-tooltip": tool.identifier.replaceAll(" ", " "),
```
(The best regex is no regex!)

src/fontra/views/editor/sidebar.js
```diff
@@ -4,6 +4,8 @@ import * as html from "/core/html-utils.js";
 export const MIN_SIDEBAR_WIDTH = 200;
 export const MAX_SIDEBAR_WIDTH = 500;
 
+const panelElementsPostionRight = ["selection-info"];
```
justvanrossum, line 7: The sidebars are very dynamic: it should not be needed to hard-code the "right"-ness of the selection-info panel. sidebar.identifier is "left" or "right", so you can deduce the alignment from that.

justvanrossum:
Oh, one more thing: why do these not align nicely to the center of the tool button?

image

This may be an issue with tooltip.css, @fatih-erikli, could you help out please?

justvanrossum:
Yeah, it seems it's just the arrow that isn't positioned properly, the tooltip box itself looks like it's in the right place.

src/fontra/views/editor/editor.html
ollimeier: Sure, will change this.

src/fontra/views/editor/editor.js
```diff
@@ -446,7 +446,12 @@ export class EditorController {
 
     const editToolsElement = document.querySelector("#edit-tools");
     const toolButton = html.div(
-      { "class": "tool-button selected", "data-tool": tool.identifier },
+      {
+        "class": "tool-button selected",
+        "data-tool": tool.identifier,
+        "data-tooltip": tool.identifier.replaceAll(" ", " "),
```
justvanrossum, line 452:
```suggestion
        "data-tooltip": tool.identifier.replaceAll("-", " "),
```
My mistake in the previous suggestion, but please always check critically.

src/fontra/client/css/tooltip.css
ollimeier: I removed this line of code and used 'capitalizeFirstLetter' instead.

src/fontra/views/editor/sidebar.js
ollimeier: Thanks for the hint. Makes sense. I changed it.

src/fontra/views/editor/editor.js
ollimeier: Works now.

ollimeier:
@justvanrossum Ready for second review.

test-js/test-utils.js
```diff
@@ -180,11 +181,17 @@ describe("hyphenatedToCamelCase", () => {
...
   });
-  it("should not delete the hypen when the second part is not a lowercase letter", () => {
+  it("should not delete the hyphen when the second part is not a lowercase letter", () => {
     expect(hyphenatedToCamelCase("test-1")).equals("test-1");
   });
 });
 
+describe("hyphenatedToLabel", () => {
+  it("should camelize", () => {
+    expect(hyphenatedToLabel("test-case")).equals("Test case");
```
justvanrossum, line 191: Please also test a case with more than one hyphen.

justvanrossum (approved):

PR #1121, opened 2024-02-27
Implement shift-constrain for pen tool add point

### LINKED ISSUE #1059: [pen tool] adding points while holding shift should constrain the resulting line in 45 degree increments
(At least for straight lines)

ollimeier:
Fixes #1059.

### DIFF
--- a/src/fontra/views/editor/edit-tools-pen.js
+++ b/src/fontra/views/editor/edit-tools-pen.js
@@ -141,20 +141,12 @@ export class PenTool extends BaseTool {
       await this.editor.tools["pointer-tool"].handleDrag(eventStream, initialEvent);
       return;
     }
-    console.log("Olli PenTool.handleDrag");
-    console.log("Olli eventStream", eventStream);
-    console.log("Olli initialEvent", initialEvent);
 
     if (this.sceneModel.pathConnectTargetPoint?.segment) {
-      console.log(
-        "Olli if this.sceneModel.pathConnectTargetPoint?.segment: _handleInsertPoint"
-      );
       await this._handleInsertPoint();
     } else if (this.sceneModel.pathInsertHandles) {
-      console.log("Olli if this.sceneModel.pathInsertHandles: _handleInsertHandles");
       await this._handleInsertHandles();
     } else {
-      console.log("Olli else: _handleInsertPoint");
       this._resetHover();
       await this._handleAddPoints(eventStream, initialEvent);
     }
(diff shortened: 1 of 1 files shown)


src/fontra/views/editor/edit-tools-pen.js
```diff
@@ -417,6 +417,15 @@ function setupExistingAnchorPoint(context, path, point, shiftConstrain) {
 }
 
 function insertAnchorPoint(context, path, point, shiftConstrain) {
+  if (shiftConstrain) {
+    if (context.anchorIndex >= 1) {
+      // the path need at least 2 points
+      // to be able to shift constrain
+      const previousPoint = path.getPoint(context.anchorIndex - 1);
+      point = getHandle(point, previousPoint, shiftConstrain);
```
justvanrossum, line 425: - Hmm, we may want to rename `getHandle()`, as we're not dealing with a handle here. I see that the logic is correct.
- The `anchorIndex >= 1` is not what we need to check
- Check for `context.createContour`: we should only constrain when it is `false` (if we're creating a new contour, the inserted point is the first of a new contour, and we don't have a reference point)
- But then we have `context.contourIndex` and `context.contourPointIndex`, and you can get the reference point with `path.getContourPoint(context.contourIndex, context.contourPointIndex)`

justvanrossum: Or actually, let's not reuse `getHandle()` for this, as it redundantly checks for `constrain`, so we can use `shiftConstrain()` directly. `vector.roundVector()` is also already done.

ollimeier: I get your points. Funny enough, but when we use the function shiftConstrain, we get in a conflict with the variable 'shiftConstrain'. We wanted to solve this in a different task. Should I fix it only within 'insertAnchorPoint' or should we do it for the while file for consistency?

justvanrossum: Please only fix it in `insertAnchorPoint()`.

src/fontra/views/editor/edit-tools-pen.js
```diff
@@ -416,7 +416,20 @@ function setupExistingAnchorPoint(context, path, point, shiftConstrain) {
   );
 }
 
-function insertAnchorPoint(context, path, point, shiftConstrain) {
+function insertAnchorPoint(context, path, point, shiftKey) {
+  if (shiftKey) {
+    // Shift-constrain the point 0/45/90 degrees
+    if (context.createContour === false) {
```
justvanrossum, line 422: Please combine this in the `if (shiftKey)` condition, like this:
```js
  if (shiftKey && !context.createContour) {
```
(Generally: don't compare with `false` or `true` directly.)

Issue #1123, opened 2024-02-27
Menu panel should support typing the menu title to select it


justvanrossum:

elevchyt:
From my understanding, this issue refers to focusing on elements of the right-click menu based on their alphabetical order?
For example in the image, if the user presses "C", then "Copy" should be focused first, then if they re-press "C" the "Cut" option should be selected and so on.

![image](https://github.com/googlefonts/fontra/assets/50188474/096b3d3b-57e0-4e0f-ab40-07e1ab0e97f0)

If this is it, I'd love to try solving this issue!

justvanrossum:
I think it's a bit more involved than that: the user just starts typing one or more letters, and the first item matching those letters should be selected. I notice (on macOS) that what you describe also works, but I never noticed that before.

elevchyt:
Understood, I'll post results soon.

elevchyt:
https://github.com/googlefonts/fontra/pull/1127

PR #1127, opened 2024-02-28
support for searching the menu panel via typing item titles

elevchyt:
Made it so that items in the menu panel can be searched by typing a part (or the entirety) of one of the enabled menu options.

_(There seems to be some kind of error when pressing "Enter" to confirm a selected option, but that was the case in the main branch, too. I will create an issue for it and post more info there.)_

### DIFF
--- a/src/fontra/client/web-components/menu-panel.js
+++ b/src/fontra/client/web-components/menu-panel.js
@@ -94,6 +94,8 @@ class MenuPanel extends SimpleElement {
     this.position = position;
     this.positionContainer = positionContainer;
     this.menuElement = html.div({ class: "menu-container", tabindex: 0 });
+    this.menuSearchTimer;
+    this.menuSearchText = "";
 
     // No context menu on our context menu please:
     this.menuElement.oncontextmenu = (event) => event.preventDefault();
@@ -178,6 +180,8 @@ class MenuPanel extends SimpleElement {
 
   handleKeyDown(event) {
     event.stopImmediatePropagation();
+
+    this.searchMenuItems(event.key);
     switch (event.key) {
       case "Escape":
         this.dismiss();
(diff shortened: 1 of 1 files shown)


justvanrossum:
(I edited the description to contain the phrase `This fixes #1123`, which neatly links the PR to the issue.)

justvanrossum:
> (There seems to be some kind of error when pressing "Enter" to confirm a selected option, but that was the case in the main branch, too. I will create an issue for it and post more info there.)

Oh wow, you're right. I swear this used to work, though.

elevchyt:
I created an issue for it if that's ok.
Update the PR description to link to that issue.

src/fontra/client/web-components/menu-panel.js
```diff
@@ -197,6 +201,36 @@ class MenuPanel extends SimpleElement {
     }
   }
 
+  searchMenuItems(key) {
+    // Returns true only for letters, numbers & spaces
+    const isValidSearchInput = /^[a-zA-Z0-9 ]$/.test(key);
+    if (isValidSearchInput) {
```
justvanrossum, line 207: Suggest to return immediately:
```js
    if (!isValidSearchInput) {
      return;
    }
```

(This reduces the indentation for the rest of the function, which in my book is always a win.)

src/fontra/client/web-components/menu-panel.js
```diff
@@ -197,6 +201,36 @@ class MenuPanel extends SimpleElement {
...
+          const itemText = item.textContent.toLowerCase();
+          if (itemText.startsWith(this.menuSearchText)) {
+            foundMatchingItem = true;
+            this.selectItem(item);
+            break;
+          }
+        }
+      }
+
+      // If an item matching the search text is not found, then allow the user to immediately start searching again
```
justvanrossum, line 222: Please wrap this comment to approx. 90 chars or less (just split into two lines).

justvanrossum (commented):
Thank you so much! This works great. I left two comments.

src/fontra/client/web-components/menu-panel.js
```diff
@@ -197,6 +201,39 @@ class MenuPanel extends SimpleElement {
     }
   }
 
+  searchMenuItems(key) {
+    // Returns true only for letters, numbers & spaces
```
justvanrossum, line 205:
```suggestion
    // Accept only letters, numbers & spaces
```

elevchyt:
Thanks for the feedback! All changes have been implemented!

Issue #1128, opened 2024-02-28
Menu panel doesn't support "Enter" presses to confirm selection


elevchyt:
Having an item selected and pressing "Enter" doesn't confirm its selection and throws the error below.

![error on enter](https://github.com/googlefonts/fontra/assets/50188474/ae4a783a-cb18-4e67-bee7-3ccd5de09920)

justvanrossum:
Ah, this must be a regression from a while back :( Great catch!

elevchyt:
Haha thanks, I'll try to fix this one too if that's alright. 😀

PR #1130, opened 2024-02-28
Fixed selected item "enter" key confirmation to use appropriate event

elevchyt:
This fixes https://github.com/googlefonts/fontra/issues/1128.


justvanrossum (approved):
Awesome, thanks!

PR #1132, opened 2024-02-28
Delete .menuSearchTimer

fatih-erikli:
Looks like this line is forgotten. Initializing it with an empty value or deleting it will be better I think.


fatih-erikli:
It's a conflict with that line, I don't want to include it to my pr in that way.

justvanrossum (approved):
If this fixes the conflict, then let’s indeed delete it. Thanks!

fatih-erikli:
No it isn't, but I don't want this line to be included in the commit that I will create for resolving the conflict.

Issue #1134, opened 2024-02-29
[menu bar] Once hovering/clicking, arrow key left and right should navigate to previous/next menu panel


justvanrossum:
<img width="333" alt="image" src="https://github.com/googlefonts/fontra/assets/4246121/46a6506f-10af-45e0-b091-0f3dba198657">

elevchyt:
Apologies, I seem to not understand how to make the menu bar appear.
I'm running this in the browser, doing `fontra --launch filesystem test-py\data\mutatorsans`.

Is there some other way to run this?

justvanrossum:
Is your checkout up-to-data? This was merged this morning only.

PR #1136, opened 2024-02-29
1134 menu bar left right navigate

elevchyt:
Had to remove `event.stopImmediatePropagation()` from the `menu-panel.js` due to it blocking events being fired from `menu-bar.js`.
Please, tell me if this is something that might be breaking something (from my testing it was fine).

Also, I did a tiny fix for the "View" menu item since it didn't work and threw an error if nothing was typed to be shown inside the viewbox.
error1
ok

Any feedback is greatly appreciated!

This resolves https://github.com/googlefonts/fontra/issues/1134.

The "View" menu change fixes https://github.com/googlefonts/fontra/issues/1135.

### DIFF
--- a/src/fontra/client/web-components/menu-bar.js
+++ b/src/fontra/client/web-components/menu-bar.js
@@ -154,6 +154,7 @@ export class MenuBar extends SimpleElement {
   }
 
   navigateMenuBar(arrowKey) {
+    this.unhoverMenuItems();
     const currentSelection = this.contentElement.querySelector(".current");
     const menuItemElements = this.contentElement.children;
     const currentSelectionIndex = Array.prototype.indexOf.call(
--- a/src/fontra/client/web-components/menu-bar.js
+++ b/src/fontra/client/web-components/menu-bar.js
@@ -36,6 +36,7 @@ export class MenuBar extends SimpleElement {
     window.addEventListener("blur", this.onBlur.bind(this));
     this.contentElement.addEventListener("mouseover", this.onMouseover.bind(this));
     this.contentElement.addEventListener("click", this.onClick.bind(this));
+    this.contentElement.addEventListener("keydown", this.handleKeyDown.bind(this));
     this.showMenuWhenHover = false;
   }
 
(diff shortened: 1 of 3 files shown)


justvanrossum:
The issue of the View menu is this one: https://github.com/googlefonts/fontra/issues/1135

src/fontra/views/editor/editor.js
```diff
@@ -312,14 +312,16 @@ export class EditorController {
               title: "Zoom to fit",
               enabled: () => {
                 let viewBox = this.sceneController.getSelectionBox();
-                const size = rectSize(viewBox);
-                if (size.width < 4 && size.height < 4) {
-                  const center = rectCenter(viewBox);
-                  viewBox = centeredRect(center.x, center.y, 10, 10);
-                } else {
-                  viewBox = rectAddMargin(viewBox, 0.1);
+                if (viewBox) {
```
justvanrossum, line 315: Since "enabled" is supposed to return a bool, I would reverse this:
```js
    if (!viewBox) {
      return false;
    }
```
(That also makes the diff to the original smaller)

src/fontra/client/web-components/menu-panel.js
```diff
@@ -266,8 +266,6 @@ export class MenuPanel extends SimpleElement {
   }
 
   handleKeyDown(event) {
-    event.stopImmediatePropagation();
```
justvanrossum, line 269: Thanks for pointing that out. I will have to play with that. There may be a reason why that's there, but I don't know off the top of my head.

src/fontra/client/web-components/menu-bar.js
```diff
@@ -120,6 +121,38 @@ export class MenuBar extends SimpleElement {
...
+
+  navigateMenuBar(arrowKey) {
+    const currentSelection = this.contentElement.querySelector(".current");
+    const menuItemElements = this.contentElement.children;
+    const currentSelectionIndex = Array.prototype.indexOf.call(
+      menuItemElements,
+      currentSelection
+    );
+    let newSelectionIndex;
+    arrowKey == "ArrowLeft"
```
justvanrossum, line 142: The ternary operator (`?`) is an expression itself, and one generally shouldn't use the branches for side effects (in this case: assign to a variable).

This can be more elegantly written using the `?` expression as an expression (untested):
```js
    const newSelectionIndex = currentSelectionIndex + (arrowKey == "ArrowLeft" ? -1 : +1);
```
(Not sure if the parentheses are needed, prettier will tell you...)

elevchyt: This is really smart! Thank you!

src/fontra/views/editor/editor.js
elevchyt: Ahh, I really should have done that. 😅

elevchyt:
Thank you for the great feedback!
I just committed the changes you recommended.

justvanrossum:
The hover highlight of the menu title should be reset when selecting the next/previous menu panel:

image

(Here, "View" is still highlighed, but only "Glyph" should be highlighted.)

elevchyt:
Yeah, I wasn't sure about that and forgot to ask.
Will look into it right now.

elevchyt:
I just pushed my solution to this.
At first, I tried to dispatch a mouseleave event programmatically, but that doesn't change the hover state that CSS uses.
My current approach toggles the hovered state programmatically.

justvanrossum:
Hm, it's weird, I confirm I have your latest changes, but I still see the "double" hover title:

https://github.com/googlefonts/fontra/assets/4246121/b9f68364-bf6d-42aa-9b43-9c6d377b0374

The "View" menu should not be highlighted once navigated to a neighbor. Do you also see that?

elevchyt:
My bad.
I removed a line of code in testing and committed my changes without it.
Should work as expected now.

PR #1137, opened 2024-03-01
Fixes menu bar color in dark theme

navv-1:
This is screenshot from chrome on windows
image


justvanrossum (approved):
Good catch, thanks! It's odd that it does behave correctly on macOS.

justvanrossum:
This is extremely helpful! Much appreciated.

PR #1139, opened 2024-03-01
Refactoring 'unicode' to 'codePoint'

### LINKED ISSUE #1105: Use 'code point' instead of 'unicode'
Just:
„There are many instances in the code where we use the word "unicode" instead of "code point" [...]. We should make it a habit to always use the term "code point" consistently.
Especially argument names may be a little tricky, if they are used as keyword args
“

Related to #1099

ollimeier:
Fixes #1105

### DIFF
--- a/src/fontra/backends/ufo_utils.py
+++ b/src/fontra/backends/ufo_utils.py
@@ -7,7 +7,7 @@ logger = logging.getLogger(__name__)
 
 
 _glyphNamePat = re.compile(rb'<glyph\s+name\s*=\s*"([^"]+)"')
-_unicodePat = re.compile(rb'<unicode\s+hex\s*=\s*"([^"]+)"')
+_codePointPat = re.compile(rb'<unicode\s+hex\s*=\s*"([^"]+)"')
 
 
 def extractGlyphNameAndUnicodes(
@@ -24,5 +24,5 @@ def extractGlyphNameAndUnicodes(
                 "actual file name does not match predicted file name: "
                 f"{refFileName} {fileName} {glyphName}"
             )
-    unicodes = [int(u, 16) for u in _unicodePat.findall(data)]
-    return glyphName, unicodes
+    codePoints = [int(u, 16) for u in _codePointPat.findall(data)]
+    return glyphName, codePoints
--- a/src/fontra/backends/ufo_utils.py
+++ b/src/fontra/backends/ufo_utils.py
@@ -7,7 +7,7 @@ logger = logging.getLogger(__name__)
 
 
 _glyphNamePat = re.compile(rb'<glyph\s+name\s*=\s*"([^"]+)"')
-_codePointPat = re.compile(rb'<unicode\s+hex\s*=\s*"([^"]+)"')
+_unicodePat = re.compile(rb'<unicode\s+hex\s*=\s*"([^"]+)"')
 
 
 def extractGlyphNameAndCodePoints(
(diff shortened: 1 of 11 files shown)


ollimeier:
@justvanrossum This is nothing for today. But as I started it already, I wanted to bring this to an end. All tests are passing. But as this effects many different places in the code, we may want to be careful with such an update. Probably something for next week.

justvanrossum:
Please also do this on the fontra-rcjk repo, which actually imports `extractGlyphNameAndUnicodes()`, so would break...

https://github.com/googlefonts/fontra-rcjk/blob/0866d4358ad1ee9afd68133b3cff0886c1331247/src/fontra_rcjk/backend_fs.py#L12

src/fontra/backends/ufo_utils.py
```diff
@@ -7,10 +7,10 @@
 
 
 _glyphNamePat = re.compile(rb'<glyph\s+name\s*=\s*"([^"]+)"')
-_unicodePat = re.compile(rb'<unicode\s+hex\s*=\s*"([^"]+)"')
+_codePointPat = re.compile(rb'<unicode\s+hex\s*=\s*"([^"]+)"')
```
justvanrossum, line 10: Suggest to revert this rename: this regex literally matches `<unicode ...` so the original name makes some sense.

src/fontra/backends/ufo_utils.py
```diff
@@ -24,5 +24,5 @@ def extractGlyphNameAndUnicodes(
                 "actual file name does not match predicted file name: "
                 f"{refFileName} {fileName} {glyphName}"
             )
-    unicodes = [int(u, 16) for u in _unicodePat.findall(data)]
-    return glyphName, unicodes
+    codePoints = [int(u, 16) for u in _codePointPat.findall(data)]
```
justvanrossum, line 27: But: keep the variable name, only revert the `_codePointPat` name.

src/fontra/views/editor/panel-selection-info.js
```diff
@@ -141,18 +141,18 @@ export default class SelectionInfoPanel extends Panel {
...
       // Glyph does not yet exist in the font, but we can grab the unicode from
       // selectedGlyphInfo.character anyway
-      unicodes = [selectedGlyphInfo.character.codePointAt(0)];
+      codePoints = [selectedGlyphInfo.character.codePointAt(0)];
     }
 
-    const unicodesStr = makeUnicodesString(unicodes);
-    let baseUnicodesStr;
-    if (glyphName && !unicodes.length) {
+    const codePointsStr = makeCodePointsString(codePoints);
```
justvanrossum, line 151: I'm a little bot torn about the changes here: we're producing a string that is shown to the user with a "Unicode:" label (which I think is good). Same for `baseUnicodesStr`. Let's discuss with a fresh mind.

src/fontra/backends/ufo_utils.py
ollimeier: OK. Did it.

src/fontra/views/editor/panel-selection-info.js
ollimeier: @justvanrossum 
The question is, when do we want or have to use the term unicode and when is codePoint better? Is a unicode a string representation of a codePoint? The decimal value 65 is a codePoint, but converted to Hex and displayed as a string like '0041' is a unicode?
65 = codePoint
'0041' = unicode

justvanrossum: I think of `U+0041` as a "unicode code point", so in the end it doesn't matter all that much how we name the variables here :) Let's keep your change, I think it's fine.

Issue #1141, opened 2024-03-01
Tooltip in sidebar can get clipped


justvanrossum:
<img width="186" alt="image" src="https://github.com/googlefonts/fontra/assets/4246121/be2eb0fb-3cbb-4d0e-bbfe-09527056dc50">

Probably this is because the sidebars have `overflow: hidden`.

justvanrossum:
Thanks for the testing and suggestions! I will go with the bottom alignment, as you suggested.

Closing this in favor of #1153.

justvanrossum:
Fixed by #1153.

PR #1143, opened 2024-03-02
changed sidebar z-index to match topbar's z-index

elevchyt:
Looks like this is happening because the `sidebar-container` class had a z-index of 100, while the `top-bar-container` class had a 
z-index of 200.

Changing the `sidebar-container` z-index to 200 seems to fix it pretty nicely without breaking anything else.

This fixes https://github.com/googlefonts/fontra/issues/1141.

### DIFF
--- a/src/fontra/views/editor/editor.css
+++ b/src/fontra/views/editor/editor.css
@@ -131,7 +131,6 @@ body {
   grid-row-end: 2;
   grid-column-start: 1;
   grid-column-end: 4;
-  z-index: 200;
   height: var(--editor-top-bar-height);
   background: var(--editor-top-bar-background-color);
   border-bottom: 1px solid var(--editor-top-bar-border-color);
--- a/src/fontra/views/editor/editor.css
+++ b/src/fontra/views/editor/editor.css
@@ -222,7 +222,7 @@ body {
   position: absolute;
   display: none;
 
-  box-shadow: 0px 0px 8px #0006;
+  box-shadow: 0px 8px 8px #0006;
   top: 0;
   width: 50px; /* arbitrary > blur radius */
   height: 100%;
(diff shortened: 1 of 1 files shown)


navv-1:
hmm, it seems this has a side effect on the menu panel
image

elevchyt:
Thanks! Just pushed a change.
Should work as intended now.

justvanrossum:
This causes the shadow box to bleed into the menu bar:

image

(That shadow stuff is a bit messy indeed, but that might have been the reason why the menu bar is on top of the sidebar.)

elevchyt:
After looking at the layout for a while, I understand now that the side panel is basically two panels stuck together and the one on the right is the only one having a shadow.
image

Could I suggest that `sidebar-shadow-box` gets `box-shadow: 0px 8px 8px #0006;` instead of the current one which is `box-shadow: 0px 0px 8px #0006;`?

Seems to be the easiest fix, but it changes the shadows look very slightly.
Alternative solutions gladly accepted.

elevchyt:
I added the change so you can check it out.

justvanrossum:
Can you post a before/after screenshot that highlights the difference?

justvanrossum:
Never mind, I looked. It's the bit of shadow at the top, right?

image

elevchyt:
Yes

navv-1:
What if you move the tooltip position to the bottom? So, the shadow does not need to be changed.
image

justvanrossum:
That's an option, too. Let me think.

justvanrossum:
Yes, let's do that (align tooltips below): I am tweaking the appearance of the menu bar a bit, and there are other reasons why it's nice to have it on top of the rest.

justvanrossum:
See https://github.com/googlefonts/fontra/pull/1146

justvanrossum:
Thanks for the testing and suggestions! I will go with the bottom alignment, as you suggested.

Closing this in favor of #1153.

PR #1170, opened 2024-03-06
Bugfix: UI issue with edit local axes

### LINKED ISSUE #1168: Edit local axes: dialog can overflow
image

ollimeier:
Fixes #1168


justvanrossum (approved):
I wish I could understand why it worked before and doesn't now, though. This used to not be needed.

PR #1171, opened 2024-03-06
Bugfix: UI issues with ui-form-label

### LINKED ISSUE #1167: [ui-form] labels are no longer truncated with ...
image

ollimeier:
Fixes #1167


justvanrossum (commented):
Please try with a long axis name that contains spaces or hyphens, then it will wrap into multiple lines:
image
This probably was already the case, but perhaps while you're at it we can prevent wrapping?

justvanrossum (approved):
Great, thanks!

PR #1178, opened 2024-03-07
Tidy up the menu bar

### LINKED ISSUE #1177: Menu bar dimension
The top bar container height is 35px and my screen width is 1920px, so, I think the menu bar need to tidy up.
image

navv-1:
Now, menu bar height matches with container and the width doesn't overflow.
fix #1177


justvanrossum (approved):
Thank you so much!

PR #1184, opened 2024-03-08
Add shape tool

ollimeier:
I made a new branch from @fabiocaccamo branch and fixed the conflicts. Now I try to understand how I actually can draw shapes and add them.


ollimeier:
@justvanrossum Topic short cuts: with the position of the new tool we change the used short cuts. We may want to rethink this. I was wondering if we should have a look to other apps which are currently used heavily by designers like figma:
https://www.figmashortcuts.com/
For example:
Pen -> P
Pencil -> ⇧ + P
Ellipse -> O
Rectangle -> R

justvanrossum:
> Topic short cuts

Maybe. I'm mostly ok with the current numeric approach: it just matches the order of the tools. Also, at this point I'm not too worried about the implicit change by adding the shape tool.

That is not to say I am open to rethink this, but I don't think it's a high priority issue. Feel free to open an issue so we can get back to this later.

ollimeier:
I am closing this PR because this will be part of #1225

PR #1180, opened 2024-03-08
Fixing scroll issue

ollimeier:
Fixes #1175

Issue #1175, opened 2024-03-07
Selection info shows scroll bar


ollimeier:
Feedback from user Mark:

Not sure if this is intended: now on the right panel, when the panel is too narrow or if the entry name is too long, we get a horizontal scroll bar under the name. But when I scroll to the right I still can’t see the whole name. :sweat_smile:

![Capture d’écran 2024-03-07 à 15 25 25](https://github.com/googlefonts/fontra/assets/21055547/2395c2e5-09e3-4f37-903d-75387e2b2d7d)

ollimeier:
A solution for not displaying the scrollbar might be:
```
-ms-overflow-style: none;
scrollbar-width: none;
```

navv-1:
I think just change `overflow-x: auto;` to `overflow-x: hidden;`

ollimeier:
@navv-1 Sure, overflow-x: hidden; will remove the scroll bar,BUT you won't be able to scroll at all in x direction. So, no possibility to read the full name.

ollimeier:
@justvanrossum and @navv-1 I would prefer to be able to scroll, but hide the scroll bar:
https://github.com/googlefonts/fontra/pull/1180

justvanrossum:
The name should be truncated with "...", scrollable at that small space makes no sense to me at all.

PR #1187, opened 2024-03-09
Modal dialog improvement

### LINKED ISSUE #1176: Modal dialog shows scroll bar
I think this is not intended
image

navv-1:
This pr will fix this issue,
image

and fix #1176

### DIFF
--- a/src/fontra/client/web-components/modal-dialog.js
+++ b/src/fontra/client/web-components/modal-dialog.js
@@ -60,7 +60,6 @@ export class ModalDialog extends SimpleElement {
 
       outline: none; /* to catch key events we need to focus, but we don't want a focus border */
       max-width: 32em;
-      min-height: 10em;
       max-height: 80vh;
       overflow-wrap: normal;
       font-size: 1.15em;
--- a/src/fontra/client/web-components/modal-dialog.js
+++ b/src/fontra/client/web-components/modal-dialog.js
@@ -60,7 +60,7 @@ export class ModalDialog extends SimpleElement {
 
       outline: none; /* to catch key events we need to focus, but we don't want a focus border */
       max-width: 32em;
-      max-height: 32em; /* changed from vh because of problematic when window resized */
+      max-height: 80vh;
       overflow-wrap: normal;
       font-size: 1.15em;
       background-color: var(--ui-element-background-color);
--- a/src/fontra/client/web-components/modal-dialog.js
+++ b/src/fontra/client/web-components/modal-dialog.js
@@ -60,6 +60,7 @@ export class ModalDialog extends SimpleElement {
 
       outline: none; /* to catch key events we need to focus, but we don't want a focus border */
       max-width: 32em;
+      min-height: 10em;
       max-height: 80vh;
       overflow-wrap: normal;
       font-size: 1.15em;
--- a/src/fontra/client/web-components/modal-dialog.js
+++ b/src/fontra/client/web-components/modal-dialog.js
@@ -36,6 +36,17 @@ export class ModalDialog extends SimpleElement {
       white-space: normal;
     }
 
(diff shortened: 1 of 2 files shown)


justvanrossum:
This must be related to https://github.com/googlefonts/fontra/issues/1175 and https://github.com/googlefonts/fontra/pull/1180/files, and I had an uneasy feeling about the fix.

I have an uneasy feeling about the fix in this PR, too: I don't think `max-height: 32em;` is what we want: it should be able to get as big as possible, with some space above and below. 

Indeed, I just tested: try "Add Component", this should show a big list, and it should be as big as possible (within roughly the 80vh).

navv-1:
Actually, we can also remove the max-height and let it overflow since we have the scrollbar on the edge
image

justvanrossum:
Hmm,  we shouldn't have that scrollbar at the edge. The dialog should fit on the page (with some margin above and below), and any scrolling should be handled by its contents.

Your original problem is that we get too many scrollbars when the window isn't tall enough, right?

justvanrossum:
Can you compare how the situation was before https://github.com/googlefonts/fontra/pull/1146? I feel that PR had larger consequences than intended.

navv-1:
Look at the add and remove button, if we use vh we will get this result when the window isn't tall enough
image

justvanrossum:
So what I'm asking is, did that issue exist before #1146 was merged?

navv-1:
this is after #1145 merged
image
image

justvanrossum:
Ok, so we need to back up more. Was it ok before #1145? Was it ever ok? I need to know whether this is a regression or something that was always broken.

navv-1:
I tried after #929 merged and it's not different

justvanrossum:
And before that one?

navv-1:
Yes, even before the menu bar was merged. I think it's not regression.

navv-1:
> Indeed, I just tested: try "Add Component", this should show a big list, and it should be as big as possible (within roughly the 80vh).

Instead of looking at a long list, I prefer to use the search bar and get a shorter list.

justvanrossum:
> Instead of looking at a long list, I prefer to use the search bar and get a shorter list.

I disagree. I don't want a hardcoded maximum height for the dialog.

I feel that https://github.com/googlefonts/fontra/commit/e98e75c3728d6ba020ca5e3f45aed5d6231da5d9 / https://github.com/googlefonts/fontra/pull/1170 was wrong, and we should fix that problem in a different way.

navv-1:
I just brought back the `max-height`. The only problem that I found is in the add and remove button. But this kind of situation maybe would never happen in the real world :D
image

justvanrossum:
Thanks!

That might be related to the minHeight of the axis list:

https://github.com/googlefonts/fontra/blob/c6d24ef0f36e26aab86b1e43e08f372208447ba0/src/fontra/views/editor/panel-designspace-navigation.js#L965

justvanrossum:
Changing that to "3em" seems to improve it.

navv-1:
I changed that to 10vh and I think it's much better.

image

I also add `min-height: 10em` to the dialog box to prevent it from shrinking too short. It makes dialog a bit higher when it has a short message. I hope it's still acceptable.

image

justvanrossum:
Hm, I'm not happy with the use of vh in the list minHeight. This should be an absolute value, and should not depend on the view height. You can tweak the exact value of 3em, but please use em.

Secondly, I am not happy with the taller dialog for short messages. If this is to make sure the dialog doesn't break for a very small window height: I'm actually ok with some breakage under such tight conditions. The app is not really meant for such small windows.

justvanrossum (approved):

justvanrossum:
Thank you once again for your contributions! It's much appreciated.

PR #1202, opened 2024-03-13
Safari issue: index label

### LINKED ISSUE #1183: [index labels] consider a fallback for canvas context.roundRect()
roundRect() is only supported in Safari >= 16.0.

While Fontra does not support all Safari 15 versions, apparently it still work for 15.6.1.

Let's consider providing a fallback for roundRect. In the case of our labels, perhaps it's good enough to just fall back to a plain rect.

If we agree that's the case, let's add a helper function to the visualisation layer defs module. Perhaps it can look something like this:

```js
function maybeRoundRect(context, ...args) {
  if (context.roundRect} {
    context.roundRect(...)
  } else {
    context.rect(...)
  }
}
```

Or should be "properly" add a polyfill to the context class/prototype?

ollimeier:
Fixes #1183

### DIFF
--- a/src/fontra/views/editor/visualization-layer-definitions.js
+++ b/src/fontra/views/editor/visualization-layer-definitions.js
@@ -575,7 +575,8 @@ registerVisualizationLayerDefinition({
 
       context.fillStyle = parameters.boxColor;
       context.beginPath();
-      context.roundRect(
+      drawRoundRect(
+        context,
         startPoint.x - width / 2,
         -startPoint.y - bottomY + margin,
         width,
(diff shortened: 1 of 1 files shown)


src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -1109,3 +1107,15 @@ function* iterComponentOriginsByIndex(components, componentIndices) {
 //   colorsDarkMode: {},
 //   draw: (context, positionedGlyph, parameters, model, controller) => { /* ... */ },
 // }
+
+function drawRoundRect(context, ...args) {
```
justvanrossum, line 1111: Let's spell out the arguments:
```js
roundRect(x, y, width, height, radii)
```

ollimeier: Done.

PR #1225, opened 2024-03-19
Adding shape tools for rectangle and ellipse

### LINKED ISSUE #198: Add "shape" tool, to draw rectangles and ovals
Part of #9.

### LINKED ISSUE #1227: Tooltips can be clicked on, even after they became invisible
As observed here: https://github.com/googlefonts/fontra/pull/1225#discussion_r1531750216

Clicking a tooltip should not behave like that.

### LINKED ISSUE #1233: Improve tooltip behavior
Generally looking good!

I wonder: when click-holding on the tool, the tooltip shows up, but maybe it shouldn't. Could be a separate task, but I think the tooltip should not appear when the tooltipped item is being clicked on (maybe the `:active` CSS selector can be used)

Secondly, I think the tools that have selectable sub tools need a visual indicator that it does have subtools. Perhaps a tiny triangle in the bottom right of the icon area.

_Originally posted by @justvanrossum in https://github.com/googlefonts/fontra/pull/1225#pullrequestreview-1955166488_

ollimeier:
Fixes #198
Fixes #1227
Fixes #1233


src/fontra/client/core/var-path.js
```diff
@@ -546,7 +546,7 @@ export class VarPackedPath {
     this._appendPoint(x, y, VarPackedPath.ON_CURVE);
   }
 
-  cubicCurveTo(x1, y1, x2, y2, x3, y3) {
+  bezierCurveTo(x1, y1, x2, y2, x3, y3) {
```
justvanrossum, line 549: This change is what is causing the js unit tests fail. Instead, please add a *new* method `bezierCurveTo`, that simply calls `cubicCurveTo`, and document it as a "compatibility method for Path2D".

src/fontra/client/css/tooltip.css
```diff
@@ -38,7 +38,7 @@
   color: var(--tooltip-foreground-color);
   background: var(--tooltip-background-color);
   z-index: 1;
-  opacity: 0;
+  visibility: hidden;
```
justvanrossum, line 41: Why this change? I think visibility is not animatable fluidly as opacity is. I'm pretty sure we have a transition.

src/fontra/views/editor/edit-tools-shape-ellipse.js
```diff
@@ -0,0 +1,77 @@
...
+import { range } from "../core/utils.js";
+import { ShapeToolRect } from "./edit-tools-shape.js";
+
+const bezierArcMagic = 0.5522847498; // constant for drawing circular arcs w/ Beziers
+
+export class ShapeToolEllipse extends ShapeToolRect {
+  iconPath = "/tabler-icons/circle-plus-2.svg";
+  identifier = "shape-tool-ellipse";
+
+  drawShapePath(path, x, y, width, height, revesed, centered) {
```
justvanrossum, line 11: `reversed`

src/fontra/views/editor/edit-tools-shape-ellipse.js
```diff
@@ -0,0 +1,77 @@
...
+  for (let i = 0; i < 4; i++) {
+    path.bezierCurveTo(
+      Math.round(cx + rx * h1x + shiftX),
+      Math.round(cy + ry * h1y + shiftY),
+      Math.round(cx + rx * h2x + shiftX),
+      Math.round(cy + ry * h2y + shiftY),
+      Math.round(cx + rx * x + shiftX),
+      Math.round(cy + ry * y + shiftY)
+    );
+    let tempH1x = h1x,
```
justvanrossum, line 63: You should be able to write
```js
[h1x, h1y] = [-h1y, h1x];
```

src/fontra/views/editor/edit-tools-shape.js
```diff
@@ -0,0 +1,188 @@
...
+    if (event.shiftKey) {
+      // make square, not rectangle
+      if ((width > 0 && height > 0) || (width < 0 && height < 0)) {
+        height = width;
+      } else {
+        height = -width;
+      }
+    }
+
+    let revesed = event.ctrlKey ? true : false; // reversed contour direction
```
justvanrossum, line 124:
```suggestion
    let reversed = event.ctrlKey; // reversed contour direction
```
This property is already a boolean. Same for the next line.

justvanrossum: At the same time, I think all flags should be handled here, like the shiftKey is handled here.

Define drawShapePath like this:
```js
    drawShapePath(path, x, y, width, height) {
```
And feed it the "right" arguments based on the modifiers.

Reverse also doesn't need to be special cased. There are two ways to do it:
- reverse the path after it is constructed
- use a negative width or height, adjust x or y

src/fontra/client/css/tooltip.css
ollimeier: This is necessary, because otherwise the hidden tooltip causes issues with the toolbar. You cannot click on a toolbar item, when there is a hidden tooltip. does that make sense to you? I could demo it, if you want. But this change is definitely necessary.

justvanrossum: I think we have a different problem with tooltips that I just noticed.

Example: go to the fullscreen button, hover until the tooltip comes. Then click on the tooltip: it's as if you're clicking on the button itself.

Worse: hover over the fullscreen button until the tooltip appears, then hover over the tooltip until it disappears, then click: it is as if the button was clicked...

The tooltip should probably have `pointer-events: none` set.

I think your `visibility` changes tries to address this problem, but in the wrong way, as we should be able to use opacity.

src/fontra/client/core/var-path.js
ollimeier: Thanks. I actually missed to mention that I was aware of the failing unit test, but wanted to talk to you about how we should deal with that. (Not enough time yesterday.) If I should adjust the unittest or a different solution. Now I know how to proceed :) 
Will fix it immediately.

src/fontra/client/css/tooltip.css
ollimeier: I see. I will give it a try and let you know.

src/fontra/views/editor/edit-tools-shape-ellipse.js
ollimeier: Thanks. Done.

src/fontra/client/css/tooltip.css
justvanrossum: I opened a separate issue: https://github.com/googlefonts/fontra/issues/1227

src/fontra/views/editor/edit-tools-shape-ellipse.js
ollimeier: Thanks. Changed it.

src/fontra/client/css/tooltip.css
ollimeier: Thanks a lot for the hint with 'pointer-events: none' this fixes the tooltip issue. Done.

src/fontra/views/editor/edit-tools-shape.js
ollimeier: I agree. As mentioned yesterday, I had some issues with that. I started to add a reverse method, but as we have to deal with two different objects (path2D and VarPackedPath) the same way, I was not able to make it work yesterday, why I ended up with this solution. As said: not nice, but it works.

ollimeier: For reverence, reversePath vs. flip:
Screenshot 2024-03-20 at 14 26 29
Different start point location.

src/fontra/client/core/glyph-svg.js
```diff
@@ -31,7 +31,12 @@ export class SVGPath2D {
     this.items.push("L" + this._format(x, y));
   }
 
+  // compatibility method for Path2D
```
justvanrossum, line 34: Why this change? I don't think glyph-svg is relevant for this PR.

src/fontra/views/editor/edit-tools-shape-ellipse.js
```diff
@@ -0,0 +1,31 @@
+import * as rectangle from "../core/rectangle.js";
+import { range } from "../core/utils.js";
+import { ShapeToolRect } from "./edit-tools-shape.js";
+
+const bezierArcMagic = 0.5522847498; // constant for drawing circular arcs w/ Beziers
```
justvanrossum, line 5: Please move all relevant content the edit-tools-shape.js module. We don't need a separate module for this.

src/fontra/views/editor/edit-tools-shape.js
```diff
@@ -0,0 +1,211 @@
...
+import { range } from "../core/utils.js";
+import { VarPackedPath, packContour } from "../core/var-path.js";
+import { BaseTool, shouldInitiateDrag } from "./edit-tools-base.js";
+import { registerVisualizationLayerDefinition } from "./visualization-layer-definitions.js";
+
+export class ShapeToolRect extends BaseTool {
+  iconPath = "/tabler-icons/square-plus-2.svg";
+  identifier = "shape-tool-rectangle";
+
+  handleHover(event) {}
```
justvanrossum, line 11: This is redundant, and overwritten below.

src/fontra/views/editor/edit-tools-shape.js
```diff
@@ -0,0 +1,211 @@
...
+    if (event.altKey) {
+      width = width * 2;
+      height = height * 2;
+      x = x - width / 2;
+      y = y - height / 2;
+    }
+
+    const unpackedContours = this.getUnpackedContour(x, y, width, height);
+
+    const path = new VarPackedPath();
```
justvanrossum, line 142:
```js
   const path = VarPackedPath.fromUnpackedPath(unpackedContours);
```

src/fontra/views/editor/edit-tools-shape-ellipse.js
```diff
@@ -0,0 +1,31 @@
...
+  let points = [];
+  let [x, y] = [1, 0];
+  points.push({ x: cx + rx * x, y: cy + ry * y, smooth: true });
+  for (let i = 0; i < 4; i++) {
+    points.push({ x: cx + rx * (x - y * t), y: cy + ry * (x * t + y), type: "cubic" });
+    points.push({ x: cx + rx * (x * t - y), y: cy + ry * (x + y * t), type: "cubic" });
+    points.push({ x: cx + rx * -y, y: cy + ry * x, smooth: true });
+    [x, y] = [-y, x];
+  }
+  points.pop(); // remove last point
```
justvanrossum, line 29: Instead, you can move the first points.push() to inside the loop, then remove the last points.push(). Should give the exact same result.

justvanrossum (commented):
Pop-down tool behavior:
- a single click should not show the options
- click-hold should show the options after a delay
- then they should stay, even if the user releases the mouse
- they should go once the user chooses a tool (either by new click, or releasing from the original click)

Visually, there should be some sort of border around the optional tools, or shadow.

The stroke width for the icons is a bit too thick. For the fullscreen icon, this is customized with the `tool-icon--fullscreen-enter` and `tool-icon--fullscreen-exit` classes. I think this should be generalized, so we just add a single class to those icons that we took from tabler, to correct that in a consistent way.

The rectangle goes counter-clockwise, the ellipse goes clockwise. I think both should go counter-clockwise (as per most common convention).

src/fontra/client/core/glyph-svg.js
ollimeier: Done.

ollimeier: Had to undo this, because otherwise checks fail:
Screenshot 2024-03-21 at 15 20 24

src/fontra/views/editor/edit-tools-shape-ellipse.js
ollimeier: I removed unnecessary imports.

src/fontra/views/editor/edit-tools-shape.js
ollimeier: Thanks. done.

src/fontra/views/editor/edit-tools-shape.js
ollimeier: Thanks, changed it by using VarPackedPath.fromUnpackedContours.

src/fontra/views/editor/edit-tools-shape-ellipse.js
ollimeier: Thanks. Yes, it works the same. Changed it.

src/fontra/client/core/glyph-svg.js
justvanrossum: Compare what it looks like in the main branch. This function should not be touched in this PR.

ollimeier:
Hi @justvanrossum ,

a single click should not show the options **-> Done.**
click-hold should show the options after a delay **-> Done.**
then they should stay, even if the user releases the mouse **-> Done.**
they should go once the user chooses a tool (either by new click, or releasing from the original click) **-> Done.**
Visually, there should be some sort of border around the optional tools, or shadow. **-> Done.**

The stroke width for the icons is a bit too thick. **-> Done.**
For the fullscreen icon, this is customized with the tool-icon--fullscreen-enter and tool-icon--fullscreen-exit classes. I think this should be generalized, so we just add a single class to those icons that we took from tabler, to correct that in a consistent way. **-> Seems like all SVG from tabler include the CSS class 'icon-tabler' why I included this to modify the stroke width. Further refactoring might be needed, but probably this should be separated from this task??? **

The rectangle goes counter-clockwise, the ellipse goes clockwise. I think both should go counter-clockwise (as per most common convention). **-> This is not done, yet. Still TODO**

ollimeier:
The rectangle goes counter-clockwise, the ellipse goes clockwise. I think both should go counter-clockwise (as per most common convention).  **-> Done.**

src/fontra/client/core/glyph-svg.js
ollimeier: Sorry, my bad. done.

ollimeier:
@justvanrossum Ready for review.

src/fontra/views/editor/edit-tools-shape.js
```diff
@@ -0,0 +1,204 @@
...
+
+  drawShapeVarPackedPath(mouseRect, event) {
+    let x = mouseRect.xMin;
+    let y = mouseRect.yMin;
+    let width = mouseRect.xMax - mouseRect.xMin;
+    let height = mouseRect.yMax - mouseRect.yMin;
+
+    // make square, not rectangle
+    if (event.shiftKey) {
+      if ((width > 0 && height > 0) || (width < 0 && height < 0)) {
```
justvanrossum, line 123: Suggest trying something like:
```js
    const size = (Math.abs(width) + Math.abs(height)) / 2;
    width = size * Math.sign(width);
    height = size * Math.sign(height);
```

src/fontra/views/editor/edit-tools-shape.js
```diff
@@ -0,0 +1,204 @@
...
+        height = width;
+      } else {
+        height = -width;
+      }
+    }
+
+    // positon at center
+    if (event.altKey) {
+      width = width * 2;
```
justvanrossum, line 132: If you do the x/y modification first, you won't have to divide by two.

Also, you can write `width *= 2;`

justvanrossum (commented):
Generally looking good!

I wonder: when click-holding on the tool, the tooltip shows up, but maybe it shouldn't. Could be a separate task, but I think the tooltip should not appear when the tooltipped item is being clicked on (maybe the `:active` CSS selector can be used)

Secondly, I think the tools that have selectable sub tools need a visual indicator that it does have subtools. Perhaps a tiny triangle in the bottom right of the icon area.

ollimeier:
> Generally looking good!
> 
> I wonder: when click-holding on the tool, the tooltip shows up, but maybe it shouldn't. Could be a separate task, but I think the tooltip should not appear when the tooltipped item is being clicked on (maybe the `:active` CSS selector can be used)
> 
> Secondly, I think the tools that have selectable sub tools need a visual indicator that it does have subtools. Perhaps a tiny triangle in the bottom right of the icon area.

The triangle bottom code may look like this:
.subtool-button:first-child:after{
  content: "";
  border-top: 5px solid transparent;
  border-right: 5px solid var(--foreground-color);
  margin-top: -3px;
  pointer-events: none;
}

src/fontra/views/editor/edit-tools-shape.js
ollimeier: Thanks for the suggestion. Changed it. Works well.

ollimeier:
> Generally looking good!
> 
> I wonder: when click-holding on the tool, the tooltip shows up, but maybe it shouldn't. Could be a separate task, but I think the tooltip should not appear when the tooltipped item is being clicked on (maybe the `:active` CSS selector can be used)
> 
> Secondly, I think the tools that have selectable sub tools need a visual indicator that it does have subtools. Perhaps a tiny triangle in the bottom right of the icon area.

I added working triangle. It was actually more complicated than thought, because it got in a conflict with the tooltip (triangle), why I had to find a slightly different solution. But it works now.

src/fontra/client/core/var-path.js
justvanrossum: I don't think this change is needed anymore.

src/fontra/views/editor/edit-tools-shape-ellipse.js
justvanrossum: That's not what I meant. The module edit-tools-shape-ellipse.js should not exist, and its contents should be part of edit-tools-shape.js.

src/fontra/views/editor/edit-tools-shape.js
```diff
@@ -0,0 +1,202 @@
...
+    // positon at center
+    if (event.altKey) {
+      x = x - width;
+      y = y - height;
+      width *= 2;
+      height *= 2;
+    }
+
+    return VarPackedPath.fromUnpackedContours(
+      this.getUnpackedContour(x, y, width, height)
```
justvanrossum, line 137:
```suggestion
      this.getUnpackedContours(x, y, width, height)
```
Should be plural: it returns an array of contours.

src/fontra/views/editor/edit-tools-shape.js
```diff
@@ -0,0 +1,202 @@
...
+    contour.points.reverse();
+    if (contour.isClosed) {
+      const [lastPoint] = contour.points.splice(-1, 1);
+      contour.points.splice(0, 0, lastPoint);
+    }
+    return contour;
+  }
+}
+
+export function getUnpackedContourRect(x, y, width, height, contourType = "cubic") {
```
justvanrossum, line 163: Same here, "contours" instead of "contour".

src/fontra/views/editor/editor.js
```diff
@@ -643,24 +664,74 @@ export class EditorController {
     });
   }
 
-  addEditTool(tool) {
+  addEditTool(tool, wrapper_id = "edit-tools") {
```
justvanrossum, line 667: Can you describe briefly what the `wrapper_id` does and why it is needed?

Separate from that: following Fontra naming conventions it should be called `wrapperID`.

src/fontra/views/editor/editor.js
```diff
@@ -783,13 +854,13 @@ export class EditorController {
     };
   }
 
-  setSelectedTool(toolIdentifier) {
+  setSelectedTool(toolIdentifier, wrapper_id = "unkown") {
```
justvanrossum, line 857: Similarly, can you describe why this argument is needed here? This is a "public" method call, the caller should not need to be aware of any wrapping.

src/fontra/views/editor/edit-tools-shape.js
```diff
@@ -0,0 +1,202 @@
...
+          if (!point.type) {
+            selection.add(`point/${firstIndex + index}`);
+          }
+        }
+        this.sceneController.selection = selection;
+
+        for (const [layerName, layerGlyph] of Object.entries(editLayerGlyphs)) {
+          layerGlyph.path.appendPath(pathNew);
+        }
+        return "add shape";
```
justvanrossum, line 98: Would be nice if this could be "rectangle", "square", "ellipse" or "circle".

Suggest to add a static attr to each shape class:
   `static shapeNames = ["rectangle", "ellipse"];`
And index by shift key:
```js
    return `add ${this.constructor.shapeNames[shiftKey ? 1 : 0]}`
```

src/fontra/views/editor/edit-tools-shape-ellipse.js
ollimeier: OK, did it.

src/fontra/views/editor/edit-tools-shape.js
ollimeier: Done.

src/fontra/views/editor/edit-tools-shape.js
ollimeier: done.

src/fontra/views/editor/editor.js
ollimeier: I wrap all sub tools into one div here, so that I can eg stack them visually and do further modifications like moving the child to first place when selected etc.
https://github.com/googlefonts/fontra/blob/0ca296368f68528b0603554413d5cce88b041749/src/fontra/views/editor/editor.js#L624

ollimeier: Renamed wrapper_id to wrapperID

src/fontra/views/editor/editor.js
ollimeier: Only the wrapper is 'selected' not the subtool, therefor `editToolItem.dataset.tool === toolIdentifier` does not work. This is my solution.

src/fontra/views/editor/edit-tools-shape.js
ollimeier: I am not 100% sure what you want. The following change will at least be a bit more precise and still be flexible if we would extend the tools:
https://github.com/googlefonts/fontra/blob/2426d9a003298f51295fb5596a94a05994c125aa/src/fontra/views/editor/edit-tools-shape.js#L98

ollimeier: I reverted this change, because it caused a fail in the tests.

src/fontra/client/core/var-path.js
ollimeier: Do you want me to remove it, or do we keep it to be in sync with 'Path2D'?

justvanrossum: It does not need to be in sync. It is not needed, so should be removed.

src/fontra/views/editor/edit-tools-shape.js
justvanrossum: (Sometimes the node tests glitch. The failure was not caused by this change.)

Sorry, my original comment was sloppy. I meant this (and forget about `static` and `.constructor`):

Rectangle tool:
```js
  shapeNames = ["rectangle", "square"];
```
Ellipse tool:
```js
  shapeNames = ["ellipse", "circle"];
```

And for the undo label:
```js
return `add ${this.shapeNames[shiftKey ? 1 : 0]}`
```

src/fontra/views/editor/editor.js
justvanrossum: You need to solve this differently. The method should only take the tool identifier.

For example, it could detect that it doesn't have a dataset.tool property, and check its childen.

justvanrossum: We may need to rethink a little bit. I think we need to distinguish between "select this tool or tool group" and "select this subtool". Perhaps the "tool group" needs an identifier, too. Let's chat about how to proceed.

This also affects the comand-<N> shortcuts, too: they currently don't visually work correctly for the shape tool, and I don't think the numbering should "dive into" the shape tool group. I think command-3 should choose the currently chosed shape tool, and 4 should select the ruler.

src/fontra/client/core/var-path.js
ollimeier: OK. Done.

src/fontra/views/editor/edit-tools-shape.js
ollimeier: OK, thanks. Done.

src/fontra/views/editor/editor.js
ollimeier: This sounds like a more complex topic. I don't know how to solve this currently.

src/fontra/views/editor/edit-tools-shape.js
```diff
@@ -5,7 +5,6 @@ import { BaseTool, shouldInitiateDrag } from "./edit-tools-base.js";
 import { registerVisualizationLayerDefinition } from "./visualization-layer-definitions.js";
 
 export class ShapeTool extends BaseTool {
-  iconPath = "/tabler-icons/square-plus-2.svg"; // placeholder, do we need it????
```
justvanrossum: Perhaps it doesn't even need to subclass BaseTool.

src/fontra/views/editor/editor.js
ollimeier: I make this as resolved as we did a coding session together.

src/fontra/views/editor/editor.js
ollimeier: I mark this as resolved as we did a coding session together.

src/fontra/views/editor/edit-tools-shape.js
justvanrossum: I removed the base class.

PR #1230, opened 2024-03-21
UI improvements for sidebearing value

### LINKED ISSUE #1133: [selection info] Sidebearing number fields should display rounded value (to one decimal)
Currently they show more digits than can fit:

image

It should be enough to show them with one decimal precision.

Entering with more precision should still be possible.

ollimeier:
Fixes #1133

Hi @justvanrossum,
I am kind of lost. This task should be very easy, but it turns out it's more complex then I thought. 
My first idea was, it should be possible to style the input number to display only one decimal place via CSS when it's active or not active. It turns out, this is not possible. Then I tried to find a javascript solution, please see the code.
Would be great to get you feedback before I spend/loose more time on it. Thanks

### DIFF
--- a/src/fontra/client/web-components/ui-form.js
+++ b/src/fontra/client/web-components/ui-form.js
@@ -180,10 +180,7 @@ export class Form extends SimpleElement {
     this._lastValidFieldValues[fieldItem.key] = fieldItem.value;
     const inputElement = document.createElement("input");
     inputElement.type = "number";
-    inputElement.value =
-      fieldItem.numDigits === undefined
-        ? fieldItem.value
-        : round(fieldItem.value, fieldItem.numDigits);
+    inputElement.value = maybeRound(fieldItem.value, fieldItem.numDigits);
 
     if ("minValue" in fieldItem) {
       inputElement.min = fieldItem.minValue;
@@ -234,8 +231,7 @@ export class Form extends SimpleElement {
     };
     this._fieldGetters[fieldItem.key] = () => inputElement.value;
     this._fieldSetters[fieldItem.key] = (value) =>
-      (inputElement.value =
-        fieldItem.numDigits === undefined ? value : round(value, fieldItem.numDigits));
+      (inputElement.value = maybeRound(value, fieldItem.numDigits));
     valueElement.appendChild(inputElement);
   }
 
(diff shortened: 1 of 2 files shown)


justvanrossum:
What are the event handlers for? What happens if all you change is this:
```
inputElement.value = fieldItem.value;
```
to this:
```
inputElement.value = getNiceNumber(fieldItem.value);
```

ollimeier:
@justvanrossum Here a short demo video how the difference look like:
https://github.com/googlefonts/fontra/assets/21055547/f7798387-f2a0-43c3-9b89-f4d70f769a36

justvanrossum:
Ok, now I understand what you're trying to do. Let me think. I'm not sure we need to go this far.

The `window.addEventListener()` is problematic as you keep adding more global event listeners without ever removing them.

Instead of getNiceNumber() please use `round(n, digits)` from utils.js.

There's a `round` missing in the setter (`this._fieldSetters[fieldItem.key] = (value) => ...`).

Let's (at least for now) get rid of the event handlers, and just do the two relevant places where inputElement.value is assigned to.

justvanrossum:
I changed my mind: if we can find a solution to the global event handler, I'm willing to go with the dynamic approach ("hover or edit and you'll see/edit all the digits").

ollimeier:
I'm no longer sure if it's even a good idea to change the whole approach. When I design a typeface, I want to see the actual value immediately and not just when I hover over the value. I have the feeling we are overthinking. If there is a value with many decimal places, I would like to see it, so I can change it.

ollimeier:
@justvanrossum Or what do you think?

src/fontra/client/web-components/ui-form.js
```diff
@@ -187,20 +180,9 @@ export class Form extends SimpleElement {
...
-    inputElement.onmouseover = (event) => {
-      inputElement.value = this._lastValidFieldValues[fieldItem.key];
-    };
-    inputElement.onclick = (event) => {
-      inputElement.value = this._lastValidFieldValues[fieldItem.key];
-    };
-    window.addEventListener("focusout", (event) => {
-      inputElement.value = getNiceNumber(this._lastValidFieldValues[fieldItem.key]);
-    });
+    inputElement.value = fieldItem.numDigits
```
justvanrossum, line 183: You need to explicitly compare with undefined, because zero is a valid input.

Perhaps write a helper function to avoid duplication with the setter.

PR #1247, opened 2024-04-03
Add Transformation Panel

ollimeier:
Fixes #339


src/fontra/views/editor/panel-selection-transformation.js
```diff
@@ -0,0 +1,542 @@
...
+      }
+    }
+
+    this.infoForm.setFieldDescriptions(formContents);
+    if (glyphController) {
+      await this._setupSelectionInfoHandlers(glyphName);
+    }
+  }
+
+  _getSelectedBounds(layerGlyph, pointIndices, componentIndices) {
```
ollimeier, line 344: @justvanrossum This is the context related to 'bounds of a layerGlyph component'

src/fontra/client/tabler-icons/arrows-maximize.svg
justvanrossum, line 1: It seems this image is not being used.

src/fontra/views/editor/edit-behavior.js
```diff
@@ -84,6 +84,7 @@ class EditBehavior {
       behavior
     );
     this.pointEditFuncs = pointEditFuncs;
+    this.participatingPointIndices = participatingPointIndices;
```
justvanrossum, line 87: Please don't edit this module until we've together established whether something should change.

src/fontra/views/editor/panel-selection-transformation.js
```diff
@@ -0,0 +1,534 @@
...
+      }
+      arrayPointsIndices.forEach((item) =>
+        newPointIndices.add(item + contours[i].startIndex)
+      );
+    }
+    return Array.from(newPointIndices).sort((a, b) => a - b);
+  }
+
+  async _transformLayerGlyph({
+    originX = this.originX,
```
justvanrossum, line 440: It seems you never set this value from the caller.

Also, I suggest to pass a Transform object instead of all separate parameters. To set the pinpoint for a transform, you can do something like this:
```
    t = Transform().translate(pinX, pinY)
    t = t.transform(theOtherTransform)
    t = t.translate(-pinX, -pinY)
```

src/fontra/views/editor/edit-behavior.js
ollimeier: Sure, any other idea how I get get the participatingPointIndices?

justvanrossum: You shouldn't need them: everything should be handled by EditBehavior itself. You just pass it the selection. See the arrow key nudge code, it should be very similar to that.

ollimeier: I tried to understand the arrow key nudge code, but it's very complex. The arrow key tool has a delta which is equal for every point, but in this case every point has a different delta, depending on the pinPoint and the transformation. It was way too complicated for me to figure out how to find a solution based on the arrow key tool / EditBehavior. I only need the effected point indices. Please don't get me wrong, I do my best to find a solution, but it's sometimes not easy. 
I may not understand the code enough to find the best solution. I am very sorry.

justvanrossum: Ok, let's work on it together tomorrow.

src/fontra/client/images/skew.svg
justvanrossum, line 1: Our "handmade" icons are maintained in a UFO in the icons/ folder, and exported with a script. Please add a "skew" glyph there, and make sure it fits in the same 1000x1000 box that the other icons use.

src/fontra/client/web-components/ui-form.js
```diff
@@ -78,10 +82,30 @@ export class Form extends SimpleElement {
...
+      white-space: nowrap;
+      margin-left: 1.3em;
+      margin-right: 1.3em;
+    }
+
+    .ui-form-icon.ui-form-icon-button {
+      display: inline-block;
+    }
+
+    .ui-form-center {
```
justvanrossum, line 103: Perhaps this could be integrated into the `origin-radio-buttons` CSS?

ollimeier: It's newly introduced because of the radio button icon, but it's more general for centering an element. Do you still want to move it to the panel-transformation?

src/fontra/views/editor/panel-transformation.js
```diff
@@ -0,0 +1,495 @@
...
+          ".ui-form-radio-button"
+        );
+        iconRadioButtons.forEach((radioButton) => {
+          radioButton.checked = false;
+        });
+      }
+    };
+  }
+
+  _getSelectedBounds(layerGlyphController, pointIndices, componentIndices) {
```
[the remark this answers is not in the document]
ollimeier: I created a new issue for a follow-up PR, why I am closing this as resolved.

src/fontra/client/web-components/ui-form.js
justvanrossum: Ok, I see, let's leave it here.

icons/fontra-icons.ufo/glyphs/skew.glif
```diff
@@ -0,0 +1,183 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<glyph name="skew" format="2">
+  <advance width="938"/>
```
justvanrossum, line 3: The width should be 1000. Please also check the vertical alignment compared to the other icons: this one sits a bit lower than the rest.

ollimeier:
@justvanrossum I double checked everything. From my side everything looks good.

icons/fontra-icons.ufo/glyphs/skew.glif
ollimeier: Interesting, I did set the width to 1000 at the beginning. I am wondering what caused the change? Anyway, thanks for finding it, let me fix it.

ollimeier: Fix it.

PR #1268, opened 2024-04-10
Issue 1264 refactor affine

### LINKED ISSUE #1264: Move/rename makeAffineTransform and decomposeAffineTransform to transform.js
- `makeAffineTransform()` -> `decomposedToTransform()`
- `decomposeAffineTransform()` -> `decomposedFromTransform()`

We don't have a `DecomposedTransform` class like fonttools has, but at least this would make those functions more similar to `DecomposedTransform.fromTransform()` and `DecomposedTransform().toTransform()`.

This depends on #1247.

ollimeier:
Fixes #1264

### DIFF
--- a/src/fontra/client/core/transform.js
+++ b/src/fontra/client/core/transform.js
@@ -170,7 +170,7 @@ export class Transform {
 
   toDecomposed() {
     //Decompose into a DecomposedTransform.
-    return DecomposedTransform.fromTransform(this);
+    return decomposedFromTransform(this);
   }
 
   toArray() {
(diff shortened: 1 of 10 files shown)


ollimeier:
@justvanrossum Even though I was not able to fix it, yet – I have some good findings: With the help of chai-almost I was able to reduce the number of fails and find a pattern. The big differences are always related to the 'yx'-values. I will investigate more tomorrow.

justvanrossum:
I committed two things:
- a minor fix that was done to the fonttools implementation (`>=` vs `>` in two places)
- Compute `delta` after the `sign()` call, this matches the fonttools implementation, and makes the tests pass

src/fontra/client/core/transform.js
```diff
@@ -167,6 +167,11 @@ export class Transform {
     return new this.constructor(xx, xy, yx, yy, dx, dy);
   }
 
+  toDecomposed() {
+    //Decompose into a DecomposedTransform.
```
justvanrossum, line 171:
```suggestion
    // Decompose into a DecomposedTransform.
```

justvanrossum (approved):
One nit, otherwise good to go!

PR #1271, opened 2024-04-11
Simplify decomposed transformation

justvanrossum:
Put the functionality of transforming a decomposed transform into a function. This simplifies client code, and additionally makes us not lose the tCenterX/Y when decomposing a nested composite.


justvanrossum:
(Not super happy with the `prependTransformToDecomposed()` name, but can't think of anything better right now.)

PR #1273, opened 2024-04-11
Adding method getSelectionBounds + unittest

### LINKED ISSUE #1266: getSelectionBounds for StaticGlyphController
I think we're still only using the on-curve points here.

Ideally, I think StaticGlyphController should get a method `getSelectionBounds(selection)` that would do the right thing given our "sparse" selection.

Maybe we should do this as a follow-up PR, and not let this improvement delay the landing of this PR.

_Originally posted by @justvanrossum in https://github.com/googlefonts/fontra/pull/1247#discussion_r1558998755_

ollimeier:
Fixes #1266

### DIFF
--- a/src/fontra/client/core/glyph-controller.js
+++ b/src/fontra/client/core/glyph-controller.js
@@ -631,11 +631,8 @@ export class StaticGlyphController {
     const selectionRects = [];
     if (pointIndices.length) {
       const path = filterPathByPointIndices(this.instance.path, pointIndices);
-      const selRect = rectFromPoints(
-        pointIndices.map((i) => path.getPoint(i)).filter((point) => !!point)
-      );
-      if (selRect) {
-        selectionRects.push(selRect);
+      if (path) {
+        selectionRects.push(path.getBounds());
       }
     }
 
--- a/src/fontra/client/core/glyph-controller.js
+++ b/src/fontra/client/core/glyph-controller.js
@@ -12,7 +12,7 @@ import {
 import { VariationError } from "./errors.js";
 import { filterPathByPointIndices } from "./path-functions.js";
 import { PathHitTester } from "./path-hit-tester.js";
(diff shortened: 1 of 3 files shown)


src/fontra/client/core/glyph-controller.js
```diff
@@ -619,6 +620,35 @@ export class StaticGlyphController {
...
+
+    for (const componentIndex of componentIndices) {
+      const component = this.components[componentIndex];
+      if (!component || !component.bounds) {
+        continue;
+      }
+      selectionRects.push(component.bounds);
+    }
+
+    if (selectionRects.length) {
```
justvanrossum, line 647: You can unconditionally do
```js
    return unionRect(...selectionRects);
```
As unionRect can handle an empty list.

src/fontra/views/editor/panel-transformation.js
```diff
@@ -316,38 +316,9 @@ export default class TransformationPanel extends Panel {
...
-    }
-  }
-
-  _getPinPoint(layerGlyphController, pointIndices, componentIndices, originX, originY) {
-    let bounds = this._getSelectedBounds(
-      layerGlyphController,
-      pointIndices,
-      componentIndices
+  _getPinPoint(layerGlyphController, originX, originY) {
+    let bounds = layerGlyphController.getSelectionBounds(
```
justvanrossum, line 320:
```suggestion
    const bounds = layerGlyphController.getSelectionBounds(
```

src/fontra/client/core/glyph-controller.js
ollimeier: done.

justvanrossum: With "unconditionally" I meant you can leave out the "if" statement. There's no need for the .length check.

ollimeier: Sorry. Changed it.

justvanrossum (approved):
Thank you!

PR #1297, opened 2024-04-19
[designspace] implement reading and writing of anchors

### LINKED ISSUE #1289: [designspace backend] Implement read and write for glyph anchors
Should include a small test case.

MutatorSans cap E contains an anchor, so is useful for testing.

ollimeier:
Fixes #1289

### DIFF
--- a/src/fontra/backends/designspace.py
+++ b/src/fontra/backends/designspace.py
@@ -1140,9 +1140,12 @@ def ufoLayerToStaticGlyph(glyphSet, glyphName, penClass=PackedPathPointPen):
     glyphSet.readGlyph(glyphName, glyph, pen, validate=False)
     components = [*pen.components] + unpackVariableComponents(glyph.lib)
     staticGlyph = StaticGlyph(
-        path=pen.getPath(), components=components, xAdvance=glyph.width
+        path=pen.getPath(),
+        components=components,
+        xAdvance=glyph.width,
+        anchors=glyph.anchors,
     )
-    # TODO: anchors
+
     # TODO: yAdvance, verticalOrigin
     return staticGlyph, glyph
 
--- a/src/fontra/backends/designspace.py
+++ b/src/fontra/backends/designspace.py
@@ -44,6 +44,7 @@ from ..core.classes import (
     StaticGlyph,
     VariableGlyph,
     structure,
+    unstructure,
 )
 from ..core.glyphdependencies import GlyphDependencies
 from ..core.path import PackedPathPointPen
@@ -1192,6 +1193,7 @@ def populateUFOLayerGlyph(
     layerGlyph.height = staticGlyph.yAdvance
     staticGlyph.path.drawPoints(pen)
     variableComponents = []
+    layerGlyph.anchors = [unstructure(a) for a in staticGlyph.anchors]
     for component in staticGlyph.components:
         if component.location or forceVariableComponents:
             # Store as a variable component
--- a/src/fontra/backends/designspace.py
+++ b/src/fontra/backends/designspace.py
@@ -27,7 +27,7 @@ from fontTools.pens.recordingPen import RecordingPointPen
 from fontTools.ufoLib import UFOReaderWriter
 from fontTools.ufoLib.glifLib import GlyphSet
 
-from ..core.classes import (  # unstructure, # TODO: unstructure doesn't wotk with Anchors, yet.
+from ..core.classes import (
     Anchor,
     Axes,
     AxisValueLabel,
@@ -43,7 +43,6 @@ from ..core.classes import (  # unstructure, # TODO: unstructure doesn't wotk wi
     MultipleAxisMapping,
     StaticGlyph,
     VariableGlyph,
-    structure,
 )
 from ..core.glyphdependencies import GlyphDependencies
 from ..core.path import PackedPathPointPen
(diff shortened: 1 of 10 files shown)


ollimeier:
@justvanrossum I have implemented the first version of _anchor support_ to the backend, but: `unstructure` doesn't work with Anchors, yet. I am wondering if it's related to `"customData": {}`? 
`fontra-copy` works :) I used it to get/create the example anchors for 'E' in our MutatorSans.fontra file.
I have also implemented a few unittests. Would be great if you would have time to give me some feedback on that if this goes into the right direction?

src/fontra/core/clipboard.py
```diff
@@ -64,7 +64,10 @@ def parseGLIF(data: str) -> StaticGlyph | None:
     except GlifLibError:
         return None
     return StaticGlyph(
-        path=pen.getPath(), components=pen.components, xAdvance=ufoGlyph.width
+        path=pen.getPath(),
+        components=pen.components,
+        xAdvance=ufoGlyph.width,
+        anchors=ufoGlyph.anchors,
```
justvanrossum, line 70: This can not work: the anchors need to be converted to Anchor objects. But let me think, it feels like we may need a better/more central way to convert .glif to StaticGlyph.

src/fontra/workflow/actions.py
```diff
@@ -224,6 +225,10 @@ def _scaleStaticGlyph(
             if glyph.verticalOrigin
             else glyph.verticalOrigin
         )
+        anchors = [
+            Anchor(name=a.name, x=a.x * self.scaleFactor, y=a.y * self.scaleFactor)
```
justvanrossum, line 229: This would drop any customData. The proper pattern is to use `replace()`, from the dataclasses module, like this:
```python
        anchors = [
            replace(a, x=a.x * self.scaleFactor, y=a.y * self.scaleFactor)
            for a in glyph.anchors
        ]
```
Which is also shorter as we don't need to set the name attr.

test-py/test_backends_designspace.py
```diff
@@ -191,6 +193,37 @@ async def test_addLocalAxis(writableTestFont):
...
 
+async def test_getAnchors(writableTestFont):
+    glyph = await writableTestFont.getGlyph("E")
+
+    layerName = "MutatorSansLightCondensed/foreground"
+    layer = glyph.layers[layerName]
+
+    assert 1 == len(layer.glyph.anchors)
+    assert (
+        structure({"name": "top", "x": 207, "y": 746}, Anchor) == layer.glyph.anchors[0]
```
justvanrossum, line 204: Better write 
```python
    Anchor(name="top", x=207, y=746) == ...
```

justvanrossum (commented):
Great start, thanks!

I noticed I didn't register `Anchor` as a class that may omit fields when they are set to their default values. Please insert
```python
registerHook(Anchor, customData=_unstructureDictSortedRecursively)
```
Right after this line in classes.py:
```python
registerHook(GlyphAxis, customData=_unstructureDictSortedRecursively)
```
This will likely make the tests fail again, on the E glyph in the .fontra package, where you'll find the empty "customData" item is now gone. Let me know if anything is unclear or doesn't behave as predicted.

src/fontra/core/clipboard.py
justvanrossum: For now, please just insert Anchor conversion code, just like you did in designspace.py. Those two bits of code are different enough that I don't immediately see a good way to refactor it to have more code in common.

src/fontra/backends/designspace.py
```diff
@@ -1139,10 +1142,14 @@ def ufoLayerToStaticGlyph(glyphSet, glyphName, penClass=PackedPathPointPen):
     pen = penClass()
     glyphSet.readGlyph(glyphName, glyph, pen, validate=False)
     components = [*pen.components] + unpackVariableComponents(glyph.lib)
+    anchors = [structure(a, Anchor) for a in glyph.anchors]
```
justvanrossum, line 1145: You know what, I take my advice to use "structure" here back. It's a bit of a hack, and could potentially break if UFO ever added more attributes to the anchor (unlikely, but still).
```suggestion
    anchors = [Anchor(name=a.name, x=a.x, y=a.y) for a in glyph.anchors]
```
It makes things more explicit, and show that the UFO anchor is _not_ an exact "unstructured" version of Fontra's Anchor object. As also evidenced by your findings regarding `unstructure`.

src/fontra/core/clipboard.py
ollimeier: Thanks. Done.

src/fontra/workflow/actions.py
ollimeier: Nice. Thanks a lot for the hint! Done.

test-py/test_backends_designspace.py
ollimeier: Done.

src/fontra/core/classes.py
```diff
@@ -218,7 +218,7 @@ class Component:
 
 @dataclass(kw_only=True)
 class Anchor:
-    name: str
+    name: Optional[str] = None
```
justvanrossum, line 221: I have a small request:
```suggestion
    name: Optional[str]
```
(Please apply manually locally, it will affect the tests.)

It turns out, having a default value for "name", but not for x and y, makes x and y appear _first_ in the unstructured dict. By removing the default (which forces us to supply a name on construction, but it may be None), the "name" attribute will appear first again, which is how I like it.

You will most likely have to update the E json file as a result of this change.

ollimeier: Agree. Name at the top is way better. Done.

src/fontra/backends/designspace.py
```diff
@@ -1139,10 +1141,14 @@ def ufoLayerToStaticGlyph(glyphSet, glyphName, penClass=PackedPathPointPen):
     pen = penClass()
     glyphSet.readGlyph(glyphName, glyph, pen, validate=False)
     components = [*pen.components] + unpackVariableComponents(glyph.lib)
+    anchors = [Anchor(name=a.get("name"), x=a["x"], y=a["y"]) for a in glyph.anchors]
```
justvanrossum, line 1144: Ok, now I have an idea to refactor this a tiny bit better, so we can share some code with the clipboard code:

Create a new helper function called `unpackAnchors(anchors)`, call it with `glyph.anchors`, and it should produce the list of `Anchor` instances. Place the new function after `unpackVariableComponents`. The clipboard code can import it, and we've saved a tiny bit of duplication.

PR #1300, opened 2024-04-23
Adding support for anchors

### LINKED ISSUE #908: Support anchors

ollimeier:
Fixes #908

### DIFF
--- a/src/fontra/client/core/glyph-controller.js
+++ b/src/fontra/client/core/glyph-controller.js
@@ -525,7 +525,6 @@ export class StaticGlyphController {
     this.errors = errors;
     this.canEdit = layerName != undefined;
     this.components = [];
-    this.anchors = instance.anchors || [];
   }
 
   async setupComponents(getGlyphFunc, parentLocation) {
@@ -572,6 +571,10 @@ export class StaticGlyphController {
 
   get verticalOrigin() {
     return this.instance.verticalOrigin;
+  }
+
+  get anchors() {
+    return this.instance.anchors;
   }
 
   get path() {
--- a/src/fontra/client/core/glyph-controller.js
+++ b/src/fontra/client/core/glyph-controller.js
@@ -525,7 +525,7 @@ export class StaticGlyphController {
     this.errors = errors;
     this.canEdit = layerName != undefined;
     this.components = [];
-    //this.anchors = instance.anchors;// || [];
+    this.anchors = instance.anchors || [];
   }
 
   async setupComponents(getGlyphFunc, parentLocation) {
--- a/src/fontra/client/core/glyph-controller.js
+++ b/src/fontra/client/core/glyph-controller.js
@@ -525,6 +525,7 @@ export class StaticGlyphController {
     this.errors = errors;
     this.canEdit = layerName != undefined;
     this.components = [];
+    //this.anchors = instance.anchors;// || [];
   }
 
   async setupComponents(getGlyphFunc, parentLocation) {
--- a/src/fontra/client/core/var-glyph.js
+++ b/src/fontra/client/core/var-glyph.js
@@ -56,7 +56,7 @@ export class StaticGlyph {
     }
     source.components =
       (noCopy ? obj.components : obj.components?.map(copyComponent)) || [];
-    //source.anchors = obj.anchors;
+    source.anchors = obj.anchors || [];
     return source;
   }
 
--- a/src/fontra/client/core/var-glyph.js
+++ b/src/fontra/client/core/var-glyph.js
@@ -56,6 +56,7 @@ export class StaticGlyph {
     }
     source.components =
       (noCopy ? obj.components : obj.components?.map(copyComponent)) || [];
+    //source.anchors = obj.anchors;
     return source;
   }
 
--- a/src/fontra/client/core/var-glyph.js
+++ b/src/fontra/client/core/var-glyph.js
@@ -56,7 +56,7 @@ export class StaticGlyph {
     }
     source.components =
       (noCopy ? obj.components : obj.components?.map(copyComponent)) || [];
-    source.anchors = noCopy ? obj.anchors : copyCustomData(obj.anchors || []);
+    source.anchors = noCopy ? obj.anchors || [] : copyCustomData(obj.anchors || []);
     return source;
   }
 
--- a/src/fontra/client/core/var-glyph.js
+++ b/src/fontra/client/core/var-glyph.js
@@ -56,7 +56,7 @@ export class StaticGlyph {
     }
     source.components =
       (noCopy ? obj.components : obj.components?.map(copyComponent)) || [];
-    source.anchors = obj.anchors || [];
+    source.anchors = noCopy ? obj.anchors : copyCustomData(obj.anchors || []);
     return source;
   }
 
--- a/src/fontra/views/editor/edit-behavior.js
+++ b/src/fontra/views/editor/edit-behavior.js
@@ -37,6 +37,7 @@ export class EditBehaviorFactory {
     this.contours = unpackContours(instance.path, pointSelection || []);
     this.components = unpackComponents(instance.components, relevantComponentIndices);
     this.anchors = unpackAnchors(instance.anchors, anchorSelection || []);
+    this.anchorSelection = anchorSelection || [];
     this.componentOriginIndices = componentOriginIndices || [];
     this.componentTCenterIndices = componentTCenterSelection || [];
     this.behaviors = {};
@@ -58,6 +59,7 @@ export class EditBehaviorFactory {
         this.contours,
         this.components,
         this.anchors,
+        this.anchorSelection,
         this.componentOriginIndices,
         this.componentTCenterIndices,
         behaviorType,
@@ -74,6 +76,7 @@ class EditBehavior {
     contours,
     components,
     anchors,
+    anchorSelection,
     componentOriginIndices,
     componentTCenterIndices,
     behavior,
@@ -119,9 +122,9 @@ class EditBehavior {
 
     const anchorRollbackChanges = [];
     this.anchorEditFuncs = [];
-    for (const [anchorIndex, anchor] of enumerate(anchors)) {
+    for (const anchorIndex of anchorSelection) {
       const [editFunc, anchorRollback] = makeAnchorEditFunc(
-        anchor,
+        anchors[anchorIndex],
         anchorIndex,
         this.roundFunc
       );
--- a/src/fontra/views/editor/edit-behavior.js
+++ b/src/fontra/views/editor/edit-behavior.js
@@ -270,20 +270,17 @@ function makeComponentOriginEditFunc(component, componentIndex, roundFunc) {
 }
 
 function makeAnchorEditFunc(anchor, anchorIndex, roundFunc) {
-  const origin = {
-    x: anchor.x,
-    y: anchor.y,
-  };
+  const oldAnchor = { ...anchor };
   return [
     (transform) => {
-      const editedOrigin = transform.constrained(origin);
+      const editedAnchor = transform.constrained(oldAnchor);
       return makeAnchorChange(
         anchorIndex,
-        roundFunc(editedOrigin.x),
-        roundFunc(editedOrigin.y)
+        roundFunc(editedAnchor.x),
+        roundFunc(editedAnchor.y)
       );
     },
-    makeAnchorChange(anchorIndex, origin.x, origin.y),
+    makeAnchorChange(anchorIndex, oldAnchor.x, oldAnchor.y),
   ];
 }
 
--- a/src/fontra/views/editor/edit-behavior.js
+++ b/src/fontra/views/editor/edit-behavior.js
@@ -417,7 +417,7 @@ function unpackComponents(components, selectedComponentIndices) {
 }
 
 function unpackAnchors(anchors, selectedAnchorIndices) {
-  const unpackedAnchors = new Array();
+  const unpackedAnchors = new Array(anchors.length);
   for (const anchorIndex of selectedAnchorIndices) {
     unpackedAnchors[anchorIndex] = anchors[anchorIndex];
   }
--- a/src/fontra/views/editor/edit-behavior.js
+++ b/src/fontra/views/editor/edit-behavior.js
@@ -1,7 +1,7 @@
 import { consolidateChanges } from "../core/changes.js";
 import { polygonIsConvex } from "../core/convex-hull.js";
 import { Transform, decomposedToTransform } from "../core/transform.js";
-import { parseSelection, reversed, unionIndexSets } from "../core/utils.js";
+import { enumerate, parseSelection, reversed, unionIndexSets } from "../core/utils.js";
 import { copyComponent } from "../core/var-glyph.js";
 import * as vector from "../core/vector.js";
 import {
@@ -21,6 +21,7 @@ export class EditBehaviorFactory {
     const {
       point: pointSelection,
       component: componentSelection,
+      anchor: anchorSelection,
       componentOrigin: componentOriginSelection,
       componentTCenter: componentTCenterSelection,
     } = parseSelection(selection);
@@ -35,6 +36,7 @@ export class EditBehaviorFactory {
     );
     this.contours = unpackContours(instance.path, pointSelection || []);
     this.components = unpackComponents(instance.components, relevantComponentIndices);
+    this.anchors = unpackAnchors(instance.anchors, anchorSelection || []);
     this.componentOriginIndices = componentOriginIndices || [];
     this.componentTCenterIndices = componentTCenterSelection || [];
     this.behaviors = {};
@@ -55,6 +57,7 @@ export class EditBehaviorFactory {
       behavior = new EditBehavior(
         this.contours,
         this.components,
+        this.anchors,
         this.componentOriginIndices,
         this.componentTCenterIndices,
         behaviorType,
@@ -70,6 +73,7 @@ class EditBehavior {
   constructor(
     contours,
     components,
+    anchors,
     componentOriginIndices,
     componentTCenterIndices,
     behavior,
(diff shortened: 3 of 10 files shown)


ollimeier:
@justvanrossum I am wondering how we want to handle anchors via selection? I am wondering if `parseSelection` should return `anchors` as well, so we can handle them separately:
```
    const {
      point: pointSelection,
      component: componentSelection,
      componentOrigin: componentOriginSelection,
      componentTCenter: componentTCenterSelection,
      anchors: anchorsSelection,
    } = parseSelection(selection);
```
Also, we may want to consider a different approach of how we select different objects. For example, when selecting all via short cut `⌘ + 'A'`, I would not like to have anchors selected (immediately). Maybe similar to how click selection works (multi steps) currently with outlines: one click -> one segment is selected, double click -> the whole contour is selected. 
In case of anchors: 
first `⌘ + 'A'` -> select all points + components
second `⌘ + 'A' `-> select all points + components + anchors
third `⌘ + 'A'` -> select all anchors

ollimeier:
> @justvanrossum I am wondering how we want to handle anchors via selection? I am wondering if `parseSelection` should return `anchors` as well, so we can handle them separately:
> 
> ```
>     const {
>       point: pointSelection,
>       component: componentSelection,
>       componentOrigin: componentOriginSelection,
>       componentTCenter: componentTCenterSelection,
>       anchors: anchorsSelection,
>     } = parseSelection(selection);
> ```
> 
> Also, we may want to consider a different approach of how we select different objects. For example, when selecting all via short cut `⌘ + 'A'`, I would not like to have anchors selected (immediately). Maybe similar to how click selection works (multi steps) currently with outlines: one click -> one segment is selected, double click -> the whole contour is selected. In case of anchors: first `⌘ + 'A'` -> select all points + components second `⌘ + 'A' `-> select all points + components + anchors third `⌘ + 'A'` -> select all anchors

Or with a different short cut `⌘ + alt + A`  (+ alternative via context menu `Select all anchors`)

src/fontra/client/core/glyph-controller.js
```diff
@@ -525,6 +525,7 @@ export class StaticGlyphController {
     this.errors = errors;
     this.canEdit = layerName != undefined;
     this.components = [];
+    this.anchors = instance.anchors || [];
```
justvanrossum, line 528: I don't think this is needed.

src/fontra/client/core/var-glyph.js
```diff
@@ -56,6 +56,7 @@ export class StaticGlyph {
     }
     source.components =
       (noCopy ? obj.components : obj.components?.map(copyComponent)) || [];
+    source.anchors = obj.anchors || [];
```
justvanrossum, line 59: You need to deep copy the anchors (unless `noCopy` is true). You can use the `copyCustomData()` function for that, which is a little poorly named once we use it for anchors.

justvanrossum (commented):
Quick first round of feedback.

src/fontra/client/core/glyph-controller.js
ollimeier: It does not work without it. Either I do something wrong or it's needed.

src/fontra/client/core/var-glyph.js
ollimeier: Thanks. Done.

src/fontra/client/core/glyph-controller.js
justvanrossum: Ah, I see it. You need to define a getter:
```python
  get anchors() {
    return this.instance.anchors;
  }
```

justvanrossum:
Selection: good questions. We also need to see how they respond to rect-select. See also selectionAtPoint() and selectionAtRect().

parseSelection() does not need to be changed to be used like you suggest: it already parses any `xxxx/ddd` pattern.

src/fontra/client/core/glyph-controller.js
ollimeier: Thanks. Done.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -1083,6 +1210,19 @@ function* iterPointsByIndex(path, pointIndices) {
   }
 }
 
+function* iterAnchorsPointsByIndex(anchors, anchorIndices) {
+  if (!anchorIndices || !anchors.length) {
+    return;
+  }
+  for (const index of anchorIndices) {
```
justvanrossum, line 1217: Since anchors are compatible with points (have x and y properties), this can be simplified:
```js
  for (const index of anchorIndices) {
    yield anchors[index];
  }
```

src/fontra/views/editor/editor.js
```diff
@@ -1780,6 +1847,174 @@ export class EditorController {
...
+    if (!tempAnchor) {
+      return;
+    }
+
+    const newAnchor = {
+      name: tempAnchor.name ? tempAnchor.name : "anchorName",
+      // x will be specified individually for each layer
+      y: tempAnchor.y ? tempAnchor.y : Math.round(point.y),
+    };
+    const relativeScaleX =
```
justvanrossum, line 1866: This will cause problems when xAdvance is zero (not unusual for combining accents)

src/fontra/views/editor/editor.js
```diff
@@ -1780,6 +1847,174 @@ export class EditorController {
...
+        warnings.push("⚠️ The name must not be empty");
+      }
+      if (
+        !(
+          nameController.model.anchorName ||
+          nameController.model.anchorX ||
+          nameController.model.anchorY
+        )
+      ) {
+        warnings.push("⚠️ At least one element need to be specified");
```
justvanrossum, line 1907: I don't understand this warning. I think the "Edit" dialog should just be completely populated, there is no need for placeholder magic at all here.

src/fontra/views/editor/editor.js
```diff
@@ -1791,23 +2026,59 @@ export class EditorController {
...
+
+    if (selectObjects) {
       for (const [pointIndex, pointType] of enumerate(glyphPath.pointTypes)) {
         if ((pointType & VarPackedPath.POINT_TYPE_MASK) === VarPackedPath.ON_CURVE) {
           newSelection.add(`point/${pointIndex}`);
         }
       }
-
-      for (const [componentIndex] of glyphComponents.entries()) {
+      for (const [componentIndex] of enumerate(glyphComponents)) {
```
justvanrossum, line 2071: Ah, I see that previous use of `.entries()` is a little ugly. But using `enumerate()` and not using the second value is equally ugly. I think the best would be this:
```js
     for (const componentIndex of range(glyphComponents.length)) {
```

src/fontra/views/editor/editor.js
```diff
@@ -1791,23 +2026,59 @@ export class EditorController {
...
         newSelection.add(`component/${componentIndex}`);
       }
     }
 
+    if (selectAnchors) {
+      for (const [pointIndex, anchor] of enumerate(positionedGlyph.glyph.anchors)) {
```
justvanrossum, line 2077: Same here, just use `range()`. Also, it's `anchorIndex`, not `pointIndex`:
```js
      for (const anchorIndex of range(positionedGlyph.glyph.anchors.length)) {
```

justvanrossum (commented):
Good work! Mostly minor comments.

src/fontra/views/editor/editor.js
justvanrossum: When I double-click an anchor, I get this dialog:
image
This warning is only confusing.

justvanrossum: Btw if I double-click an anchor, and fill in any value, there is an error in the console that's coming from the server:
image

The server error suggests you are sending a value that is not a float or an int:
```
2024-04-26 12:27:31 fontra.core.remote ERROR    uncaught exception: ClassValidationError('While structuring Anchor', [AssertionError(), AssertionError()], <class 'fontra.core.classes.Anchor'>)
  + Exception Group Traceback (most recent call last):
  |   File "/Users/just/code/git/BlackFoundry/fontra/src/fontra/core/remote.py", line 105, in _performCall
  |     returnValue = await methodHandler(*arguments, connection=self)
  |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/Users/just/code/git/BlackFoundry/fontra/src/fontra/core/fonthandler.py", line 290, in editFinal
  |     await self.updateLocalDataAndWriteToBackend(finalChange, connection)
  |   File "/Users/just/code/git/BlackFoundry/fontra/src/fontra/core/fonthandler.py", line 320, in updateLocalDataAndWriteToBackend
  |     await self._updateLocalDataAndWriteToBackend(change, sourceConnection, False)
  |   File "/Users/just/code/git/BlackFoundry/fontra/src/fontra/core/fonthandler.py", line 337, in _updateLocalDataAndWriteToBackend
  |     applyChange(rootObject, change)
  |   File "/Users/just/code/git/BlackFoundry/fontra/src/fontra/core/changes.py", line 91, in applyChange
  |     _applyChange(subject, change)
  |   File "/Users/just/code/git/BlackFoundry/fontra/src/fontra/core/changes.py", line 113, in _applyChange
  |     changeFunc(subject, *args, itemCast=itemCast)
  |   File "/Users/just/code/git/BlackFoundry/fontra/src/fontra/core/changes.py", line 16, in setItem
  |     item = itemCast(item)
  |            ^^^^^^^^^^^^^^
  |   File "/Users/just/code/git/BlackFoundry/fontra/src/fontra/core/classes.py", line 443, in structure
  |     return _cattrsConverter.structure(obj, cls)
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/Users/just/code/git/BlackFoundry/fontra/venv/lib/python3.12/site-packages/cattrs/converters.py", line 332, in structure
  |     return self._structure_func.dispatch(cl)(obj, cl)
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "<cattrs generated structure fontra.core.classes.Anchor>", line 25, in structure_Anchor
  |     if errors: raise __c_cve('While structuring ' + 'Anchor', errors, __cl)
  |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | cattrs.errors.ClassValidationError: While structuring Anchor (2 sub-exceptions)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "<cattrs generated structure fontra.core.classes.Anchor>", line 10, in structure_Anchor
    |     res['x'] = __c_structure_x(o['x'], __c_type_x)
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/Users/just/code/git/BlackFoundry/fontra/src/fontra/core/classes.py", line 325, in _structureNumber
    |     assert isinstance(d, (float, int))
    | AssertionError
    | Structuring class Anchor @ attribute x
    +---------------- 2 ----------------
    | Traceback (most recent call last):
    |   File "<cattrs generated structure fontra.core.classes.Anchor>", line 15, in structure_Anchor
    |     res['y'] = __c_structure_y(o['y'], __c_type_y)
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/Users/just/code/git/BlackFoundry/fontra/src/fontra/core/classes.py", line 325, in _structureNumber
    |     assert isinstance(d, (float, int))
    | AssertionError
    | Structuring class Anchor @ attribute y
    +------------------------------------
```

justvanrossum: The same happens when I edit a value in the "Add Anchor" dialog.

src/fontra/views/editor/editor.js

src/fontra/views/editor/editor.js

src/fontra/views/editor/editor.js
ollimeier: Fixed it.

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: Thanks. Done.

src/fontra/views/editor/edit-behavior.js
```diff
@@ -35,6 +36,8 @@ export class EditBehaviorFactory {
     );
     this.contours = unpackContours(instance.path, pointSelection || []);
     this.components = unpackComponents(instance.components, relevantComponentIndices);
+    this.anchors = unpackAnchors(instance.anchors, anchorSelection || []);
+    this.anchorSelection = anchorSelection || [];
```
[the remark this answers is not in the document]
ollimeier: Thanks. Good point. Done.

src/fontra/views/editor/edit-behavior.js
```diff
@@ -113,10 +120,23 @@ class EditBehavior {
       }
     }
 
+    const anchorRollbackChanges = [];
+    this.anchorEditFuncs = [];
+    for (const anchorIndex of anchorSelection) {
```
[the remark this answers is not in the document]
ollimeier: Done.

src/fontra/views/editor/editor.js
ollimeier: I fixed the error message.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -1215,10 +1215,8 @@ function* iterAnchorsPointsByIndex(anchors, anchorIndices) {
     return;
   }
   for (const index of anchorIndices) {
-    const anchor = anchors[index];
-    const pt = { x: anchor.x, y: anchor.y };
-    if (pt) {
-      yield pt;
+    for (const index of anchorIndices) {
```
justvanrossum, line 1218: You have a double loop now.

src/fontra/views/editor/editor.js
justvanrossum: I don't like the fix: you still divide by zero where it makes no sense.

src/fontra/views/editor/editor.js
```diff
@@ -1780,6 +1845,175 @@ export class EditorController {
...
+  }
+
+  async doAddAnchor() {
+    const point = this.sceneController.selectedGlyphPoint(this.contextMenuPosition);
+    const { anchor: tempAnchor } = await this.doAddEditAnchorDialog(undefined, point);
+    if (!tempAnchor) {
+      return;
+    }
+
+    let newAnchor = {
```
justvanrossum, line 1859: This can still be a const.

src/fontra/views/editor/editor.js
```diff
@@ -1780,6 +1845,175 @@ export class EditorController {
...
+      name: tempAnchor.name ? tempAnchor.name : "anchorName",
+      x: tempAnchor.x ? tempAnchor.x : Math.round(point.x),
+      y: tempAnchor.y ? tempAnchor.y : Math.round(point.y),
+    };
+    const instance = this.sceneModel.getSelectedPositionedGlyph().glyph.instance;
+    let relativeScaleX;
```
justvanrossum, line 1865:
```js
   const relativeScaleX = layerGlyph.xAdvance ? relativeScaleX = point.x / instance.xAdvance : null;
```

(No need to round here, right?)

src/fontra/views/editor/editor.js
```diff
@@ -1780,6 +1845,175 @@ export class EditorController {
...
+    if (instance.xAdvance != 0) {
+      relativeScaleX = Math.round(point.x) / instance.xAdvance;
+    }
+
+    await this.sceneController.editLayersAndRecordChanges((layerGlyphs) => {
+      for (const layerGlyph of Object.values(layerGlyphs)) {
+        if (!tempAnchor.x && layerGlyph.xAdvance != 0) {
```
justvanrossum, line 1872:
```js
        if (relativeScaleX != null) {
          newAnchor.x = layerGlyph.xAdvance * relativeScaleX;
        }
```

(Not sure why you check `!tempAnchor.x`)

ollimeier: Thanks for pointing me to this. The idea was correct, but the solution was wrong. tempAnchor.x could be also 0, in that case it would be wrong. We still need to check if tempAnchor.x is a valid number, because if that is the case, then this values would be used and not the calculated one.
Thanks! I fixed it.

src/fontra/views/editor/editor.js
ollimeier: Right. I changed it based on your suggestion.

src/fontra/views/editor/editor.js
ollimeier: Right. Changed it.

ollimeier:
> The background visualization for anchors is too dominant. In multi edit mode the other anchors are darker than the main anchor:
> 
> image
> And in "view background" mode they are very close in terms of gray value:
> 
> image
> I think they should be lighter in both cases (but more so in the multi-edit case)

I changed the colors. Better now?

src/fontra/views/editor/editor.js
justvanrossum: I still don't understand the need for checking tempAnchor.x. I currently get an unrounded x value if I place a new anchor.

There is something very unclear about this code, and I cannot figure out its intention.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -984,20 +984,21 @@ registerVisualizationLayerDefinition({
   zIndex: 490,
   screenParameters: {
     strokeWidth: 1,
-    originMarkerRadius: 4,
+    anchorRadius: 4,
   },
-  colors: { color: "#AAA8" },
-  colorsDarkMode: { color: "#8888" },
+  colors: { color: "#AAA8", colorAnchor: "#00AAA820" },
```
justvanrossum: I think you are also changing the hue here. I would expect the anchorColor to be a variant of #AAA8, say #AAA4 (ie. just changing the opacity). Same for dark mode.

(Note that #AAA8 is a shortcut for #AAAAAA88)

ollimeier: Thanks for the explanation. I changed the colors. But it seems the blue in dark mode does not get any transparency and I don't know why. I tried several different values.

src/fontra/views/editor/editor.js
ollimeier: I fixed the unrounded x.

src/fontra/views/editor/editor.js
```diff
@@ -1418,11 +1456,14 @@ export class EditorController {
           };
     }
 
-    const { point: pointIndices, component: componentIndices } = parseSelection(
-      this.sceneController.selection
-    );
+    const {
+      point: pointIndices,
+      component: componentIndices,
+      anchor: anchorSelection,
```
justvanrossum, line 1462: Please rename this variable to `anchorIndices`, in symmetry with the `componentIndices` right above it.

justvanrossum (commented):
This may be the last nit before we can merge! Please review the small changes I pushed friday evening.

PR #1307, opened 2024-04-26
spacebar down for hand tool + cleanSceneView

### LINKED ISSUE #121: [feature] pressing space bar (fill mode) also activates the hand tool for panning
Hey, I just tried this for the first time and I love it already!

I have a UI/UX suggestion. It'd be nice if, while pressing the space bar and in black-on-white fill mode, the hand tool were also activated such that one can pan around by dragging the cursor.
I think also other font editors do this (I forgot if it's Glyphs.app or Robofont or both, don't have a mac right now to confirm).

Anyway, thanks for considering.

ollimeier:
Fixes #121

### DIFF
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -611,6 +611,7 @@ export class EditorController {
   }
 
   initTools() {
+    this.toolIdentifierSelected = undefined;
     this.tools = {};
     this.topLevelTools = {};
     const editToolClasses = [PointerTool, PenTool, ShapeTool, PowerRulerTool, HandTool];
@@ -873,7 +874,10 @@ export class EditorController {
     };
   }
 
-  setSelectedTool(toolIdentifier) {
+  setSelectedTool(toolIdentifier, setToolIdentifier = true) {
+    if (setToolIdentifier) {
+      this.toolIdentifierSelected = toolIdentifier;
+    }
     let selectedToolIdentifier = toolIdentifier;
 
     for (const editToolItem of document.querySelectorAll(
@@ -1938,6 +1942,7 @@ export class EditorController {
     for (const overlay of document.querySelectorAll(".cleanable-overlay")) {
       overlay.classList.add("overlay-layer-hidden");
     }
+    this.setSelectedTool("hand-tool", false);
   }
 
   spaceKeyUpHandler(event) {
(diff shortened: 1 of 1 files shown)


src/fontra/views/editor/editor.js
```diff
@@ -611,6 +611,7 @@ export class EditorController {
   }
 
   initTools() {
+    this.toolIdentifierSelected = undefined;
```
justvanrossum, line 614: Please call this `selectedToolIdentifier`.

src/fontra/views/editor/editor.js
```diff
@@ -873,7 +874,10 @@ export class EditorController {
     };
   }
 
-  setSelectedTool(toolIdentifier) {
+  setSelectedTool(toolIdentifier, setToolIdentifier = true) {
```
justvanrossum, line 877: This new argument is not needed if we rearrange the logic a little bit.

I think the space-down and space-up code should be themselves be responsible for remembering which tool was selected. I will comment more below.

In the meantime, `this.selectedToolIdentifier` should be set at the end of this method, right after `this.sceneController.setSelectedTool(this.tools[selectedToolIdentifier]);`, and it should use the `selectedToolIdentifier` variable.

src/fontra/views/editor/editor.js
```diff
@@ -1938,6 +1942,7 @@ export class EditorController {
     for (const overlay of document.querySelectorAll(".cleanable-overlay")) {
       overlay.classList.add("overlay-layer-hidden");
     }
+    this.setSelectedTool("hand-tool", false);
```
justvanrossum, line 1945: Then, here we should remember the currently selected tool, so we can restore it in the space-up code.

I suggest this:
```js
      this.savedSelectedToolIdentifier = this.selectedToolIdentifier;
      this.setSelectedTool("hand-tool");
```

src/fontra/views/editor/editor.js
```diff
@@ -1946,6 +1951,7 @@ export class EditorController {
     for (const overlay of document.querySelectorAll(".cleanable-overlay")) {
       overlay.classList.remove("overlay-layer-hidden");
     }
+    this.setSelectedTool(this.toolIdentifierSelected);
```
justvanrossum, line 1954: Then here, we do
```js
    this.setSelectedTool(this. savedSelectedToolIdentifier);
    delete this.savedSelectedToolIdentifier;
```

This in total is a bit more code, but the intention is much much clearer.

src/fontra/views/editor/editor.js

src/fontra/views/editor/editor.js
```diff
@@ -611,6 +611,7 @@ export class EditorController {
   }
 
   initTools() {
+    this.selectedToolIdentifier = undefined;
```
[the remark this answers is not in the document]
ollimeier: Thanks. Done.

PR #1314, opened 2024-04-29
Include anchors to transformation panel tools

### LINKED ISSUE #1313: Add anchor support for 'Align and Distribute Objects'

ollimeier:
Fixes #1313

### DIFF
--- a/src/fontra/client/core/glyph-controller.js
+++ b/src/fontra/client/core/glyph-controller.js
@@ -626,11 +626,15 @@ export class StaticGlyphController {
   }
 
   getSelectionBounds(selection) {
-    let { point: pointIndices, component: componentIndices } =
-      parseSelection(selection);
+    let {
+      point: pointIndices,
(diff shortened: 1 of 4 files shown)


src/fontra/client/core/glyph-controller.js
```diff
@@ -648,6 +652,16 @@ export class StaticGlyphController {
       selectionRects.push(component.bounds);
     }
 
+    for (const anchorIndex of anchorIndices) {
+      const anchor = this.instance.anchors[anchorIndex];
+      selectionRects.push({
```
justvanrossum, line 657: You could use `centeredRect()` with width and height set to zero.

justvanrossum (commented):
Looking good! I haven't played with it yet.

src/fontra/client/core/glyph-controller.js
ollimeier: Thanks. Done.

PR #1316, opened 2024-04-30
Fix #1303: first time clicking on the edge of a sidebar

### LINKED ISSUE #1303: First time clicking on the edge of a sidebar makes it change width in a weird way
For example, in the Reference Font panel, it will make the panel so wide as for all chars to fit.

This only happens the first time after loading a page. Somehow this first click does something weird.

hwoongkang:
In `initResizeGutter()` of `views/editor/sidebar.js`:
https://github.com/googlefonts/fontra/blob/a723e8dc0d1613b7ee78d3bb64af49a1a99aeb58/src/fontra/views/editor/sidebar.js#L135-L156

width is initialized as `undefined`, and gets determined in the `onpointermove` event listener.

However, "clicking" does not triggers `pointermove` event, but only `pointerup` event.

So line 156 needs change, as width `undefined` is applied to the `applyWidth(width, saveLocalStroage)` method.

However also in line 179-181, null check is needed. https://github.com/googlefonts/fontra/blob/a723e8dc0d1613b7ee78d3bb64af49a1a99aeb58/src/fontra/views/editor/sidebar.js#L179-L181

Therefore I think refactoring `applyWidth(width, saveLocalStroage)` itself to contain null check is better.

This fixes #1303.

### DIFF
--- a/src/fontra/views/editor/sidebar.js
+++ b/src/fontra/views/editor/sidebar.js
@@ -103,6 +103,9 @@ export class Sidebar {
   }
 
   applyWidth(width, saveLocalStorage = false) {
+    if (width === undefined) {
+      return;
+    }
     if (saveLocalStorage) {
       localStorage.setItem(`fontra-sidebar-width-${this.identifier}`, width);
     }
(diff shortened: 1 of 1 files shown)


justvanrossum (approved):
Thank you so much! I agree with your refactoring.

PR #1317, opened 2024-04-30
Glyph locking

### LINKED ISSUE #1305: [feature request] glyph locking
Eventually we want to be able to lock glyphs, and have per-user permissions about it, so that for example a project manager can lock a glyph.

For now, we want a more "cooperative" locking scheme: there will be a button/checkbox to change the locking state of a glyph, but anyone can change it. It will just serve as a small hurdle against accidental editing.

The UI (checkbox, in the form of an lock/unlock icon?) should probably be part of the glyph selection info panel. The locking state would be stored as an item in the glyph's customData dict.

ollimeier:
Fixes #1305

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,7 @@ The following list of features is not complete, but gives you a rough idea of wh
 - Built-in support for Non-Linear Interpolation
 - Remove overlap and other path operators
 - Transformation panel ✅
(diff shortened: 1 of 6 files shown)


ollimeier:
I am closing this, because it gos in a totally wrong direction.

PR #1321, opened 2024-04-30
Glyph locking

### LINKED ISSUE #1305: [feature request] glyph locking

### LINKED ISSUE #1330: canCut() and canDelete() should check for read-only mode
Especially for "select-mode", I think edit-mode might be covered (but it's worth checking).

No real damage can be done as FontHandler will still stop such edits (if I remember correctly...), but currently the UI makes it look like we can cut or delete. Reloading will reveal the cut/delete didn't really take place.

Related to #1321. Feel free to fix is as part of that PR while you're at it.

ollimeier:
fixes #1305
fixes #1330


src/fontra/views/editor/panel-selection-info.js
```diff
@@ -138,6 +138,11 @@ export default class SelectionInfoPanel extends Panel {
     this.haveInstance = !!instance;
 
     const selectedGlyphInfo = this.sceneController.sceneModel.getSelectedGlyphInfo();
+    const varGlyphController =
+      await this.sceneController.sceneModel.getSelectedVariableGlyphController();
+    const glyphLocked = varGlyphController
```
justvanrossum, line 143: I _think_ you could simplify to:
```js
    const glyphLocked = !!varGlyphController?.glyph.customData.locked;
```
ie. make sure you account for `varGlyphController` possibly being undefined, but once it isn't, customData is guaranteed to exists.

Btw. you should use a "fontra." prefix for the customData key. I suggest `fontra.glyph.locked`. We don't have a good place for constants yet, so for now just use a string.

src/fontra/views/editor/panel-selection-info.js
```diff
@@ -160,6 +164,21 @@ export default class SelectionInfoPanel extends Panel {
...
+        key: "glyphLocking",
+        type: "header",
+        label: "Glyph info",
+        auxiliaryElement: html.createDomElement("icon-button", {
+          "id": "glyphLocking",
+          "style": `width: 1.3em;`,
+          "src": glyphLocked
+            ? "/tabler-icons/lock.svg"
+            : "/tabler-icons/lock-open-2.svg",
+          "onclick": (event) => this._glyphLocking(varGlyphController.glyph),
```
justvanrossum, line 177: Suggest to call this method `_toggleGlyphLock`

ollimeier: Done.

src/fontra/views/editor/panel-selection-info.js
```diff
@@ -409,6 +428,43 @@ export default class SelectionInfoPanel extends Panel {
...
+
+      if (!result) {
+        // User cancelled
+        return;
+      }
+    }
+
+    await this.sceneController.editGlyphAndRecordChanges(
+      (glyph) => {
+        this.sceneController.selection = new Set();
```
justvanrossum, line 450: Why do you reset the selection? That is not needed. Selection and copying is (and should) still possible in a locked glyph.

src/fontra/views/editor/panel-selection-info.js
```diff
@@ -409,6 +428,42 @@ export default class SelectionInfoPanel extends Panel {
...
+        return;
+      }
+    }
+
+    await this.sceneController.editGlyphAndRecordChanges(
+      (glyph) => {
+        glyph.customData["fontra.glyph.locked"] =
+          !glyph.customData["fontra.glyph.locked"];
+
+        const iconElement =
```
justvanrossum, line 453: Please take this section out of the `editGlyphAndRecordChanges` section, and use the `varGlyph` variable you have available there.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -11,6 +11,11 @@ import {
   withSavedState,
 } from "/core/utils.js";
 
+// the following icon SVG path code is from https://tabler.io/icons
```
justvanrossum, line 14: Please move this section to just before the `registerVisualizationLayerDefinition` where you use it.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -172,6 +177,29 @@ registerVisualizationLayerDefinition({
...
 });
 
+registerVisualizationLayerDefinition({
+  identifier: "fontra.glyph.locking",
+  name: "Glyph locking",
+  selectionMode: "editing",
+  userSwitchable: true,
+  zIndex: 700,
+  screenParameters: {
+    iconSize: 15,
```
justvanrossum, line 187: I tried with 12, and for me that's large enough.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -172,6 +177,29 @@ registerVisualizationLayerDefinition({
   },
 });
 
+registerVisualizationLayerDefinition({
```
justvanrossum, line 180: I think we need two visualisation layers:
- one for edit mode, *not* userSwitchable, always on
- one for non-edit mode, userSwitchable, off by default ("Glyph lock icon for non-editing glyphs")

src/fontra/views/editor/panel-selection-info.js
```diff
@@ -160,6 +164,21 @@ export default class SelectionInfoPanel extends Panel {
 
     const formContents = [];
     if (glyphName) {
+      formContents.push({
+        key: "glyphLocking",
```
justvanrossum, line 168: Header items don't need a key.

src/fontra/views/editor/panel-selection-info.js
ollimeier: Done.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -172,6 +177,29 @@ registerVisualizationLayerDefinition({
...
+  userSwitchable: true,
+  zIndex: 700,
+  screenParameters: {
+    iconSize: 15,
+  },
+  colors: { strokeColor: "#000C" },
+  colorsDarkMode: { strokeColor: "#FFFC" },
+  draw: (context, positionedGlyph, parameters, model, controller) => {
+    if (!!positionedGlyph.varGlyph.glyph.customData["fontra.glyph.locked"]) {
+      const box = positionedGlyph.unpositionedBounds;
```
justvanrossum, line 193: Please use `positionedGlyph.glyph.controlBounds` instead, and check whether it is null/undefined, fall back to the baseline if so.

(The unpositioned bounds do some weird things for empty glyphs, which is confusing here.)

justvanrossum:
There's an event handler in panel-designspace-navigation.js that also needs to listen to the `glyphEditCannotEditLocked` event:

https://github.com/googlefonts/fontra/blob/7f18fbefea7f4c8fc482ee5227e9ebf695489fd3/src/fontra/views/editor/panel-designspace-navigation.js#L232-L237

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: done.

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: done.

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: done

src/fontra/views/editor/panel-selection-info.js
ollimeier: Sorry. I changed it.

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: done

ollimeier:
> There's an event handler in panel-designspace-navigation.js that also needs to listen to the `glyphEditCannotEditLocked` event:
> 
> https://github.com/googlefonts/fontra/blob/7f18fbefea7f4c8fc482ee5227e9ebf695489fd3/src/fontra/views/editor/panel-designspace-navigation.js#L232-L237

I am very sorry, but I don't fully understand what I should do here. When I do some modifications to the sources/layers of a locked glyph a message pups up already.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -172,6 +172,51 @@ registerVisualizationLayerDefinition({
...
+
+registerVisualizationLayerDefinition({
+  identifier: "fontra.glyph.locking",
+  name: "Glyph locking",
+  selectionMode: "editing",
+  zIndex: 700,
+  screenParameters: { iconSize: 12 },
+  colors: { strokeColor: "#000C" },
+  colorsDarkMode: { strokeColor: "#FFFC" },
+  draw: (context, positionedGlyph, parameters, model, controller) => {
```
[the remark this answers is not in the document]
ollimeier: Yes, that works. Do you want me to refactor this in other parts of the code as well eg. _drawEmptyGlyphLayer or _drawSelectedGlyphLayer?

justvanrossum:
> When I do some modifications to the sources/layers of a locked glyph a message pups up already.

But the UI still changes. For example, dbl-click the on/off control (first column in the source list) the UI changes _as if_ the dbl-click was succesful, despite the dialog saying that it wasn't. By reloading the UI it gets reverted to the correct state. (Ideally it would be better to prevent the dbl-click to begin with, but that requires deeper thinking of how to implement that.

> but I don't fully understand what I should do here.

The code that gets triggered by the "glyphEditCannotEditReadOnly" event, should _also_ be triggered by the "glyphEditCannotEditLocked" event. In other words, if you receive "glyphEditCannotEditLocked", _updateAxes and _updateSources should be called.

src/fontra/views/editor/visualization-layer-definitions.js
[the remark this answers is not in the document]
justvanrossum: Sure, that would be a nice cleanup, thanks.

src/fontra/views/editor/panel-selection-info.js
```diff
@@ -409,6 +427,40 @@ export default class SelectionInfoPanel extends Panel {
...
+    }
+
+    const iconElement = this.infoForm.shadowRoot.querySelectorAll("#glyphLocking")[0];
+    iconElement.src = varGlyph.customData["fontra.glyph.locked"]
+      ? "/tabler-icons/lock-open-2.svg"
+      : "/tabler-icons/lock.svg";
+
+    await this.sceneController.editGlyphAndRecordChanges(
+      (glyph) => {
+        glyph.customData["fontra.glyph.locked"] =
```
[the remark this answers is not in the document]
ollimeier: Done.

justvanrossum (commented):
Two issues:
- the lock drawing code errors when the glyph doesn't exist (enter `/someglyph` in the text entry field to see)
- It is possible to delete or cut a whole glyph in select-mode

test-py/test_backends_designspace.py
```diff
@@ -222,6 +222,28 @@ async def test_addAnchor(writableTestFont):
     )
 
 
+async def test_read_glyph_locked(writableTestFont):
```
justvanrossum, line 225: You can use the `testFont` fixture here instead: you are not writing to the font after all.

test-py/test_backends_designspace.py
```diff
@@ -222,6 +222,28 @@ async def test_addAnchor(writableTestFont):
...
+    glyphName = "space"
+    glyph = await writableTestFont.getGlyph(glyphName)
+
+    assert glyph.customData.get("fontra.glyph.locked") is True
+
+
+async def test_add_glyph_locked(writableTestFont):
```
justvanrossum, line 232: Maybe rename to `test_write_glyph_locked`

src/fontra/views/editor/editor.js
```diff
@@ -585,6 +589,14 @@ export class EditorController {
     );
   }
 
+  async showDialogGlyphEditCannotEditLocked(create = false) {
```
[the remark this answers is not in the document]
ollimeier: Good point. Changed it.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -172,6 +168,50 @@ registerVisualizationLayerDefinition({
   },
 });
 
+// the following icon SVG path code is from https://tabler.io/icons
```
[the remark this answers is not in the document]
ollimeier: done

test-py/test_backends_designspace.py
ollimeier: right. done.

test-py/test_backends_designspace.py
ollimeier: done

justvanrossum:
Feature request:
- In the selection info panel, always show the lock as "locked", if the font is in read-only mode
- _maybe_ also trigger the canvas lock visualisation if the font is in read-only mode?

justvanrossum (commented):
Sorry, forgot to post these yesterday.

src/fontra/views/editor/editor.js
```diff
@@ -1510,6 +1525,13 @@ export class EditorController {
   }
 
   canPaste() {
+    const glyphLocking =
```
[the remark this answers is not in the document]
ollimeier: done already.

src/fontra/views/editor/editor.js
```diff
@@ -1501,6 +1525,13 @@ export class EditorController {
   }
 
   canPaste() {
+    const glyphLocked =
+      this.sceneModel.getSelectedPositionedGlyph()?.varGlyph.glyph.customData?.[
```
justvanrossum, line 1529: Try to paste onto a non-existing glyph: `Cannot read properties of null (reading 'glyph')`

This fixes it:
```suggestion
      this.sceneModel.getSelectedPositionedGlyph()?.varGlyph?.glyph.customData?.[
```

src/fontra/views/editor/editor.js
```diff
@@ -1278,6 +1285,14 @@ export class EditorController {
   }
 
   async doCut(event = null) {
+    const glyphLocking =
```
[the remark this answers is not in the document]

src/fontra/views/editor/editor.js
```diff
@@ -1270,6 +1270,13 @@ export class EditorController {
   }
 
   canCut() {
+    const glyphLocking =
+      this.sceneModel.getSelectedPositionedGlyph()?.varGlyph.glyph.customData?.[
```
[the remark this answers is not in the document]

src/fontra/views/editor/editor.js
ollimeier: Fixed it in `isSelectedGlyphLocked`

src/fontra/views/editor/scene-model.js
```diff
@@ -103,6 +103,12 @@ export class SceneModel {
     return getSelectedGlyphName(this.selectedGlyph, this.glyphLines);
   }
 
+  isSelectedGlyphLocked() {
+    return this.getSelectedPositionedGlyph()?.varGlyph?.glyph.customData?.[
```
justvanrossum, line 107:
```suggestion
    return !!this.getSelectedPositionedGlyph()?.varGlyph?.glyph.customData?.[
```
To ensure the return value is a bool (!!undefined -> false)

src/fontra/views/editor/editor.js
```diff
@@ -1261,6 +1270,9 @@ export class EditorController {
   }
 
   canCut() {
+    if (this.sceneModel.isSelectedGlyphLocked() || this.fontController.readOnly) {
```
justvanrossum, line 1273: Suggest to switch the two conditions: the readOnly check is cheaper, and if you do it first, it will shortcut the entire expression if readOnly is true, so `isSelectedGlyphLocked()` won't even be called. (Also the other occurrances.)

ollimeier: Very good point, I have not thought of that. Thanks.

src/fontra/views/editor/panel-selection-info.js
```diff
@@ -160,6 +164,21 @@ export default class SelectionInfoPanel extends Panel {
...
+        label: "Glyph info",
+        auxiliaryElement: html.createDomElement("icon-button", {
+          "id": "glyphLocking",
+          "style": `width: 1.3em;`,
+          "src":
+            glyphLocked || this.fontController.readOnly
+              ? "/tabler-icons/lock.svg"
+              : "/tabler-icons/lock-open-2.svg",
+          "onclick": (event) => this._toggleGlyphLock(varGlyphController.glyph),
+          "data-tooltip": glyphLocked ? "Unlock glyph" : "Lock glyph",
```
justvanrossum, line 178: The tooltip should read "Glyph is read-only" in the `this.fontController.readOnly` case.

src/fontra/views/editor/scene-model.js
```diff
@@ -103,6 +103,12 @@ export class SceneModel {
     return getSelectedGlyphName(this.selectedGlyph, this.glyphLines);
   }
 
+  isSelectedGlyphLocked() {
+    return !!this.getSelectedPositionedGlyph()?.varGlyph?.glyph.customData?.[
```
justvanrossum, line 107: Just a tiny nit:
```suggestion
    return !!this.getSelectedPositionedGlyph()?.varGlyph?.glyph.customData.[
```
Once we have a glyph, customData is guaranteed to exist.

src/fontra/views/editor/panel-selection-info.js
ollimeier: done

PR #1342, opened 2024-05-07
Status color of glyphs

### LINKED ISSUE #926: Add function to show status colour of glyphs in canvas
It'd be great to have a function to toggle show/hide the status colour (of the default layer) of each glyph in the canvas, maybe as a background colour of each glyph, or a coloured bar at the bottom.

This helps to have an overview of the status of a specific set of glyphs that we input in the text box.

ollimeier:
Fixes #926

Here a screenshot:
Screenshot 2024-05-07 at 11 44 28

### DIFF
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -345,9 +345,10 @@ export function arrayExtend(thisArray, itemsArray) {
 }
 
 export function rgbaToCSS(rgba) {
-  const channels = rgba.map((channel) => Math.round(channel * 255));
-  if (channels[3] == 255) {
-    channels.pop();
+  const channels = rgba.slice(0, 3).map((channel) => Math.round(channel * 255));
+  const alpha = rgba.length === 4 ? rgba[3] : undefined;
+  if (alpha !== undefined && 0 <= alpha && alpha < 1) {
+    channels.push(alpha);
   }
   return `${channels.length === 4 ? "rgba" : "rgb"}(${channels.join(",")})`;
 }
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -353,22 +353,6 @@ export function rgbaToCSS(rgba) {
   return `${channels.length === 4 ? "rgba" : "rgb"}(${channels.join(",")})`;
 }
 
-// according to the documentation, the alpha value must be a value
-// between 0.0 and 1.0, why it's not possible to use rgbaToCSS
-// more: https://doc.qt.io/qt-5/qml-qtquick-context2d.html#fillStyle-prop
-export function rgbaToFillStyle(rgba) {
-  let alpha = undefined;
-  if (rgba.length === 4) {
-    alpha = rgba[3];
-    rgba.pop();
-  }
-  const channels = rgba.map((channel) => Math.round(channel * 255));
-  if (alpha !== undefined && alpha != 1) {
-    channels.push(alpha);
-  }
-  return `${channels.length === 4 ? "rgba" : "rgb"}(${channels.join(",")})`;
-}
-
 export function clamp(number, min, max) {
   return Math.max(Math.min(number, max), min);
 }
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -352,22 +352,6 @@ export function rgbaToCSS(rgba) {
   return `${channels.length === 4 ? "rgba" : "rgb"}(${channels.join(",")})`;
 }
 
-// according to the documentation, the alpha value must be a value
-// between 0.0 and 1.0, why it's not possible to use rgbaToCSS
-// more: https://doc.qt.io/qt-5/qml-qtquick-context2d.html#fillStyle-prop
-export function rgbaToFillStyle(rgba) {
-  let alpha = undefined;
-  if (rgba.length === 4) {
-    alpha = rgba[3];
-    rgba.pop();
-  }
-  const channels = rgba.map((channel) => Math.round(channel * 255));
-  if (alpha !== undefined && alpha != 1) {
-    channels.push(alpha);
-  }
-  return `${channels.length === 4 ? "rgba" : "rgb"}(${channels.join(",")})`;
-}
-
 export function clamp(number, min, max) {
   return Math.max(Math.min(number, max), min);
 }
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -352,6 +352,22 @@ export function rgbaToCSS(rgba) {
   return `${channels.length === 4 ? "rgba" : "rgb"}(${channels.join(",")})`;
 }
 
+// according to the documentation, the alpha value must be a value
+// between 0.0 and 1.0, why it's not possible to use rgbaToCSS
+// more: https://doc.qt.io/qt-5/qml-qtquick-context2d.html#fillStyle-prop
+export function rgbaToFillStyle(rgba) {
+  let alpha = undefined;
+  if (rgba.length === 4) {
+    alpha = rgba[3];
+    rgba.pop();
+  }
+  const channels = rgba.map((channel) => Math.round(channel * 255));
+  if (alpha !== undefined && alpha != 1) {
+    channels.push(alpha);
+  }
+  return `${channels.length === 4 ? "rgba" : "rgb"}(${channels.join(",")})`;
+}
+
 export function clamp(number, min, max) {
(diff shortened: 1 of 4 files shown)


justvanrossum (commented):
Thanks! Looks very good.

Possible refinements:
- reduce the opacity of the color while in edit mode, maybe 50%?
- while I don't think the thickness of the color bar should be fully fixed in screen pixels, I feel it is getting too thin at very small sizes, and too dominant when zooming in. Suggestion: define a minimum and maximum thickness in screen pixels, and clamp the 50 value between them. Something like this:
```js
  const thickness = clamp(50, parameters.minThickness, parameters.maxThickness);
```

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -1020,6 +1021,45 @@ registerVisualizationLayerDefinition({
...
+      ];
+    if (status === undefined) {
+      return;
+    }
+
+    context.fillStyle = rgbaToCSS(statusFieldDefinitions[status].color);
+    context.beginPath();
+    context.rect(
+      0,
+      (-150 * model.fontController.unitsPerEm) / 1000,
```
justvanrossum, line 1055:
```suggestion
      -0.150 * model.fontController.unitsPerEm,
```

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -1020,6 +1021,45 @@ registerVisualizationLayerDefinition({
...
+      positionedGlyph.glyph.xAdvance,
+      50
```
justvanrossum, line 1057:
```suggestion
      0.05 * model.fontController.unitsPerEm
```

justvanrossum (commented):
- Ane minor suggestion
- The value of "50" should also be related to the units per em (but see also previous review)

ollimeier:
Thanks a lot for your feedback, Just. 
I added a opacity for the status color. But to make it work I had to create a new function `rgbaToFillStyle`, because `rgbaToCSS` does not work with the alpha value for canvas 2d, please see the following documentation:
[https://doc.qt.io/qt-5/qml-qtquick-context2d.html#fillStyle-prop](https://doc.qt.io/qt-5/qml-qtquick-context2d.html#fillStyle-prop)
Or do you have a different idea how to solve this?

Also, to make it work with two different colors (with and without transparency) for editing and non-editing mode had to split it into two separate `registerVisualizationLayerDefinition` or is there a way to make it possible with one, which I am not aware of?

Beside that, I tried to implement `const thickness = clamp(50, parameters.minThickness, parameters.maxThickness);`, but when I added minThickness and maxThickness to the parameters, the values were scaled (related to the zoom), resulting in weird values/sizes for the status color box in the background.


I hope this all make somehow sense to you.
Here a short demo video of the modifications:

https://github.com/googlefonts/fontra/assets/21055547/1d635f46-2844-4a61-9e49-2a30d2a4235a

justvanrossum (commented):
> I added a opacity for the status color. But to make it work I had to create a new function rgbaToFillStyle, because rgbaToCSS does not work with the alpha value for canvas 2d, please see the following documentation:
https://doc.qt.io/qt-5/qml-qtquick-context2d.html#fillStyle-prop

It is not that rgbaToCSS() does not work with the alpha value for canvas 2d, it does not work with the alpha value, period. You have discovered a bug :) So instead of adding a new function, you should fix the bug in the existing function.

> Also, to make it work with two different colors (with and without transparency) for editing and non-editing mode had to split it into two separate registerVisualizationLayerDefinition or is there a way to make it possible with one, which I am not aware of?

Good point. While it is possible to compare positionedGlyph to model.getSelectedPositionedGlyph(), this is not ideal, and I have created https://github.com/googlefonts/fontra/pull/1345 to make this easier. Rebase to main, and you can use the positionedGlyph.isEditing flag.

> Beside that, I tried to implement const thickness = clamp(50, parameters.minThickness, parameters.maxThickness);, but when I added minThickness and maxThickness to the parameters, the values were scaled (related to the zoom), resulting in weird values/sizes for the status color box in the background.

Having them scaled was exactly my intention. I want min and max values that are in pixels ("let the status bar not be thinner than X pixels and not be thicker than Y pixels"). So, calculate a base thickness by doing 0.05 * unitsPerEm (*not* scaled to pixels, this is in font units), and clamp that value between the *scaled* min and max values. Suggested minThickness: 3; suggested maxThickness: 15. We can fine tune later. This is exactly what I described earlier, though, but I hope it is clearer now.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -1020,6 +1022,60 @@ registerVisualizationLayerDefinition({
...
+    positionedGlyph.varGlyph.sources[sourceIndex].customData[
+      "fontra.development.status"
+    ];
+  if (status === undefined) {
+    return;
+  }
+
+  const thickness = clamp(parameters.thickness, 20, 150);
+  let color = statusFieldDefinitions[status].color;
+  color[3] = parameters.opacity;
```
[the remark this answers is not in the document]
ollimeier: Thank you very much. I very much hope it's better now.

src/fontra/client/core/utils.js
```diff
@@ -352,6 +352,22 @@ export function rgbaToCSS(rgba) {
   return `${channels.length === 4 ? "rgba" : "rgb"}(${channels.join(",")})`;
 }
 
+// according to the documentation, the alpha value must be a value
+// between 0.0 and 1.0, why it's not possible to use rgbaToCSS
+// more: https://doc.qt.io/qt-5/qml-qtquick-context2d.html#fillStyle-prop
```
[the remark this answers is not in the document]
ollimeier: I reverted all my changes and reworked it based on your feedback. For example, this new function has been removed.

src/fontra/client/core/utils.js
```diff
@@ -345,9 +345,10 @@ export function arrayExtend(thisArray, itemsArray) {
 }
 
 export function rgbaToCSS(rgba) {
-  const channels = rgba.map((channel) => Math.round(channel * 255));
-  if (channels[3] == 255) {
-    channels.pop();
+  const channels = rgba.slice(0, 3).map((channel) => Math.round(channel * 255));
+  const alpha = rgba.length === 4 ? rgba[3] : undefined;
```
justvanrossum, line 349: In JS, you can just ask for out-of-bounds indices and get undefined:
```suggestion
  const alpha = rgba[3];
```

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -1020,6 +1022,55 @@ registerVisualizationLayerDefinition({
...
+      color[3] = color[3] * 0.5;
+    }
+
+    const thickness = clamp(
+      0.05 * model.fontController.unitsPerEm,
+      parameters.minThickness,
+      parameters.maxThickness
+    );
+    context.fillStyle = rgbaToCSS(color);
+    context.beginPath();
```
justvanrossum, line 1068: There is `context.fillRect()` as a shortcut for these three lines.

ollimeier:
I am closing this PR, because there are some weird conflicts. I will create a new one in a sec.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -1020,6 +1022,55 @@ registerVisualizationLayerDefinition({
...
+        "fontra.development.status"
+      ];
+    if (status === undefined) {
+      return;
+    }
+
+    const color = [...statusFieldDefinitions[status].color];
+    if (positionedGlyph.isEditing) {
+      // in editing mode make opacity 50%
+      color[3] = color[3] * 0.5;
```
justvanrossum, line 1059:
```suggestion
      color[3] = color[3] * 0.4;
```

justvanrossum (commented):
Some minor comments. Almost done!

src/fontra/client/core/utils.js
ollimeier: Thank you. I changed it. Please see #1346

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: Thank you. I changed it. Please see #1346

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: Thank you. I changed it. Please see #1346

PR #1343, opened 2024-05-07
Alternative styling for disabled range-slider

### LINKED ISSUE #1341: [avar2/range-slider] styling alternate: disabled and (vertically) small
We need a styling alternate of the range-slider:
- the range input needs to be disabled (the user can't move the thumb)
- the numeric field needs to be disabled
- It needs to be more compact vertically
- the font size in the numeric field can be smaller

I think it would be best to add this to the range-slider code as styling alternates, triggered by the presence of a CSS class. But I'm open to suggestions.

ollimeier:
Fixes #1341 

`const sliderDisabled = false; //true;` -> this is a hack to change the axis sliders to the disabled version

The following screenshot shows the original style of the range-sliders:
Screenshot 2024-05-07 at 16 20 47

The following screenshot shows a possible alternative styling for the disabled range-sliders (together with number and label):
Screenshot 2024-05-07 at 16 21 05


ollimeier:
And the same comparison for dark mode:
Default:
Screenshot 2024-05-07 at 16 35 28

Disabled:
Screenshot 2024-05-07 at 16 35 06

justvanrossum (commented):
Great start.

I would suggest to leave out the outline for the numeric value completely. 

Idea: the thumb could be an rounded rect, with the width the same as the diameter of the "normal" thumb, but the heigh smaller (a little "pill").

Ideally, if you could mock up alternating sliders: normal one, small/disabled one, etc., that would be great, as it would be close to the intended use case. In the end, we also need to make sure the small/disabled one visually "belongs" to the normal one above it, but that should be up to the layout that places the sliders.

ollimeier:
For demonstration I looped through all axis a second time and displayed them as 'disabled range-sliders'. I hope this give you a better impression. I like your feedback '_leave out the outline for the numeric value completely_' a lot. It's much better now:

Screenshot 2024-05-08 at 14 31 05
Screenshot 2024-05-08 at 14 31 14

justvanrossum:
Suggestion for the mockup: instead of looping through all axes, just duplicate each axis with a disabled version of itself.

ollimeier:
> Looking very good. I would like the vertical spacing to be quite a bit tighter, though.

Hi Just, I am not sure how to make the vertical spacing a bit tighter without changing the overall design of the axis in this panel. As you can see in the following screenshot there is a gap in between all rows. The gap is set to `gap: 0.4em;`. We can change this gap value, but it will have an effect to the general design of how the axis are displayed in this panel. Why I have not touched it yet.
Any ideas or suggestions are very welcome.

Screenshot 2024-05-09 at 11 24 46

justvanrossum (commented):
This done, let's clean it up and merge. Mostly just delete the testing code I think.

justvanrossum (approved):

PR #1344, opened 2024-05-08
Basic localization support

### LINKED ISSUE #1340: Localization support
Will Fontra consider supporting languages other than English?
Also, I just started testing localization support myself.
<https://github.com/Aaron-212/fontra/tree/localization-support>

Aaron-212:
Close #1340 

### Todos
- [x] add support to plural forms
- [x] localize strings that are used directly into html elements
- [x] localize strings that are used directly into html attributes
- [x] read localization data from different .json files instead of hardcoding it inside .js
- [x] a switch-language menu
- [x] changes should be applied without refreshing the page
      current workaround: use `location.reload(true)` to reload


Aaron-212:
Is there any way to reload the UI without refreshing the whole page?

justvanrossum:
> Is there any way to reload the UI without refreshing the whole page?

Not really. I'd need to look into editor.js, maybe it should get a method to that effect. The closes thing would be is to call `editorController.reloadData()`. Hmm, maybe that just works.

Else: what are the downsides of a page reload for what you are doing?

justvanrossum:
Random idea (to be considered for later, it may not be that relevant for this initial work): should we maintain the translations as a (google-docs or equivalent) spreadsheet, to make it easier to onboard people who can help with providing/fixing/checking translations?

Aaron-212:
`EditorController.reloadData()` breaks the whole thing 😂
image
But using `location.reload(true)` (full reload) also has some issues. Some localization data won't load after refreshing multiple times.
image
But for most times `location.reload(true)` can do the work.
image

Aaron-212:
image

Figured out an alternative way to call `editorController.reloadData()`, but most of the localization are not reloaded.

Aaron-212:
Rebased
But I still cannot figure out a proper way to use `editorController.reloadData()`😂
Guess I'll just stick to `window.location.reload(true)`

Aaron-212:
May I add a prettier plugin to sort json files?

justvanrossum:
About the json prettier plugin: that's fine as long as we have control about which files it will touch. Like, there are .json files in test-common that I'm not sure I want to be touched. There are .json files inside .fontra folder that should not be touched, but I think they are already ignored.

Aaron-212:
I'm planning to mark this PR as ready, since there're too many files modified. The main editor is almost localized, and for unlocalized dialogs and pop-ups I'll open a second PR  after this one's merged.

src/fontra/views/fontinfo/panel-axes.js
```diff
@@ -15,44 +15,45 @@ import { piecewiseLinearMap } from "../core/var-model.js";
 import { IconButton } from "../web-components/icon-button.js"; // for <icon-button>
 import { UIList } from "../web-components/ui-list.js";
 import { BaseInfoPanel } from "./panel-base.js";
+import { translate } from "/core/localization.js";
 import "/web-components/add-remove-buttons.js";
 import { dialogSetup } from "/web-components/modal-dialog.js";
 
 const presetAxes = [
   {
-    label: "Weight",
```
justvanrossum, line 23: I'm not sure what to do instead, but these are not correct: the "Weight" label is not (just) a UI label, but it is a "label" attribute for the axis to be created.

For example, when choosing optical size:

image

The last (editable) field should read "Optical Size", not "optical-size".

src/fontra/client/web-components/ui-form.js
```diff
@@ -163,9 +163,9 @@ export class Form extends SimpleElement {
       }
 
       let label = fieldItem.label || fieldItem.key || "";
-      if (label.length && fieldItem.type !== "header") {
+      /* if (label.length && fieldItem.type !== "header") {
         label += ":";
-      }
+      } */ // Conflicts with colons within localization values
```
justvanrossum, line 168: Can you elaborate on this problem?

src/fontra/views/editor/cjk-design-frame.js
```diff
@@ -2,8 +2,8 @@ import {
   registerVisualizationLayerDefinition,
   strokeLine,
 } from "./visualization-layer-definitions.js";
+import { translate } from "/core/localization.js";
 import { range } from "/core/utils.js";
-
```
justvanrossum, line 6: Please leave the blank line.

src/fontra/views/editor/panel-designspace-navigation.js
```diff
@@ -358,21 +359,25 @@ export default class DesignspaceNavigationPanel extends Panel {
...
       {
-        title: "on",
+        title: translate("sidebar.designspace-navigation.glyph-sources.on"),
         key: "active",
         cellFactory: makeIconCellFactory(
           ["/tabler-icons/circle-dotted.svg", "/tabler-icons/circle-dot.svg"],
           true
         ),
-        width: "1.2em",
+        width: "2.5em",
```
justvanrossum, line 368: Hm, this changes the layout more than I'd like. Perhaps we need an icon instead of "on" as a label, or leave out the label altogether.

src/fontra/views/editor/panel-user-settings.js
```diff
@@ -68,7 +69,22 @@ export default class UserSettingsPanel extends Panel {
     });
 
     items.push({
-      displayName: "Experimental features",
+      displayName: translate("sidebar.user-settings.language"),
+      controller: languageController,
+      descriptions: [
+        {
+          key: "language",
+          ui: "radio",
```
justvanrossum, line 77: I would prefer this to be a popup menu, but I guess ui-form doesn't support that yet. May be a follow-up refinement.

src/fontra/views/editor/panel-user-settings.js
```diff
@@ -68,7 +69,22 @@ export default class UserSettingsPanel extends Panel {
...
+          options: [
+            { key: "en", displayName: translate("language.en") },
```
justvanrossum, line 79: I would love for the language to also always be available in english, maybe even the section header as well. Use case: someone (like me) playing around with languages, and selecting one they don't master, and then can't find their way back :) Maybe add the enlish name in parenthesis after the chinese name? Simplest would be to add it to the translation strings, but perhaps it is better to enforce this in code. What do you think?

justvanrossum (commented):
Thank you so much for this huge amount of work, and for keeping it up to date!

I left some comments.

General question, if I find things that are not yet translated, would you like me to tell you, or are you aware? For example, I found the dialogs that use "_sourcePropertiesRunDialog" to be untranslated.

justvanrossum:
Please use pre-commit, you may have to run "pre-commit install" to turn it on for this repo.

I notice in the test fail that it reformats some .json files that are unrelated to this PR. I would prefer them to not be changed. Maybe you can activate the json prettier only for the files you are interested in?

src/fontra/client/web-components/ui-form.js
Aaron-212: The main issue here is the difference of latin colon `:` and CJK colon `：`, they are different.
I'd like to use the correct colon for the corresponding language, thus I comment out this function to avoid appending a latin colon after Chinese phrases.
We can hardcode the colons in the language files instead.

justvanrossum: Thanks, understood!

src/fontra/views/fontinfo/panel-axes.js
```diff
@@ -120,7 +125,7 @@ export class AxesPanel extends BaseInfoPanel {
     const selected = "wght";
 
     const controller = new ObservableController({ ...presetAxesByTag[selected] });
-    controller.addKeyListener(["name", "tag", "label"], (event) => {
+    controller.addKeyListener(["name", "tag", "label", "loclKey"], (event) => {
```
justvanrossum, line 128: This change is not needed, as the "loclKey" doesn't map to an editable field.

src/fontra/views/fontinfo/panel-axes.js
```diff
@@ -143,9 +148,7 @@ export class AxesPanel extends BaseInfoPanel {
           },
         }),
         html.label({ for: identifier }, [
-          `${translate("axes.preset." + presetAxis.label)} (${presetAxis.name}, ${
-            presetAxis.tag
-          })`,
+          `${translate(presetAxis.loclKey)} (${presetAxis.name}, ${presetAxis.tag})`,
```
justvanrossum, line 151: Suggest to add presetAxis.label to the parenthesis list:
```js
     `${translate(presetAxis.loclKey)} (${presetAxis.label}, ${presetAxis.name}, ${presetAxis.tag})`,
```

justvanrossum:
I see that if a translation is not found, the key is returned.

I'm worried about new things that will be added, which will be in English first. Imagining we may have more translations in the future, I wonder if we should fall back to english if a translation isn't found. Else we'd be forced to always have all new translations ready at the same time, which seems impractical.

Aaron-212:
IMO this isn't a big problem right now, since we can still hardcode the English strings in the source code for new strings. I'll add the fallback strategy in the next PR, so that new strings can use `translate(key)` directly.

justvanrossum:
Thanks so much!

PR #1348, opened 2024-05-09
Fix #1315: decompose component should include anchors

### LINKED ISSUE #1315: Decompose component: include anchors
TODO: 
Include anchors when decomposing a component. 
(This is not the case currently.)

hwoongkang:
First, I tried to match the style with previous component and path decomposing codes.

The case I paid extra attention is when the glyph and its component have anchors with same names;
For example, if somehow the parent glyph had a `top` anchor, and a compoment has a `top` anchor, only one of either anchors should remain, not both.

I chose to preserve anchors from the parent glyph. 

And as I was reading the code, decomposing supports decomposing multiple components at the same time. In that case, anchors from a component with a preceding order have priority.

This fixes #1315

### DIFF
--- a/src/fontra/client/core/glyph-controller.js
+++ b/src/fontra/client/core/glyph-controller.js
@@ -923,12 +923,11 @@ export async function decomposeComponents(
       });
     }
     for (const anchor of compoInstance.anchors) {
-      const { name, x, y } = anchor;
-      const [newX, newY] = t.transformPoint(x, y);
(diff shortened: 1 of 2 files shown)


ollimeier:
@justvanrossum I tested it and it works like expected – anchors with same names are handled correct from my point of view. Also the code looks good to me. 
I would say, great job :)

justvanrossum:
Thank you very much! Good work indeed.

PR #1354, opened 2024-05-10
Adding Guidelines

### LINKED ISSUE #909: Support guidelines
- global
- local
- specific to a group of glyphs?

ollimeier:
It fixes the issue 909 partly (global guidelines are not done, yet)

Hi @justvanrossum, 
here is **draft** PR – it's a starting point. I have the following questions/issues:

The good thing first: the visualization of the local guidelines works already. Even selecting and moving them around. But editing and adding causes an error currently and I was not able to fix it, yet. 
Here is a short demo video:

https://github.com/googlefonts/fontra/assets/21055547/0de5acdd-ba1f-445b-b05a-cda6ed83054b

Do you have an idea how to fix it? 

Also: I am am not 100% sure how we want to handle the different guidelines: for example in the selection I first started with `guideline/index`, similar to `component/index` or `anchor/index`, but during coding I realized that this way we cannot differ between local and global guidelines. Why I named it `guidelineLocal/index`.

Thanks a lot in advance, 
Olli

### DIFF
--- a/src/fontra/backends/designspace.py
+++ b/src/fontra/backends/designspace.py
@@ -1219,9 +1219,9 @@ def unpackGuidelines(guidelines):
         (
             Guideline(
                 name=g.get("name"),
-                x=g["x"],
-                y=g["y"],
-                angle=g["angle"],
+                x=g.get("x", 0),
+                y=g.get("y", 0),
+                angle=g.get("angle", 0),
                 locked=g.get("locked", False),
                 # TODO: Guidelines, how do we handle customData like:
                 # color=g.get("color"),
@@ -1229,7 +1229,10 @@ def unpackGuidelines(guidelines):
             )
             if g.get("name")
             else Guideline(
-                x=g["x"], y=g["y"], angle=g["angle"], locked=g.get("locked", False)
+                x=g.get("x", 0),
+                y=g.get("y", 0),
+                angle=g.get("angle", 0),
+                locked=g.get("locked", False),
             )
         )
         for g in guidelines
--- a/src/fontra/backends/designspace.py
+++ b/src/fontra/backends/designspace.py
@@ -1024,6 +1024,7 @@ def unpackDSSource(dsSource: DSSource, unitsPerEm: int) -> FontSource:
         name=dsSource.name,
         location=dsSource.location,
         verticalMetrics=verticalMetrics,
+        guidelines=unpackGuidelines(fontInfo.guidelines),
     )
 
 
@@ -1082,6 +1083,7 @@ class UFOGlyph:
 
 class UFOFontInfo:
     unitsPerEm = 1000
+    guidelines: list = []
 
 
 class UFOManager:
--- a/src/fontra/backends/designspace.py
+++ b/src/fontra/backends/designspace.py
@@ -1216,14 +1216,18 @@ def unpackAnchors(anchors):
 
 def unpackGuidelines(guidelines):
     return [
-        Guideline(
-            name=g.get("name"),
-            x=g["x"],
-            y=g["y"],
-            angle=g["angle"],
-            # TODO: Guidelines, how do we handle customData like:
-            # color=g.get("color"),
-            # identifier=g.get("color"),
+        (
+            Guideline(
+                name=g.get("name"),
+                x=g["x"],
+                y=g["y"],
+                angle=g["angle"],
+                # TODO: Guidelines, how do we handle customData like:
+                # color=g.get("color"),
+                # identifier=g.get("color"),
+            )
+            if g.get("name")
+            else Guideline(x=g["x"], y=g["y"], angle=g["angle"])
         )
         for g in guidelines
     ]
@@ -1258,7 +1262,11 @@ def populateUFOLayerGlyph(
         {"name": a.name, "x": a.x, "y": a.y} for a in staticGlyph.anchors
     ]
     layerGlyph.guidelines = [
-        {"name": g.name, "x": g.x, "y": g.y, "angle": g.angle}
+        (
+            {"name": g.name, "x": g.x, "y": g.y, "angle": g.angle}
+            if g.name
+            else {"x": g.x, "y": g.y, "angle": g.angle}
+        )
         for g in staticGlyph.guidelines
     ]
     for component in staticGlyph.components:
--- a/src/fontra/backends/designspace.py
+++ b/src/fontra/backends/designspace.py
@@ -1222,12 +1222,15 @@ def unpackGuidelines(guidelines):
                 x=g["x"],
                 y=g["y"],
                 angle=g["angle"],
+                locked=g.get("locked", False),
                 # TODO: Guidelines, how do we handle customData like:
                 # color=g.get("color"),
-                # identifier=g.get("color"),
+                # identifier=g.get("identifier"),
             )
             if g.get("name")
-            else Guideline(x=g["x"], y=g["y"], angle=g["angle"])
+            else Guideline(
+                x=g["x"], y=g["y"], angle=g["angle"], locked=g.get("locked", False)
+            )
         )
         for g in guidelines
     ]
--- a/src/fontra/backends/designspace.py
+++ b/src/fontra/backends/designspace.py
@@ -1216,24 +1216,15 @@ def unpackAnchors(anchors):
 
 def unpackGuidelines(guidelines):
     return [
-        (
-            Guideline(
-                name=g.get("name"),
-                x=g.get("x", 0),
-                y=g.get("y", 0),
-                angle=g.get("angle", 0),
-                locked=g.get("locked", False),
-                # TODO: Guidelines, how do we handle customData like:
-                # color=g.get("color"),
-                # identifier=g.get("identifier"),
-            )
-            if g.get("name")
-            else Guideline(
-                x=g.get("x", 0),
-                y=g.get("y", 0),
-                angle=g.get("angle", 0),
-                locked=g.get("locked", False),
-            )
+        Guideline(
+            name=g.get("name"),
+            x=g.get("x", 0),
+            y=g.get("y", 0),
+            angle=g.get("angle", 0),
+            locked=g.get("locked", False),
+            # TODO: Guidelines, how do we handle customData like:
+            # color=g.get("color"),
+            # identifier=g.get("identifier"),
         )
         for g in guidelines
     ]
@@ -1268,11 +1259,7 @@ def populateUFOLayerGlyph(
         {"name": a.name, "x": a.x, "y": a.y} for a in staticGlyph.anchors
     ]
     layerGlyph.guidelines = [
-        (
-            {"name": g.name, "x": g.x, "y": g.y, "angle": g.angle}
-            if g.name
-            else {"x": g.x, "y": g.y, "angle": g.angle}
-        )
+        {"name": g.name, "x": g.x, "y": g.y, "angle": g.angle}
         for g in staticGlyph.guidelines
     ]
     for component in staticGlyph.components:
--- a/src/fontra/client/core/ui-utils.js
+++ b/src/fontra/client/core/ui-utils.js
@@ -94,6 +94,27 @@ function didReorder(a, b) {
   return false;
 }
 
+export function labeledCheckbox(label, controller, key, options) {
+  const checkboxID = options?.id || `checkbox-${uniqueID()}-${key}`;
+  const items = [html.div()];
+  const inputWrapper = html.div();
+  const inputElement = html.input({ type: "checkbox", id: checkboxID });
+  inputElement.checked = controller.model[key];
+  inputWrapper.appendChild(inputElement);
+  inputWrapper.appendChild(html.label({ for: checkboxID }, [label]));
+  items.push(inputWrapper);
+
+  inputElement.onchange = () => {
+    controller.model[key] = inputElement.checked;
+  };
+
+  controller.addKeyListener(key, (event) => {
+    inputElement.checked = event.newValue;
+  });
+
+  return items;
+}
+
 export function labeledTextInput(label, controller, key, options) {
   options = { continuous: true, ...options };
   const items = [];
@@ -114,13 +135,8 @@ export function labeledTextInput(label, controller, key, options) {
     inputElement.className = options.class;
   }
   inputElement.value = formatter.toString(controller.model[key]);
-  if (inputElement.type === "checkbox") {
-    inputElement.checked = controller.model[key];
-  }
   inputElement[options.continuous ? "oninput" : "onchange"] = () => {
-    const inputValue =
-      inputElement.type === "checkbox" ? inputElement.checked : inputElement.value;
-    const { value, error } = formatter.fromString(inputValue);
+    const { value, error } = formatter.fromString(inputElement.value);
     if (!error) {
       controller.model[key] = value;
     }
@@ -163,18 +179,6 @@ export const DefaultFormatter = {
     return {
       value: value,
     };
-  },
-};
-
-export const BooleanFormatter = {
-  toString: (value) => (value != undefined ? value.toString() : ""),
-  fromString: (value) => {
-    const boolean = Boolean(value);
-    if (boolean !== true && boolean !== false) {
-      return { error: "not a boolean" };
-    } else {
-      return { value: boolean };
-    }
   },
 };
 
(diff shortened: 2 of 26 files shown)


ollimeier:
@justvanrossum I guess I am almost done with the glyph guidelines (there are still some ideas/issues to do like 'locking a guideline' or allow adding 'guidelines without names'), but the overall functionality works for the glyph level guidelines. Now I wanted to start with the font guidelines. I guess the backend is settled, but I don't know how to access a FontSource in the frontend. Any help would be much appreciated. Thanks.

src/fontra/views/editor/edit-behavior.js
```diff
@@ -22,6 +22,8 @@ export class EditBehaviorFactory {
       point: pointSelection,
       component: componentSelection,
       anchor: anchorSelection,
+      guidelineGlyph: guidelineGlyphSelection,
```
justvanrossum, line 25: This name is awkward, there is no such thing as a "guideline glyph". I would call it "guideline", and for the globas/font guidelines we can later use "fontGuideline".

src/fontra/views/editor/edit-behavior.js
```diff
@@ -22,6 +22,8 @@ export class EditBehaviorFactory {
...
+      guidelineFont: guidelineFontSelection,
```
justvanrossum, line 26: Let's either leave this out completely, or at least commented out.

justvanrossum: On second thought, please remove all references to font guidelines here, because I don't think EditBehavior can deal with them, in a very fundamental way: EditBehavior is tied to the glyph, and the glyph changes, and the glyph undo stack.

I'm a little uncomfortable that we have to add support for guidelines here in the first place, and actually the same for anchors, but I don't readily have an alternative to suggest. I will try to come up with better factoring later, but that's for after this PR.

src/fontra/views/editor/editor.js
```diff
@@ -2030,6 +2104,259 @@ export class EditorController {
     return { contentElement, warningElement };
   }
 
+  selectionHasLockedGuidelines() {
+    const {
+      guidelineGlyph: guidelineGlyphSelection,
```
justvanrossum, line 2109: Same remark here about the name. Let's call this "guideline", and use "fontGuideline" later, once we've figured out how to deal with them.

src/fontra/views/editor/editor.js
```diff
@@ -2030,6 +2104,259 @@ export class EditorController {
...
+      guidelineFont: guidelineFontSelection,
+    } = parseSelection(this.sceneController.selection);
+
+    const instance = this.sceneModel.getSelectedPositionedGlyph()?.glyph.instance;
+    for (const guidelineIndex of guidelineGlyphSelection || []) {
+      const guidelineGlyph = instance.guidelines[guidelineIndex];
+      if (guidelineGlyph.locked) {
```
justvanrossum, line 2116: Suggest to omit the variable, and just write
```js
      if (instance.guidelines[guidelineIndex].locked) {
        return true;
      }

File: src/fontra/views/editor/editor.js [hunk 5]
```diff
@@ -2030,6 +2104,259 @@ export class EditorController {
...
+
+    return false;
+  }
+
+  getLockGuidelineLabel(hasLockedGuidelines) {
+    const {
+      guidelineGlyph: guidelineGlyphSelection,
+      guidelineFont: guidelineFontSelection,
+    } = parseSelection(this.sceneController.selection);
+    const guidelinSelection = new Array().concat(
```
justvanrossum, line 2132: Suggest to avoid contructing arrays that aren't really used apart from counting.

Maybe:
```js
    const numGuidelines = (guidelineGlyph?.length || 0) + (guidelineFont?.length || 0);
    // do things based on numGuidelines
```

Same for canLockGuideline() below.

src/fontra/views/editor/editor.js
justvanrossum: Better suggestion for the whole for loop:
```js
    if (guidelineGlyphSelection?.some((index) => instance.guidelines[index].locked) {
      return true;
    }
```

justvanrossum (commented):
Not a full review (I know it's wip anyway), but here are some suggestions.

src/fontra/views/editor/edit-behavior.js
ollimeier: I removed all 'fontGuideline' comments from edit-behavior

src/fontra/views/editor/editor.js
ollimeier: I renamed all guidelines.

src/fontra/views/editor/editor.js
ollimeier: Thanks a lot. I changed it.

src/fontra/views/editor/editor.js
ollimeier: Thanks for the valid feedback. I changed it.

justvanrossum (commented):
First round of review, there may be more, but later.

src/fontra/client/core/ui-utils.js
```diff
@@ -94,6 +94,27 @@ function didReorder(a, b) {
   return false;
 }
 
+export function labeledCheckbox(label, controller, key, options) {
+  const checkboxID = options?.id || `checkbox-${uniqueID()}-${key}`;
+  const items = [html.div()];
```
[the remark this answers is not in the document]
ollimeier: When I return `inputWrapper`, I get the following error:
`TypeError: labeledCheckbox is not a function or its return value is not iterable`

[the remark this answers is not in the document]
justvanrossum: Sure, you have to change the call site, too. Like I said, the _caller_ should handle that.

src/fontra/backends/designspace.py
```diff
@@ -1209,6 +1214,31 @@ def unpackAnchors(anchors):
...
+                name=g.get("name"),
+                x=g.get("x", 0),
+                y=g.get("y", 0),
+                angle=g.get("angle", 0),
+                locked=g.get("locked", False),
+                # TODO: Guidelines, how do we handle customData like:
+                # color=g.get("color"),
+                # identifier=g.get("identifier"),
+            )
+            if g.get("name")
```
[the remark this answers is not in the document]
ollimeier: I first had it this way, and then during coding it failed, why I added the else-code. It was related to the changes when assigning eg =0 to x, or y and angle. 
However, I double checked it – and it works without the else-code. I don't know what was cause it to fail, but it's no longer necessary. Thank you very much. You're absolutely wright, of cause. 

I mark this as resolved. I changed it, also the other.

src/fontra/backends/designspace.py
```diff
@@ -1237,6 +1267,14 @@ def populateUFOLayerGlyph(
     layerGlyph.anchors = [
         {"name": a.name, "x": a.x, "y": a.y} for a in staticGlyph.anchors
     ]
+    layerGlyph.guidelines = [
+        (
+            {"name": g.name, "x": g.x, "y": g.y, "angle": g.angle}
+            if g.name
```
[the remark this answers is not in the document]
ollimeier: Thanks for the valid review. Changed it.

src/fontra/client/core/ui-utils.js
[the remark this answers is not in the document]
ollimeier: Thanks once again. And sorry for my stupidity. I changed it.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -281,6 +279,138 @@ registerVisualizationLayerDefinition({
...
+      -textWidth / 2 -
+        parameters.margin * 2 -
+        parameters.originMarkerRadius / 2 -
+        parameters.strokeWidth * 2,
+      textHeight / 2 - parameters.fontSize * 0.2 // move up -> visually centered
+    );
+  }
+
+  // draw the line
+  if (guideline.name !== undefined) {
```
justvanrossum, line 366: In the next bit, there are many calls to `strokeLineDashed` that are very repetitive. All args are always the same, except for the x1 and x2 argumens.

I suggest to build an array instead, with an item per line you need to draw, and each item is a two-item array: [x1, x2]. 

After collecting this info, you can do a trivial `for (const [x1, x2] of lines) { ...do strokeLineDashed()... }`

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -281,6 +279,138 @@ registerVisualizationLayerDefinition({
...
+    for (const guideline of positionedGlyph.glyph.guidelines) {
+      _drawGuideline(context, parameters, guideline);
+    }
+
+    // TODO: Font Guidelines
+  },
+});
+
+function _drawGuideline(context, parameters, guideline) {
+  context.save();
```
justvanrossum, line 322: - Please use withSavedState() instead of context.save()/context.restore(). That is a better pattern, and is used all over the place. They can nest as well.
- Please clean up the transformation logic, so that you don't need to "rotate back" to draw the lock icon.

The overall structure will be something like this:
```
    withSavedState(() => {
      translate to guideline origin
      draw lock icon or the "node"
      withSavedState(() => {
        rotate
        scale 1, -1
        draw name
        draw lines
      }  
    }
```

justvanrossum (commented):
Some drawing code feedback.

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: I am totally with you. I refactored it.

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: Thanks a lot for the great and valid feedback. Done.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -851,6 +981,117 @@ registerVisualizationLayerDefinition({
   },
 });
 
+registerVisualizationLayerDefinition({
```
justvanrossum, line 984: The two guideline vis layers should be next to each other. Why are they so far apart?

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -851,6 +981,117 @@ registerVisualizationLayerDefinition({
...
+          parameters,
+          guideline.x - parameters.iconSize / 2,
+          guideline.y + parameters.iconSize / 2
+        );
+      } else {
+        fillRoundNode(context, guideline, smoothSize);
+      }
+    }
+
+    // // Hovered guideline
```
justvanrossum, line 1062: Double comment marker.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -851,6 +981,117 @@ registerVisualizationLayerDefinition({
...
+      }
+    }
+  },
+});
+
+function _drawLockIcon(context, parameters, x, y) {
+  context.save();
+  context.translate(x, y);
+  context.scale(parameters.iconSize / 24, (-1 * parameters.iconSize) / 24);
+  context.lineWidth = 2;
```
justvanrossum, line 1089: I'm not sure of the status of the hover display behavior we discussed. I don't currently see it happening. Is it not yet implemented or is this lineWidth = 2 somehow mistaken?

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -851,6 +981,117 @@ registerVisualizationLayerDefinition({
...
+          guideline.y + parameters.iconSize / 2
+        );
+      } else {
+        strokeRoundNode(context, guideline, smoothSize + parameters.hoverStrokeOffset);
+      }
+    }
+  },
+});
+
+function _drawLockIcon(context, parameters, x, y) {
```
justvanrossum, line 1085: Suggest to pass the concrete drawing parameters instead of the parameters object. It seem you only use `parameters.iconSize` and `strokeColor`. These kind of helper functions should be fairly generic, and tying it into the parameters object creates a tighter coupling than needed.

justvanrossum (commented):
- x, y coordinates should be shown on hover (if "Coordinates" is on), just like nodes and component origins
- When adding a new guide, the angle field is empty: I would expect it to be 0 instead

I think we should add code that keeps glyphs interpolatable even if guidelines are not. I can show you how/where to do this.

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: Thank you very much for the hint. Obviously I made it already wrong with anchors. I moved it (.selected.guidelines + .selected.anchors)

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: Do you see this:

https://github.com/googlefonts/fontra/assets/21055547/08a1afd6-3da6-4f16-a0bb-6afb1d25091c

ollimeier:
> * x, y coordinates should be shown on hover (if "Coordinates" is on), just like nodes and component origins
> * When adding a new guide, the angle field is empty: I would expect it to be 0 instead
> 
> I think we should add code that keeps glyphs interpolatable even if guidelines are not. I can show you how/where to do this.

- show coordinates: done.
- adding new guidelines with default angle 0: done.

I agree with you about 'glyph interpolation'. I would much appreciate if you would have time to show me how/where to do this.

src/fontra/views/editor/visualization-layer-definitions.js
justvanrossum: Now I do! I didn't last night.

However, when the guide is selected, and I hover, I do see the outline but the lock itself is displayed as if it's not selected. So I see no distinction between "hovered and selected" and "hovered and not selected".

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: thanks. done.

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: thanks. done.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -1322,6 +1564,15 @@ function* iterAnchorsPointsByIndex(anchors, anchorIndices) {
   }
 }
 
+function* iterGuidelinesPointsByIndex(guidelines, guidelineIndices) {
+  if (!guidelineIndices || !guidelines.length) {
+    return;
+  }
+  for (const index of guidelineIndices) {
+    yield guidelines[index];
```
justvanrossum, line 1572: There is a bug when one source has more guidelines than others:
- select the last guideline (that doesn't exist in other guidelines)
- then switch to another source

`guidelines[index]` will be undefined in that case, and causes an error when drawing the coordinates. This code should _not_ yield the guideline if it is undefined.

The same bug is in the anchors code, please fix it there, too.

src/fontra/views/editor/editor.js
```diff
@@ -1013,6 +1022,34 @@ export class EditorController {
     });
   }
 
+  async doubleClickedGuidelinesCallback(event) {
+    const glyphController = await this.sceneModel.getSelectedStaticGlyphController();
+    const instance = glyphController.instance;
+
+    const guidelineIndex = this.sceneController.doubleClickedGuidelineIndices[0];
+    let guideline = instance.guidelines[guidelineIndex];
+    const { guideline: newGuideline } = await this.doAddEditGuidelineDialog(guideline);
```
justvanrossum, line 1031: When we're not "at a source", we should get the "not at a source" dialog instead. We currently only get it _after_ dismissing this dialog. Check the .canEdit attribute of the glyph controller.

ollimeier: Thanks a lot for you solid review and testing. It missed in for anchors as well. I fixed it for both.

src/fontra/views/editor/editor.js
```diff
@@ -990,6 +999,10 @@ export class EditorController {
 
   async doubleClickedAnchorsCallback(event) {
     const glyphController = await this.sceneModel.getSelectedStaticGlyphController();
+    if (!glyphController.canEdit) {
+      this.sceneController._dispatchEvent("glyphEditLocationNotAtSource");
+      return;
+    }
```
justvanrossum, line 1005: Suggest to add blank line after this, to be consistent with the similar block in doubleClickedGuidelinesCallback.

ollimeier: done

PR #1355, opened 2024-05-13
Fix #1041: Wrap long lines based on a settable maximum line width

### LINKED ISSUE #1041: Add function to set maximum line width
It'd be helpful to be able to set a maximum line width for the text box/canvas area, so that the text automatically goes to a new line when it exceeds this threshold.

Currently when we paste a long text string into the text box, we have to separate it into shorter lines manually.

hwoongkang:
Previously, rendering glyphs in the scene depended solely on `glyphLines` property, which originates from text entry of side panel.

And I recognized that to handle alignments, advance width of glyphs and total line lengths were already being tracked.

By using stack and pushing the "remaining line" on the top of the stack when the line length exceeds certain threshold, it was possible to implement the line wrapping algorithm.

--

But the problem is that currently "selection API" depends on "indices", and various codes accessing the current selected glyphs use these line and glyph indices to get glyph info out of `glyphLines` property.

This leads to a bug:
- the text I used is:
```
abcdeabcdeabcdeabcdeabcdeabcde
cdefg
```
- it gets wrapped as:
```
abcdeabcdeabcdeabcde
abcdeabcde
cdefg
```

when I select `b` from the "wrapped second line", and edit it, `d` gets edited because it is the second glyph of the second line, with respect to `glyphLines`

Therefore I handled this case by changing related codes to use `positionedLines` instead of `glyphLines`. Since type signatures of two data differs, some transforming is done. This is what second commit is about.

But it feels too much "ad hoc" for me, so any suggestions or reviews will be very helpful

### DIFF
--- a/src/fontra/views/editor/scene-controller.js
+++ b/src/fontra/views/editor/scene-controller.js
@@ -155,9 +155,14 @@ export class SceneController {
     this.sceneSettingsController.addKeyListener(
       ["selectedGlyph", "glyphLines"],
       (event) => {
+        const lines = this.sceneModel.positionedLines.map((line) => {
+          return line.glyphs.map((g) => ({
+            glyphName: g.glyphName,
+          }));
+        });
         this.sceneSettings.selectedGlyphName = getSelectedGlyphName(
           this.sceneSettings.selectedGlyph,
-          this.sceneSettings.glyphLines
+          lines
         );
       },
       true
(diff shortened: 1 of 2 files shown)


justvanrossum:
Thank you for your proposal. I see the problem with the indices, but I will have to think more how to resolve it.

PR #1357, opened 2024-05-14
Fix status color issue with newly created glyph or layer

### LINKED ISSUE #1356: Status Color canvas does no show color for newly created glyph or layer
Development status color in canvas should fall back to the default status

If a glyph source does not have a development status set, the source list shows it as "red" (or, more precicely, the default status), but shows nothing in the canvas. It should similarly fall back to the default status.

ollimeier:
Fixes #1356 

https://github.com/googlefonts/fontra/assets/21055547/a1e45bae-7f7f-4926-90b2-3e47d6ad77b9

### DIFF
--- a/src/fontra/views/editor/visualization-layer-definitions.js
+++ b/src/fontra/views/editor/visualization-layer-definitions.js
@@ -1049,11 +1049,13 @@ registerVisualizationLayerDefinition({
       positionedGlyph.varGlyph.sources[sourceIndex].customData[
         "fontra.development.status"
       ];
-    if (status === undefined) {
-      return;
-    }
 
-    const color = [...statusFieldDefinitions[status].color];
+    const defaultStatusValue = statusFieldDefinitions.find(
+      (statusDef) => statusDef.isDefault
+    )?.value;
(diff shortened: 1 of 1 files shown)


src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -1049,11 +1049,13 @@ registerVisualizationLayerDefinition({
...
       ];
-    if (status === undefined) {
-      return;
-    }
 
-    const color = [...statusFieldDefinitions[status].color];
+    const defaultStatusValue = statusFieldDefinitions.find(
+      (statusDef) => statusDef.isDefault
+    )?.value;
+    const statusValue = status !== undefined ? status : defaultStatusValue;
```
justvanrossum, line 1056: Suggest to make `status` a `let`, and only look up the default value if it indeed is undefined. This would be faster in the case where status _isn't_ undefined.

You should probably also return if the default status value can't be found.

ollimeier: Thanks. Done.

justvanrossum (approved):

PR #1380, opened 2024-05-22
Refine 'phantom' range slider

### LINKED ISSUE #1378: Refine "phantom" range slider
- make dark mode more consistent with light mode
- make contrast of numeric input text higher
- make contrast of track + thumb lower
- try to reduce the gap at the top of the numeric input, which causes the whole row to be taller than needed

ollimeier:
Fixes #1378 

Screenshot 2024-05-22 at 12 40 33

Screenshot 2024-05-22 at 12 40 56

### DIFF
--- a/src/fontra/client/web-components/designspace-location.js
+++ b/src/fontra/client/web-components/designspace-location.js
@@ -4,7 +4,7 @@ import * as html from "/core/html-utils.js";
 import { UnlitElement, htmlToElement } from "/core/html-utils.js";
 
 const colors = {
-  "disabled-text-color": ["#777", "#ccc"],
+  "disabled-color": ["#ccc", "#777"],
 };
 
 export class DesignspaceLocation extends UnlitElement {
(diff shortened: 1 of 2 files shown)


src/fontra/client/web-components/designspace-location.js
```diff
@@ -4,7 +4,7 @@ import * as html from "/core/html-utils.js";
 import { UnlitElement, htmlToElement } from "/core/html-utils.js";
 
 const colors = {
-  "disabled-color": ["#ccc", "#777"],
+  "disabled-text-color": ["#777", "#ccc"],
```
justvanrossum, line 7: Please undo the changes in this file. The text request was specifically about the numeric input, not the info box.

justvanrossum: But... please change the `gap` of `.slider-group` to `0.1em`

justvanrossum (commented):
Two small requests. Otherwise looks good, thanks!

src/fontra/client/web-components/designspace-location.js
ollimeier: Done.

PR #1389, opened 2024-05-23
UI editor for status color definitions

### LINKED ISSUE #1360: Create a UI editor for status colors
Write down requirements:

ollimeier:
Fixes #1360 

I guess it's now in a stage for a first draft PR. 
As mentioned in todays meeting, there seems to be an issue with the backend or frontend (where I definitely need help with). When there is no "fontra.sourceStatusFieldDefinitions" in the customData it fails adding a it.
I added a unittest 'test_statusFieldDefinitions' which shows, that it's possible to read and write the status definitions. But somehow it fails when I try to do it via the frontend, and sadly I don't know how to fix it.

The terminal says:
```
  File "/Users/ollimeier/Documents/hithub_data/fontra/src/fontra/core/changes.py", line 102, in _applyChange
    subject = subject[pathElement]
              ~~~~~~~^^^^^^^^^^^^^
KeyError: 'fontra.sourceStatusFieldDefinitions'
```

Thanks a lot in advance :)

### DIFF
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -353,6 +353,48 @@ export function rgbaToCSS(rgba) {
   return `rgb(${channels.join(",")})`;
 }
 
+export function hexToRgbaList(hexColor) {
+  // TODO: for now this works,
+  // but need to be refactored: recude redundancies
+  let c;
+  if (/^#[A-Fa-f0-9]{8}$/.test(hexColor)) {
+    c = hexColor.substring(1).split("");
+    return [
+      parseInt(c[0] + c[1], 16),
+      parseInt(c[2] + c[3], 16),
+      parseInt(c[4] + c[5], 16),
+      round(parseInt(c[6] + c[7], 16) / 255, 2),
+    ];
+  } else if (/^#[A-Fa-f0-9]{6}$/.test(hexColor)) {
+    c = hexColor.substring(1).split("");
+    return [
+      parseInt(c[0] + c[1], 16),
+      parseInt(c[2] + c[3], 16),
+      parseInt(c[4] + c[5], 16),
+      1,
+    ];
+  } else if (/^#[A-Fa-f0-9]{4}$/.test(hexColor)) {
+    c = hexColor.substring(1).split("");
+    return [
+      parseInt(c[0] + c[0], 16),
+      parseInt(c[1] + c[1], 16),
+      parseInt(c[2] + c[2], 16),
+      round(parseInt(c[3] + c[3], 16) / 255, 2),
+    ];
+  } else if (/^#[A-Fa-f0-9]{3}$/.test(hexColor)) {
+    console.log("hexColor: ", hexColor);
+    c = hexColor.substring(1).split("");
+    return [
+      parseInt(c[0] + c[0], 16),
+      parseInt(c[1] + c[1], 16),
+      parseInt(c[2] + c[2], 16),
+      1,
+    ];
+  } else {
+    new Error("Bad hex color format. Should be #RRGGBB or #RRGGBBAA or #RGB or #RGBA");
+  }
+}
+
 export function clamp(number, min, max) {
   return Math.max(Math.min(number, max), min);
 }
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -360,39 +360,59 @@ export function hexToRgbaList(hexColor) {
   if (/^#[A-Fa-f0-9]{8}$/.test(hexColor)) {
     c = hexColor.substring(1).split("");
     return [
-      parseInt(c[0] + c[1], 16),
-      parseInt(c[2] + c[3], 16),
-      parseInt(c[4] + c[5], 16),
+      round(parseInt(c[0] + c[1], 16) / 255, 2),
+      round(parseInt(c[2] + c[3], 16) / 255, 2),
+      round(parseInt(c[4] + c[5], 16) / 255, 2),
       round(parseInt(c[6] + c[7], 16) / 255, 2),
     ];
   } else if (/^#[A-Fa-f0-9]{6}$/.test(hexColor)) {
     c = hexColor.substring(1).split("");
     return [
-      parseInt(c[0] + c[1], 16),
-      parseInt(c[2] + c[3], 16),
-      parseInt(c[4] + c[5], 16),
+      round(parseInt(c[0] + c[1], 16) / 255, 2),
+      round(parseInt(c[2] + c[3], 16) / 255, 2),
+      round(parseInt(c[4] + c[5], 16) / 255, 2),
       1,
     ];
   } else if (/^#[A-Fa-f0-9]{4}$/.test(hexColor)) {
     c = hexColor.substring(1).split("");
     return [
-      parseInt(c[0] + c[0], 16),
-      parseInt(c[1] + c[1], 16),
-      parseInt(c[2] + c[2], 16),
+      round(parseInt(c[0] + c[0], 16) / 255, 2),
+      round(parseInt(c[1] + c[1], 16) / 255, 2),
+      round(parseInt(c[2] + c[2], 16) / 255, 2),
       round(parseInt(c[3] + c[3], 16) / 255, 2),
     ];
   } else if (/^#[A-Fa-f0-9]{3}$/.test(hexColor)) {
     console.log("hexColor: ", hexColor);
     c = hexColor.substring(1).split("");
     return [
-      parseInt(c[0] + c[0], 16),
-      parseInt(c[1] + c[1], 16),
-      parseInt(c[2] + c[2], 16),
+      round(parseInt(c[0] + c[0], 16) / 255, 2),
+      round(parseInt(c[1] + c[1], 16) / 255, 2),
+      round(parseInt(c[2] + c[2], 16) / 255, 2),
       1,
     ];
   } else {
     new Error("Bad hex color format. Should be #RRGGBB or #RRGGBBAA or #RGB or #RGBA");
   }
+}
+
+export function rgbaToHex(rgba) {
+  const [r, g, b, a] = rgba;
+  const red = Math.round(r * 255)
+    .toString(16)
+    .padStart(2, "0");
+  const green = Math.round(g * 255)
+    .toString(16)
+    .padStart(2, "0");
+  const blue = Math.round(b * 255)
+    .toString(16)
+    .padStart(2, "0");
+  const alpha = Math.round(a * 255)
+    .toString(16)
+    .padStart(2, "0");
+  if (alpha === "ff") {
+    return `#${red}${green}${blue}`;
+  }
+  return `#${red}${green}${blue}${alpha}`;
 }
 
 export function clamp(number, min, max) {
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -353,66 +353,42 @@ export function rgbaToCSS(rgba) {
   return `rgb(${channels.join(",")})`;
 }
 
-export function hexToRgbaList(hexColor) {
-  // TODO: for now this works,
-  // but need to be refactored: recude redundancies
-  let c;
-  if (/^#[A-Fa-f0-9]{8}$/.test(hexColor)) {
-    c = hexColor.substring(1).split("");
-    return [
-      round(parseInt(c[0] + c[1], 16) / 255, 2),
-      round(parseInt(c[2] + c[3], 16) / 255, 2),
-      round(parseInt(c[4] + c[5], 16) / 255, 2),
-      round(parseInt(c[6] + c[7], 16) / 255, 2),
-    ];
-  } else if (/^#[A-Fa-f0-9]{6}$/.test(hexColor)) {
-    c = hexColor.substring(1).split("");
-    return [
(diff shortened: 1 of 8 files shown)


src/fontra/views/fontinfo/panel-development-status-definitions.js
```diff
@@ -0,0 +1,325 @@
...
+  grid-row-gap: 0.1em;
+  grid-column-gap: 1em;
+}
+
+.fontra-ui-font-info-status-definitions-panel-status-def-box-delete {
+  justify-self: end;
+  /*align-self: start;*/
+}
+
+.fontra-ui-font-info-status-definitions-panel-status-def-box-color {
```
justvanrossum, line 174: Suggest to add `-input` to this already long name. It took me a few seconds to realize this is about the color picker, and not about an actual color...

src/fontra/views/fontinfo/panel-development-status-definitions.js
```diff
@@ -0,0 +1,325 @@
...
+  margin: 0;
+  padding: 0;
+  outline: none;
+  border: none;
+  border-color: transparent;
+  border-radius: 0.25em;
+}
+`);
+
+class StatusDefBox extends HTMLElement {
```
justvanrossum, line 186: Let's call this class `StatusDefinitionBox`.

src/fontra/views/fontinfo/panel-development-status-definitions.js
```diff
@@ -0,0 +1,325 @@
...
+      delete statusDef["isDefault"];
+      this.replaceStatusDef(statusDef, undoLabel);
+      return;
+    }
+
+    // If checked: Set all status definitions to false, first.
+    for (const [index, statusDefTemp] of enumerate(
+      this.fontController.customData["fontra.sourceStatusFieldDefinitions"]
+    )) {
+      delete statusDefTemp["isDefault"];
```
justvanrossum, line 239: `statusDefTemp` is real data, not a copy, and you are not supposed to modify that. I think you need something like this here:

```js
    const newStatusDef = {...statusDefTemp};  // Maybe just call it `oldStatusDef` ?
    delete newStatusDef["isDefault"];
```

src/fontra/views/fontinfo/panel-development-status-definitions.js
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  _updateContents() {
+    this.innerHTML = "";
+    const statusDef = this.statusDef;
+    this.append(
+      html.input({
+        "type": "color",
+        "class": "fontra-ui-font-info-status-definitions-panel-status-def-box-color",
+        "value": rgbaToHex(statusDef.color),
```
justvanrossum, line 257: I read that the color input does not support opacity, and that the value string _must_ be a 7-char string: `#FF9988`. However, `rgbaToHex` can return a 9-char string if opcaticy isn't 1. If we want to not support opacity (which would be totally okay), we need to make sure to drop any opacity. Perhaps like this:
```js
   statusDef.color.slice(0, 3).concat([1])
```

src/fontra/client/core/utils.js
```diff
@@ -353,6 +353,44 @@ export function rgbaToCSS(rgba) {
   return `rgb(${channels.join(",")})`;
 }
 
+export function hexToRgba(hexColor) {
+  let c = hexColor.substring(1).split("");
+  let r = [];
+  if (/^#[A-Fa-f0-9]{8}$/.test(hexColor) || /^#[A-Fa-f0-9]{6}$/.test(hexColor)) {
+    for (const i of range(0, c.length, 2)) {
+      r.push(round(parseInt(c[i] + c[i + 1], 16) / 255, 2));
```
justvanrossum, line 361: 2 digits is not enough, you need three to be able to reliably round-trip.

I would go for 4, though, for some added precision.

Please also test round tripping between hexToRgba and rgbaToHex.

src/fontra/client/core/utils.js
```diff
@@ -353,6 +353,44 @@ export function rgbaToCSS(rgba) {
...
+    new Error("Bad hex color format. Should be #RRGGBB or #RRGGBBAA or #RGB or #RGBA");
+  }
+  if (r.length === 3) {
+    r.push(1);
+  }
+  return r;
+}
+
+export function rgbaToHex(rgba) {
+  const channels = rgba.slice(0, 3).map((channel) =>
```
justvanrossum, line 378: Suggest to throw an error if rgba.length is not 3 or 4.

Then, I would not do the slice, and later check the fourth element, and remove it if it is "ff". This would remove the need to repeat the hex conversion / zero padding code.

src/fontra/views/fontinfo/panel-development-status-definitions.js
```diff
@@ -0,0 +1,323 @@
...
+  }
+  return [];
+}
+
+addStyleSheet(`
+.fontra-ui-font-info-status-definitions-panel-status-def-box {
+  background-color: var(--ui-element-background-color);
+  border-radius: 0.5em;
+  padding: 1em;
+  cursor: pointer;
```
justvanrossum, line 161: The whole box now causes the cursor to be "pointer", but since you don't support drag and drop, this makes no sense: the user will now think it can drag the box while it can't.

justvanrossum (commented):
I'm still missing a field for the code integer value?

src/fontra/views/fontinfo/panel-development-status-definitions.js
ollimeier: OK. Did it.

src/fontra/views/fontinfo/panel-development-status-definitions.js
ollimeier: changed it.

src/fontra/views/fontinfo/panel-development-status-definitions.js
ollimeier: thanks. done.

src/fontra/views/fontinfo/panel-development-status-definitions.js
ollimeier: Good point. Thanks for the code snipped. Much appreciated.

src/fontra/client/core/utils.js
ollimeier: I have a solution, but I am not 100% sure if it's correct. Why I keep this open for now.

src/fontra/views/fontinfo/panel-development-status-definitions.js
ollimeier: I missed to remove it. Thanks for finding it.

src/fontra/client/core/utils.js
ollimeier: done.

ollimeier:
> I'm still missing a field for the code integer value?

OK, I added it.

src/fontra/client/core/utils.js
```diff
@@ -353,6 +353,42 @@ export function rgbaToCSS(rgba) {
...
+    new Error("Bad hex color format. Should be #RRGGBB or #RRGGBBAA or #RGB or #RGBA");
+  }
+  if (r.length === 3) {
+    r.push(1);
+  }
+  return r;
+}
+
+export function rgbaToHex(rgba) {
+  if (![3, 4].includes(rgba.length)) {
```
justvanrossum, line 378: This looks... unusual. Please just write `if (rgba.length != 3 && rgba.length != 4) ...`

src/fontra/client/core/utils.js
```diff
@@ -353,6 +353,42 @@ export function rgbaToCSS(rgba) {
...
+    throw new Error("rgbaToHex: has to have 3 or 4 items in array");
+  }
+  if (rgba[3] == 1) {
+    rgba.pop();
```
justvanrossum, line 382: You should generally not modify arrays passed as an argument. It is not a copy, and you do not "own" this object.

src/fontra/client/core/utils.js
```diff
@@ -353,6 +353,42 @@ export function rgbaToCSS(rgba) {
...
+  }
+  const channels = rgba.slice(0, 4).map((channel) =>
```
justvanrossum, line 384: You can leave out the slice, since you already know it must be 3 or 4 items long.

src/fontra/client/core/utils.js
```diff
@@ -353,6 +353,42 @@ export function rgbaToCSS(rgba) {
...
+    Math.round(channel * 255)
+      .toString(16)
+      .padStart(2, "0")
+  );
```
justvanrossum, line 388: If you add this:
```js
    if (channels[3] === "ff") {
      channels.pop()
    }
```
You don't need to modify the rgba argument.

And now you've ended up exactly with what I originally suggested.

src/fontra/views/fontinfo/panel-development-status-definitions.js
```diff
@@ -0,0 +1,362 @@
...
+      this.postChange(changes.change, changes.rollbackChange, undoLabel);
+      this.setupUI();
+    }
+  }
+
+  deleteStatusDef(statusIndex) {
+    const undoLabel = `delete status def '${this.statusDef.name}'`;
+    const root = { customData: this.fontController.customData };
+    const changes = recordChanges(root, (root) => {
+      root.customData["fontra.sourceStatusFieldDefinitions"].splice(statusIndex, 1);
```
justvanrossum, line 241: If the list gets empty as the result of this, the whole customData entry should be deleted.

justvanrossum (commented):
Something is working not quite right: if I start with a .designspace without definitions, after I click "New status definition", the document contains two identical status definitions.

Is undo supposed to work at this point?

src/fontra/client/core/utils.js
ollimeier: thanks.

src/fontra/client/core/utils.js
ollimeier: sorry and thanks. done.

src/fontra/views/fontinfo/panel-development-status-definitions.js
ollimeier: Not sure if this is related to that, but I am wondering if we want to update/modify the status definitions for each glyph source as well? For example, when editing the status definition value, do we want to update the value in all glyph sources? And if we delete one status definition, do we want to delete this value in all glyph sources?

justvanrossum: > but I am wondering if we want to update/modify the status definitions for each glyph source as well? 

No, that is completely unfeasible. Imagine a font with 40000 glyphs which is stored remotely in a database.

src/fontra/client/core/utils.js
```diff
@@ -353,6 +353,42 @@ export function rgbaToCSS(rgba) {
...
+    for (const i of range(0, c.length, 2)) {
+      r.push(round(parseInt(c[i] + c[i + 1], 16) / 255, 4));
+    }
+  } else if (/^#[A-Fa-f0-9]{4}$/.test(hexColor) || /^#[A-Fa-f0-9]{3}$/.test(hexColor)) {
+    for (const i of range(c.length)) {
+      r.push(round(parseInt(c[i] + c[i], 16) / 255, 4));
+    }
+  } else {
+    r = [1, 0, 0, 1];
+    new Error("Bad hex color format. Should be #RRGGBB or #RRGGBBAA or #RGB or #RGBA");
```
justvanrossum, line 369: Missing "throw", and likely missing test for this error.

src/fontra/client/core/utils.js
```diff
@@ -353,6 +353,42 @@ export function rgbaToCSS(rgba) {
...
+  }
+  if (r.length === 3) {
+    r.push(1);
+  }
+  return r;
+}
+
+export function rgbaToHex(rgba) {
+  if (rgba.length != 3 && rgba.length != 4) {
+    throw new Error("rgbaToHex: has to have 3 or 4 items in array");
```
justvanrossum, line 379:
```suggestion
    throw new Error("rgba argument has to have 3 or 4 items in array");
```

Note that this will break the test.

src/fontra/client/core/utils.js
ollimeier: done.

src/fontra/views/fontinfo/panel-development-status-definitions.js
ollimeier: OK, in that case I am closing this.

ollimeier:
> Something is working not quite right: if I start with a .designspace without definitions, after I click "New status definition", the document contains two identical status definitions.
> 
> Is undo supposed to work at this point?

I fixed the bug with the 'two identical status definitions', but unfortunately I don't know how to make the undo work. Any advice is very much appreciated.

src/fontra/views/fontinfo/panel-development-status-definitions.js
```diff
@@ -0,0 +1,372 @@
...
+      })
+    );
+
+    const checkBoxIdentifier = `statusDefIsDefault${this.statusIndex}`;
+    this.append(
+      html.div(
+        {
+          "style": "margin: auto;",
+          "data-tooltip":
+            "If set: This color will be used if a source status is not set.",
```
justvanrossum, line 339:
```suggestion
            "If checked, this status will be used as a fallback when a source status is not set",
```

justvanrossum (approved):

PR #1399, opened 2024-05-28
Bugfix command A

### LINKED ISSUE #1394: Outline editor: command-A doesn't do the right thing when some points are already selected
Reproduce:
- go to a glyph with outlines
- select a point
- do "select All"
- it does "select None" instead

I guess this is a regression introduced with "smart command A" behavior for anchors and/or guidelines.

ollimeier:
Fixes #1394

### DIFF
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -2436,8 +2436,8 @@ export class EditorController {
     const hasGuidelines = instance.guidelines.length > 0;
 
     if (
-      !pointIndices.length &&
-      !componentIndices.length &&
+      (pointIndices.length < instance.path.pointTypes.length ||
+        componentIndices.length < instance.components.length) &&
       !anchorIndices.length &&
       !guidelineIndices.length
       //&& !fontGuidelineIndices.length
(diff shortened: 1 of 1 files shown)


src/fontra/views/editor/editor.js
```diff
@@ -2435,8 +2435,16 @@ export class EditorController {
     const hasAnchors = instance.anchors.length > 0;
     const hasGuidelines = instance.guidelines.length > 0;
 
+    const glyphPath = positionedGlyph.glyph.path;
+    let onCurvePoints = [];
```
justvanrossum, line 2439: It would be more correct if we'd check: "does the selection contain all on-curve points?"

(Imagine a path with four on-curve points and four off-curve points. Select to on-curve and two off-curve, and your current logic alreade considers this "all points have been selected".)

I suggest to use a Set for onCurvePoints (replace onCurvePoints.push with onCurvePoints.add), then use `isSuperset(onCurvePoints, pointIndices)` to get the answer to the above question.

isSuperset comes from set-ops.js, and the second argument does not need to be a set: you can use pointIndices as-is.

justvanrossum: Compute isSuperset() only once: you use its inverse below. As least that's how I read your code.

justvanrossum (approved):

PR #1403, opened 2024-05-29
Support glyph source status color code in UFO/Designspace

### LINKED ISSUE #1364: [designspace backend] Support reading/writing the development status code to/from UFO/designspace
This may involve implementing a general way to store GlyphSource.customData as a whole in the source .glif. (Layer instances also have a customData attribute, which is distinct. Would be nice to solve both, though.)

Also, check whether we read the designspace lib correctly, which is where I expect the dev status definitions to be stored.

Relates to #1360.

ollimeier:
Fixes #1364

### DIFF
--- a/src/fontra/backends/designspace.py
+++ b/src/fontra/backends/designspace.py
@@ -60,8 +60,7 @@ VARIABLE_COMPONENTS_LIB_KEY = "com.black-foundry.variable-components"
 GLYPH_DESIGNSPACE_LIB_KEY = "com.black-foundry.glyph-designspace"
 SOURCE_NAME_MAPPING_LIB_KEY = "xyz.fontra.source-names"
 LAYER_NAME_MAPPING_LIB_KEY = "xyz.fontra.layer-names"
-CUSTOM_DATA_LIB_KEY = "xyz.fontra.customData"
-GLYPH_CUSTOM_DATA_LIB_KEY = "xyz.fontra.glyph.customData"
+GLYPH_CUSTOM_DATA_LIB_KEY = "xyz.fontra.customData"
 GLYPH_SOURCE_CUSTOM_DATA_LIB_KEY = "xyz.fontra.glyph.source.customData"
 
 
@@ -310,13 +309,13 @@ class DesignspaceBackend:
                     )
                 sourceNameMapping = ufoGlyph.lib.get(SOURCE_NAME_MAPPING_LIB_KEY, {})
                 layerNameMapping = ufoGlyph.lib.get(LAYER_NAME_MAPPING_LIB_KEY, {})
-                customData = ufoGlyph.lib.get(CUSTOM_DATA_LIB_KEY, {})
+                customData = ufoGlyph.lib.get(GLYPH_CUSTOM_DATA_LIB_KEY, {})
 
             layerName = layerNameMapping.get(
                 ufoLayer.fontraLayerName, ufoLayer.fontraLayerName
             )
             localSourcesCustomData[layerName] = ufoGlyph.lib.get(
-                CUSTOM_DATA_LIB_KEY, {}
+                GLYPH_SOURCE_CUSTOM_DATA_LIB_KEY, {}
             )
 
             layers[ufoLayer.fontraLayerName] = Layer(glyph=staticGlyph)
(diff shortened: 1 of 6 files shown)


justvanrossum (commented):
I'm afraid this isn't quite going in the right direction:
- The source's customData should be stored in the glyph lib of the _associated_ ufo glyph, not collected in the default glyph
- Any tests need to go to test_backends_designspace.py, not test_backends_fontra.py, which is for testing the .fontra backend

test-py/test_backends_fontra.py
```diff
@@ -136,3 +136,23 @@ async def test_statusFieldDefinitions(writableFontraFont):
     await writableFontraFont.putCustomData(statusTestData)
 
     assert statusTestData == await writableFontraFont.getCustomData()
+
+
+async def test_getGlyphSourceStatusCode(testFontraFont):
```
justvanrossum, line 141: These tests are redundant, as there is no special support needed for customData or status codes. Please remove these tests.

test-common/fonts/MutatorSans.fontra/glyphs/E^1.json
```diff
@@ -8,6 +8,9 @@
 "italic": 0,
 "weight": 150,
 "width": 0
+},
```
justvanrossum, line 11: Together with the redundant tests in tests_backends_fontra.py, this change can be reverted, too.

src/fontra/backends/designspace.py
```diff
@@ -459,14 +474,22 @@ async def putGlyph(
...
 
+            customData = None
             if glyphSet == self.defaultUFOLayer.glyphSet:
                 layerGlyph = defaultLayerGlyph
                 storeInLib(layerGlyph, GLYPH_DESIGNSPACE_LIB_KEY, localDS)
                 storeInLib(layerGlyph, SOURCE_NAME_MAPPING_LIB_KEY, sourceNameMapping)
                 storeInLib(layerGlyph, LAYER_NAME_MAPPING_LIB_KEY, layerNameMapping)
-                storeInLib(layerGlyph, CUSTOM_DATA_LIB_KEY, glyph.customData)
+                customData = (
+                    glyph.customData | localSourcesCustomData[ufoLayer.fontraLayerName]
```
justvanrossum, line 484: We need to differentiate between these two, we can't just fold them into one. 

Suggest to use two keys:
```python
GLYPH_CUSTOM_DATA_LIB_KEY = "xyz.fontra.glyph.customData"
GLYPH_SOURCE_CUSTOM_DATA_LIB_KEY = "xyz.fontra.glyph.source.customData"
```

src/fontra/backends/designspace.py
```diff
@@ -459,14 +474,22 @@ async def putGlyph(
...
+                )
             else:
                 layerGlyph = readGlyphOrCreate(glyphSet, glyphName, codePoints)
+                if localSourcesCustomData.get(ufoLayer.fontraLayerName) is not None:
```
justvanrossum, line 488: General Python advice:
- To check if a key exists in a dict, you write "if key in dict"
- To get a key with a fallback, use "dict.get(key)"

You _either_ first check if a key exists, and then retrieve it normally, or you "get" it and then check the value for None. You are using a mix of both here.

test-py/test_backends_fontra.py
ollimeier: done.

test-common/fonts/MutatorSans.fontra/glyphs/E^1.json
ollimeier: done.

src/fontra/backends/designspace.py
ollimeier: Thank you very much for the valid feedback. I changed it.

src/fontra/backends/designspace.py
justvanrossum: Actually, you just moved it in [this commit](https://github.com/googlefonts/fontra/pull/1403/commits/34e934aa51a03837d783ee84d5ae00100cb9d086)... I fixed it (together with a general logic error) [here](https://github.com/googlefonts/fontra/pull/1403/commits/7ff723a36ed16c4a22dd417022e89dc713801ef5).

PR #1416, opened 2024-05-31
Show guidelines: on default + when adding a new guideline

### LINKED ISSUE #1414: Add new guideline, but guideline visualization is off
User feedback: 
It's strange when adding a new guideline it will not show it when the guideline visualization is off. 
See demo video:

https://github.com/googlefonts/fontra/assets/21055547/2d6b405c-b96a-4cee-b2f5-75f1f04f34a6

Possible solution:
1. set guideline visualization to 'On' by default
2. when adding a new guideline turn on the visualization (does not matter what setting it was about)

@JeremieHornus comment: _i feel there should be no option, like for anchors: always visible_

ollimeier:
Fixes #1414

### DIFF
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -2248,6 +2248,7 @@ export class EditorController {
   }
 
   async doAddGuideline(global = false) {
+    this.visualizationLayersSettings.model["fontra.guidelines"] = true;
     const point = this.sceneController.selectedGlyphPoint(this.contextMenuPosition);
     const { guideline: tempGuideline } = await this.doAddEditGuidelineDialog(
       undefined,
--- a/src/fontra/views/editor/visualization-layer-definitions.js
+++ b/src/fontra/views/editor/visualization-layer-definitions.js
@@ -339,7 +339,7 @@ registerVisualizationLayerDefinition({
   name: "Guidelines",
   selectionMode: "editing",
   userSwitchable: true,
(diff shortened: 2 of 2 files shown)


src/fontra/views/editor/editor.js
```diff
@@ -2248,6 +2248,7 @@ export class EditorController {
   }
 
   async doAddGuideline(global = false) {
+    this.visualizationLayersSettings.model["fontra.guidelines"] = true;
```
justvanrossum, line 2251: I feel that perhaps this should be done after the dialog. Not 100% sure. Did you consider that?

ollimeier: Ah, you mean only get active when you really add a guideline and not if you cancel it?! I see your point. 
I tried it and I found it useful to have it displayed immediately, because in the case you simply forgot to active 'show guidelines' and you want to add a new guideline at the same position where you have already added a guideline but you forgot about it, you will then see it in the background, which will be a reminder: 'Ah. Oh. Yes, there is already a guideline. I simply missed to show the guidelines'. In that case you can cancel adding guidelines and see/adjust the existing guidelines. 
But if you want to have it different I am happy to change it.

justvanrossum: That’s a good argument, let’s keep it this way!

justvanrossum (approved):
Thanks for the quick fix.

PR #1417, opened 2024-05-31
Show vertical metrics in canvas

### LINKED ISSUE #1406: Show vertical metrics in canvas
... when editing a glyph via registerVisualizationLayerDefinition

ollimeier:
Fixes #1406


src/fontra/core/classes.py
```diff
@@ -85,13 +85,15 @@ class FontSource:
...
     verticalMetrics: dict[str, FontMetric] = field(default_factory=dict)
+    italicAngle: float = 0
     guidelines: list[Guideline] = field(default_factory=list)
     customData: CustomData = field(default_factory=dict)
 
 
 @dataclass(kw_only=True)
 class FontMetric:
     value: float
+    zone: int = 0
```
justvanrossum, line 96: This should be a float, like all other metric numbers.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -171,6 +171,66 @@ registerVisualizationLayerDefinition({
...
+  colorsDarkMode: { strokeColor: "#FFF6", zoneColor: "#FFF1" },
+  draw: (context, positionedGlyph, parameters, model, controller) => {
+    context.strokeStyle = parameters.strokeColor;
+    context.lineWidth = parameters.strokeWidth;
+
+    if (model.fontSourceInstance === undefined) {
+      return;
+    }
+    const verticalMetrics = model.fontSourceInstance.verticalMetrics;
+    const italicsAngle = model.fontSourceInstance.italicsAngle
```
justvanrossum, line 192: italicsAngle -> italicAngle.

justvanrossum (commented):
Spotted two things.

src/fontra/core/classes.py
ollimeier: thanks.

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: damn. Thanks for spotting it.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -171,6 +171,66 @@ registerVisualizationLayerDefinition({
...
+  defaultOn: true,
+  zIndex: 100,
+  screenParameters: { strokeWidth: 1 },
+  colors: { strokeColor: "#0004", zoneColor: "#0001" },
+  colorsDarkMode: { strokeColor: "#FFF6", zoneColor: "#FFF1" },
+  draw: (context, positionedGlyph, parameters, model, controller) => {
+    context.strokeStyle = parameters.strokeColor;
+    context.lineWidth = parameters.strokeWidth;
+
+    if (model.fontSourceInstance === undefined) {
```
justvanrossum, line 188:
```suggestion
    if (!model.fontSourceInstance) {
```

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -171,6 +171,66 @@ registerVisualizationLayerDefinition({
...
+    //   : 0;
+    const glyphWidth = positionedGlyph.glyph.xAdvance
+      ? positionedGlyph.glyph.xAdvance
+      : 0;
+
+    // glyph box
+    let pathBox = new Path2D();
+    pathBox.rect(
+      0,
+      verticalMetrics.descender.value,
```
justvanrossum, line 203: We need to account for the possibility that "ascender" or "descender" or both do not exist, and don't draw the box in that case.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -171,6 +171,66 @@ registerVisualizationLayerDefinition({
...
+      positionedGlyph.glyph.xAdvance,
+      verticalMetrics.ascender.value - verticalMetrics.descender.value
+    );
+
+    // collect paths: vertical metrics and alignment zones
+    let pathZones = new Path2D();
+    for (const [key, metric] of Object.entries(verticalMetrics)) {
+      let pathZone = new Path2D();
```
justvanrossum, line 211: Maybe skip this if `metric.zone == 0`?

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -171,6 +171,66 @@ registerVisualizationLayerDefinition({
...
+
+      let pathMetric = new Path2D();
+      pathMetric.moveTo(0, metric.value);
+      pathMetric.lineTo(glyphWidth, metric.value);
+      pathBox.addPath(pathMetric);
+    }
+
+    // draw zones (with filled path)
+    let path2DZones = new Path2D();
+    path2DZones.addPath(pathZones); //, new DOMMatrix().skewX(italicAngle));
```
justvanrossum, line 223: Why the additional path object? Same below for `path2DBox`.

src/fontra/views/editor/visualization-layer-definitions.js
justvanrossum: Right, this is indeed true for designspace. But since our data structure does not guarantee this, we still need to check.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -208,9 +208,11 @@ registerVisualizationLayerDefinition({
     // collect paths: vertical metrics and alignment zones
     let pathZones = new Path2D();
     for (const [key, metric] of Object.entries(verticalMetrics)) {
-      let pathZone = new Path2D();
-      pathZone.rect(0, metric.value, glyphWidth, metric.zone);
-      pathZones.addPath(pathZone);
+      if (metric.zone != 0) {
```
justvanrossum, line 211: Suggest to write `if (metric.zone)`, so we also catch null and undefined.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -208,9 +208,11 @@ registerVisualizationLayerDefinition({
...
+        let pathZone = new Path2D();
```
justvanrossum, line 214: This can be a const.

FWIW: If you never assign to a variable again after the initial assignment, you don't need let and const is enough. It has nothing to do with the mutability of the object.

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: This is/was related to the 'italicAngle' I commented out for now. I needed it to add `new DOMMetrix` with skewX.

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: done.

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: Ah, sorry. I misunderstood your question. the second additional is required, because the one has a fill (zones) and the other is stroke only. Is there a way to combine both in one?

justvanrossum: I think both are redundant because of the removal of the skew transform. A new Path2D is made and one path is added to it without change. Maybe best to remove the italicAngle code after all, it will be trivial enough to add back in later if needed.

Btw. I think _all_ the `let`s should be `const` in this function. I don't see a single case where a `const` wouldn't work. `const` should always be the default.

ollimeier: OK, I removed the italicAngle code and used const instead of let.

PR #1427, opened 2024-06-03
Add sources panel

### LINKED ISSUE #1411: Display font sources and vertical metrics in font info panel under 'Sources'
Similar to the axes-panel the sources-panel should be displayed with boxes, one box for each source.

Use some demo-data (maybe as json) to sketch out the UI.

In future the content/data for each source will grow. It may include hhea/OS/2 vertical metrics or other information. Ignore this for now and focus on location and the 'default' vertical metrics. If the content grows we may add foldable elements.

Maybe show locations in a summarized version (similar to designspace-navigation-panel) and when double clicked open a dialog for editing the data.

Delete icon on the right, similar to axes-panel or status-definitions-panel.

Add 'zone' as an attribute to FontMetrics.

Keep in mind how it could be refactored in future in a more general way.

ollimeier:
Fixes #1411


ollimeier:
I am closing this draft pull request, because @justvanrossum found the bug, so I can continue working on it.

PR #1430, opened 2024-06-04
Issue 1411 add sources panel

### LINKED ISSUE #1411: Display font sources and vertical metrics in font info panel under 'Sources'

ollimeier:
Fixes #1411


justvanrossum:
I committed a fix. You used the "real" underlying data for the location and vertical metrics, meaning by the time the event was listened to, the change _had already occurred_, so for the undo mechanism the "old" value was the same as the "new" value.

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,604 @@
...
+    });
+    if (changes.hasChange) {
+      this.postChange(changes.change, changes.rollbackChange, undoLabel);
+      this.setupUI();
+    }
+  }
+
+  toggleShowHide() {
+    const el = this.querySelector("#open-close-icon");
+    el.style.transform =
```
justvanrossum, line 416: Please use the following tactic (used in the accordion view, and elsewhere):
- instead of changing the style via code, you add or remove a (well named) class
- then, in the CSS you can select by that class, and apply the appropriate changes

This is the better and cleaner way, as it keeps the actual styling near the other styles instead of embedded in the code.

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,604 @@
...
+      el.style.transform === "rotate(180deg)" ? "rotate(0deg)" : "rotate(180deg)";
```
justvanrossum, line 417: Same here.

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,604 @@
...
+
+    this.style.maxHeight = this.style.maxHeight === "100%" ? "70px" : "100%";
+    for (const child of this.children) {
+      if (!child.style.maxHeight) {
+        child.style.maxHeight = "100%";
```
justvanrossum, line 422: And here.

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,604 @@
...
+      } else {
+        delete child.style;
+      }
+    }
+  }
+
+  _updateContents() {
+    const models = this.models;
+
+    for (const key in models) {
```
justvanrossum, line 432: Currently, `models` isn't _really_ a customizable / infinitely variable set of models. We literally have only three keys:
- general
- location
- verticalMetrics

In the code below, you add a generic listener callback to _all_ items, and the _inside_ the listener you switch again by key. This is inside out, and we're better off assigning the listeners explicitly. The code will be easier to understand, and possibly shorter as there's no need for all those if statements.

So, indeed use a loop to create the controllers (they are generic for each model), but then add listeners per item explicitly.

For example:
```js
    this.controllers.general.addListener((event) => {
      // listener code for "general"
    }
    this.controllers.location.addListener((event) => {
      // listener code for "location"
    }
    // etc.
```

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,604 @@
...
+        "src": "/tabler-icons/trash.svg",
+        "onclick": (event) => this.deleteSource(),
+        "data-tooltip": translate("sources.delete-source"),
+        "data-tooltipposition": "left",
+      })
     );
+
+    this.append(html.div()); // empty cell for grid with arrow
+
+    for (const key in models) {
```
justvanrossum, line 496: Likewise, this whole block can be much better written like this:
```js
    this.append(buildElement(this.controllers.general));
    this.append(buildElementLocations(this.controllers.location, this.fontAxes));
    this.append(buildElementVerticalMetrics(this.controllers.verticalMetrics));
```

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,604 @@
...
+  }
+
+  return html.div(
+    { class: "fontra-ui-font-info-column" },
+    items
+      .map(([labelName, keyName, value]) => {
+        return labeledTextInput(labelName, controller, keyName, {
+          continuous: false,
+        });
+        // TODO for isSparce
```
justvanrossum, line 525:
```suggestion
        // TODO for isSparse
```

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,604 @@
...
+  cursor: pointer;
+  display: grid;
+  grid-template-rows: auto auto;
+  grid-template-columns: max-content max-content max-content max-content auto;
+  grid-row-gap: 0.1em;
+  grid-column-gap: 1em;
+  max-height: 70px;
+}
+
+.fontra-ui-font-info-column {
```
justvanrossum, line 270: Please use the proper _full_ prefix:
`fontra-ui-font-info-sources-panel- ...`

Same in the rest of the classes below.

These classes are global, and therefore _must_ be unique, so we add a big fat prefix.

src/fontra/views/fontinfo/panel-sources.js
justvanrossum: In the case of `if (event.key == "name") {` you can use `addKeyListener("name", (event) => ...)` instead, and save another if.

src/fontra/client/core/ui-utils.js
```diff
@@ -1,5 +1,5 @@
 import * as html from "./html-utils.js";
-import { zip } from "./utils.js";
+import { round, zip } from "./utils.js";
```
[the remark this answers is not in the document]
ollimeier: Thanks I missed that. While I am in this file already, I noticed that the following two variables `choices` and `choicesID` in the lines 135 and 136 within `textInput` from your improvement are not used either:
  const choices = options?.choices;
  const choicesID = `${inputID}-choices`;

Should I remove them as well? Or do you want to it yourself or keep it for some reasons?

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,604 @@
...
+      location: { ...source.location },
+      verticalMetrics: this.prepareVerticalMetricsForController(source.verticalMetrics),
+      // TODO: hhea, OS/2 verticalMetrics, etc
+      // customData: { ...source.customData },
+    };
+    // NOTE: Font guidlines could be read/write here,
+    // but makes more sense directly in the glyph editing window.
+  }
+
+  prepareVerticalMetricsForController(verticalMetrics) {
```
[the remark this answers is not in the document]
ollimeier: Thanks. Done.

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,604 @@
...
+    let newVerticalMetrics = {};
```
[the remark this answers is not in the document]
ollimeier: Sorry. I will keep an eye on it more thoroughly. Done.

src/fontra/client/core/ui-utils.js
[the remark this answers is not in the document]
justvanrossum: Ah yes, please, that would be great.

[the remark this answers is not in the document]
ollimeier: done.

src/fontra/views/fontinfo/panel-sources.js
ollimeier: done.

src/fontra/views/fontinfo/panel-sources.js
ollimeier: done.

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,642 @@
...
+    const locationAxes = this.fontController.axes.axes;
+    const validateInput = () => {
+      const warnings = [];
+      const editedSourceName = nameController.model.sourceName;
+      if (!editedSourceName.length || !editedSourceName.trim()) {
+        warnings.push("⚠️ The source name must not be empty");
+      }
+      if (
+        Object.keys(sources)
+          .map(function (sourceIdentifier) {
```
justvanrossum, line 91: Please use an arrow function instead, because that is how we always spell anonymous funcs in Fontra.

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,642 @@
...
+
+    return {
+      verticalMetrics: getDefaultVerticalMetrics(this.fontController.unitsPerEm),
+      ...interpolatedSource,
+      ...newSource,
+    };
+  }
+
+  getSourceName(sources) {
+    const sourceNames = Object.keys(sources).map(function (sourceIdentifier) {
```
justvanrossum, line 204: Same here, please use an arrow func.

src/fontra/views/fontinfo/panel-sources.js
ollimeier: ok. done.

src/fontra/views/fontinfo/panel-sources.js
ollimeier: done.

src/fontra/views/fontinfo/panel-sources.js
ollimeier: I reworked the toggleShowHide and all css dependencies. I make this as done.

src/fontra/views/fontinfo/panel-sources.js
ollimeier: done.

src/fontra/views/fontinfo/panel-sources.js
ollimeier: done.

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,642 @@
...
+          .map((sourceIdentifier) => {
+            if (sources[sourceIdentifier].name === editedSourceName.trim()) {
+              return true;
+            }
+          })
+          .includes(true)
+      ) {
+        warnings.push("⚠️ The source name should be unique");
+      }
+      const editedItalicAngle = nameController.model.sourceItalicAngle;
```
justvanrossum, line 100: I don't think the italicAngle belongs in the dialog, just like we don't have fields for the vertical metrics.

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,642 @@
...
+    return this.sources[this.sourceIdentifier];
+  }
+
+  _getModels() {
+    const source = this.source;
+    return {
+      general: {
+        name: source.name,
+        italicAngle: source.italicAngle ? source.italicAngle : 0,
+        isSparse: source.isSparse ? source.isSparse : false,
```
justvanrossum, line 339: Please comment this out for now: this needs to become a more elaborate feature, because a sparse source should not have:
- vertical metrics
- italic angle

We will need to do this as a follow-up.

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,642 @@
...
+      },
+      location: { ...source.location },
+      verticalMetrics: prepareVerticalMetricsForController(source.verticalMetrics),
+      // TODO: hhea, OS/2 verticalMetrics, etc
+      // customData: { ...source.customData },
+    };
+    // NOTE: Font guidlines could be read/write here,
```
justvanrossum, line 346:
```suggestion
    // NOTE: Font guidelines could be read/write here,
```

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,642 @@
...
+  const defaultVerticalMetrics = {};
+  for (const [name, defaultFactor] of Object.entries(verticalMetricsDefaults)) {
+    const value = Math.round(defaultFactor.value * unitsPerEm);
+    const zone = Math.round(defaultFactor.zone * unitsPerEm);
+    defaultVerticalMetrics[name] = { value: value, zone: zone };
+  }
+
+  return defaultVerticalMetrics;
+}
+
```
justvanrossum, line 602: Please remove the blank line.

ollimeier: done

src/fontra/views/fontinfo/panel-sources.js
ollimeier: done.

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -1,14 +1,658 @@
...
+    ascender: "Ascender",
+    capHeight: "Cap Height",
+    xHeight: "x-Height",
+    baseline: "Baseline",
+    descender: "Descender",
+    general: "General",
+    location: "Location",
+    verticalMetrics: "Vertical metrics",
+  };
+  if (key in keyLabelMap) {
```
justvanrossum, line 654: These three lines can be this single line:

```js
    return keyLabelMap[key] || key;
```

justvanrossum (approved):
🎉 DONE 🎉!

PR #1440, opened 2024-06-06
Translation fallback to english

### LINKED ISSUE #1418: Localized translation falls back to "keys" when switching language before the font has fully loaded
This may be hard to reproduce with Fontra Pak, but with our server deployment, it may take several seconds before the page is loaded, and if you switch the language while loading (the spinner is spinning) the UI falls back to translation keys.

I can only reproduce in an incognito window, presumably because that prevents the cache from being successful.

image

### LINKED ISSUE #1446: [localization] Missing translations should fall back to English
This is needed so we can add English UI labels first, and fill in translations later.

ollimeier:
Fixes #1446.


justvanrossum:
I feel we first need to solve #1418.

justvanrossum:
I will close this: it contains some unrelated changes, and it also loads the "en" file twice if that's the current language. Also: there is a merge conflict caused by #1445.

PR #1453, opened 2024-06-10
Optionally set the 'overlap simple' flag for every exported glyph

### LINKED ISSUE #1450: [fontra-workflow] Need a way to set OVERLAP_SIMPLE for designspace/UFO export
This is to prevent rendering artifacts when outlines overlap exactly at the edge of a shape.

justvanrossum:
Another approach for fix #1450:

Add an optional flag to the DS backend that, when set to true, will set the overlap flag in all exported layer glyphs.

### DIFF
--- a/src/fontra/backends/designspace.py
+++ b/src/fontra/backends/designspace.py
@@ -127,6 +127,8 @@ class DesignspaceBackend:
         self.fileWatcherCallbacks: list[Callable[[Any], Awaitable[None]]] = []
         self._glyphDependenciesTask: asyncio.Task[GlyphDependencies] | None = None
         self._glyphDependencies: GlyphDependencies | None = None
+        # Set this to true to set "public.truetype.overlap" in each writte .glif's lib:
+        self.setOverlapSimpleFlag = False
         self._initialize(dsDoc)
 
     def _initialize(self, dsDoc: DesignSpaceDocument) -> None:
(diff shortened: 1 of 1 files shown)


anthrotype:
eagerly setting it to _all_ the glyphs may be counter-productive. There are some situations where the presence of the overlap flags induce Skia to simplify the paths (when rendering stroked outlines or synthesizing fake bolds) and some mysterious bugs in its PathOps module in turn produce rendering glitches (entire subpaths disappearing or incorrectly filled black), e.g. see https://b.corp.google.com/issues/338390594

my suggestion is to only activate the flags when necessary

justvanrossum:
The thing is, we have no way to determine "when it is necessary".

Also, it may not even help at all, but we haven't thoroughly tested.

We have shapes like this:

image

...which will render poorly in some circumstances. We are in the process of figuring out if setting the flag is effective at all.

>  see https://b.corp.google.com/issues/338390594

(I don't have access to that issue)

anthrotype:
> ...which will render poorly in some circumstances

yes, in FreeType and CoreText IIRC. At least for the former, setting the flag should help since it activates the slower "oversampling" rendering technique, cf https://github.com/fonttools/fonttools/issues/2059

> (I don't have access to that issue)

wrong link sorry, this should be public: https://issues.skia.org/issues/338390594

> we have no way to determine "when it is necessary"

ok fair, but there are some obvious cases where it's clear the flag is not necessary, like some of those mentioned in the skia bug above, that still trigger the mentioned bug. Would be better if a user could at least override the global setting on a per glyph basis.

Anyway, mine was just a heads up.

PR #1471, opened 2024-06-18
Fix #1388: Icon hover animation misbehaves on Safari

### LINKED ISSUE #1388: CSS: Icon hover animation misbehaves on Safari
Our icon buttons have a subtle animation on hover: they get a little bit bigger from the center while hovering.

This looks and works nicely on Firefox and Chrome, but not on Safari: it seems the scale is done from the lower left corner of the icon instead of the center.

It gets even worse in the Transformations panel, where _all_ icons start wobbling when you hover one.

Our hover CSS may have been naïvely written, but I'm stumped as to how to make Safari behave better. I there anyone who is able to debug and fix this?

hwoongkang:
After spending hours trying to solve the problem, I asked to chatGPT, and the first option it provided solved the problem.

I attach the solutions it suggested: https://chatgpt.com/share/68cc3cc6-7e15-4e8b-a29c-6fa3baba5c57

Since [Mozilla documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change) of `will-change` recommends not to use `will-change` so often because it may cause too much resource consumption, I tried all the other solutions(animating only inner svg component,`transform-origin`, `backface-visibility`, `transformZ(0)`). However they all failed.

Thus I suggest adding `will-change: transform` to the style of `IconButton`.

This fixes #1388

### DIFF
--- a/src/fontra/client/web-components/icon-button.js
+++ b/src/fontra/client/web-components/icon-button.js
@@ -14,6 +14,7 @@ export class IconButton extends UnlitElement {
       height: 100%;
       transition: 150ms;
       cursor: pointer;
+      will-change: transform;
     }
 
     button:hover {


justvanrossum (approved):
Wow, that's an amazing one-line fix! I tried it and it is indeed effective.

Thank you for the analysis, and your other attempts. It almost sounds like it is a bug in Safari, and `will-change` nicely works around it.

PR #1472, opened 2024-06-18
Fixing winding direction of shape tool contour

### LINKED ISSUE #1455: [shape tool] Winding direction depends on dragging direction
For example:
- drag right/up -> contour is clockwise
- drag right/down -> contour is counter-clockwise

ollimeier:
Fixes #1455

I tested it with the rectangle as well as with the circle + ctrlKey. 
The 'default' winding direction is counter clockwise.

### DIFF
--- a/src/fontra/views/editor/edit-tools-shape.js
+++ b/src/fontra/views/editor/edit-tools-shape.js
@@ -139,9 +139,18 @@ export class ShapeToolRect extends BaseTool {
       height *= 2;
     }
 
-    return VarPackedPath.fromUnpackedContours(
+    const varPackedPath = VarPackedPath.fromUnpackedContours(
       this.getUnpackedContours(x, y, width, height)
     );
+
+    // fix contour direction if needed
+    if ((width > 0 && height > 0) || (width < 0 && height < 0)) {
+      // to make sure the winding direction is always the same,
+      // does not matter in which direction the rectangle is drawn
+      this.reversePath(varPackedPath);
+    }
+
+    return varPackedPath;
   }
 
   getUnpackedContours(x, y, width, height) {


justvanrossum:
An better solution would be to do 
```js
    mouseRect = rectangle.normalizeRect(rectangle.roundRect({ ... }))`
```
At line 53. This fixes the problem right at the source, and avoids having to possibly reverse twice in the case of ctrlKey.

It _could_ be that this makes the default winding direction the opposite to what it is now. But that could be fixed by inverting the ctrlKey reverse logic.

ollimeier:
> An better solution would be to do
> 
> ```js
>     mouseRect = rectangle.normalizeRect(rectangle.roundRect({ ... }))`
> ```
> 
> At line 53. This fixes the problem right at the source, and avoids having to possibly reverse twice in the case of ctrlKey.
> 
> It _could_ be that this makes the default winding direction the opposite to what it is now. But that could be fixed by inverting the ctrlKey reverse logic.

OK, works well. But in that case, the starting point is always at the same position (left bottom corner). 
I liked that you can have control over the starting point depending where you start dragging. But that's probably just me???

ollimeier:
@justvanrossum The proposed change causes some negative side effects to other functionalities like 'from center' or 'square'. Is there something I do wrong? Or should I rework other parts as well? Please see the following screen recording:

https://github.com/googlefonts/fontra/assets/21055547/6633c39a-baca-4e61-8680-cc91c7fe2e73

justvanrossum:
> I liked that you can have control over the starting point depending where you start dragging.

You are entirely right, I didn't think of that. Forget my suggestion: your solution is good as it is.

justvanrossum (approved):

Issue #603, opened 2023-07-05
Scale/transform tool for components


mrk-z:
When we get the scale/transform tool for outlines, it'd be great if it would also work for components: if the component has a "width" and/or a "height" axis, using the scale/transform tool on the component would enable changing the width/height by dragging, which would be more intuitive and quicker.

<img width="798" alt="Capture d’écran 2023-07-05 à 13 50 45" src="https://github.com/googlefonts/fontra/assets/56224289/15245054-d5de-4e56-8367-4ade86141e0c">

mrk-z:
This is not urgent. It came up during discussions at one point, so I thought I'd open up an issue.

justvanrossum:
Are you sure it should be integrated with a general purpose transform tool? I somehow feel this is very much a distinct operation from a "regular" height/width scale (which would be done by transforming the transformation in the case of components).

mrk-z:
Okay maybe it should be a separate tool. But I believe this would be more useful than the "regular" scaling for components, since regular scaling would distort the shapes.

Glyphs has this function, and in Glyphs, selecting a smart component automatically gives you the handles for scaling, no need to activate a new tool.

<img width="552" alt="Capture d’écran 2023-07-05 à 14 10 43" src="https://github.com/fontra/fontra/assets/56224289/bb5caafe-249a-445f-bdf8-2fdb76547962">

mrk-z:
Actually, yes. In Glyphs selecting an outline also automatically gives you the same bounding box. No extra steps to activate a transform tool.

<img width="428" alt="Capture d’écran 2023-07-05 à 14 57 38" src="https://github.com/fontra/fontra/assets/56224289/48b974c1-651f-40ea-ab07-cb6aca679c3d">

justvanrossum:
Is that bounding box default behavior, or does a plug in provide it? Or is it something you can turn on/off in as a preference?

mrk-z:
It's one of the default tools, not a plug-in. And yes you can turn it on/off under View -> Show Bounding Box

justvanrossum:
Relates to #339.

PR #1487, opened 2024-06-21
Interactive resizing of selection by selection bounds

### LINKED ISSUE #1368: [outline editing] interactive resizing of selection via bounding box controls
This should also implement https://github.com/googlefonts/fontra/issues/603 (tweak "width" and "height" component axes to match size if possible), but perhaps that is better done as a separate task.

ollimeier:
Fixes #1368
This also prepares for #603


src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -124,15 +170,28 @@ export class PointerTool extends BaseTool {
...
       const result = await this.handleDragSelection(eventStream, initialEvent);
       delete this.sceneController.sceneModel.initialClickedPointIndex;
       return result;
+    } else if (
+      resizeHandle &&
+      this.editor.visualizationLayersSettings.model["fontra.resize.selection"]
+    ) {
+      sceneController.selection = initialSelection;
+      this.sceneController.sceneModel.clickedResizeHandle = resizeHandle;
+      await this.handleDragSelectionBoundsResize(
```
justvanrossum, line 189: Please call this method `handleBoundsResize`. We're not dragging the selection, were resizing by bounds.

src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -36,10 +54,38 @@ export class PointerTool extends BaseTool {
...
+    } else {
+      this.setCursor();
+    }
   }
 
-  setCursor() {
+  setCursor(cursor = undefined) {
+    if (cursor) {
+      this.canvasController.canvas.style.cursor = cursor;
+      return;
```
justvanrossum, line 87: Please remove the return, and make the "if" two lines below an "else if".

src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -124,15 +170,28 @@ export class PointerTool extends BaseTool {
...
-    } else if (initiateDrag) {
+    } else if (initiateDrag && !resizeHandle) {
       this.sceneController.sceneModel.initialClickedPointIndex =
         initialClickedPointIndex;
       const result = await this.handleDragSelection(eventStream, initialEvent);
       delete this.sceneController.sceneModel.initialClickedPointIndex;
       return result;
+    } else if (
+      resizeHandle &&
+      this.editor.visualizationLayersSettings.model["fontra.resize.selection"]
```
justvanrossum, line 185: Please do this check earlier, at `const resizeHandle`:
```js
    const resizeHandle = this.editor.visualizationLayersSettings.model["fontra.resize.selection"]
      ? this.getResizeHandle(initialEvent, initialSelection)
      : undefined;
```
Or (maybe even better?), add the check to getResizeHandle().

src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -346,6 +405,179 @@ export class PointerTool extends BaseTool {
...
   }
+
+  async handleDragSelectionBoundsResize(selection, eventStream, initialEvent) {
+    const sceneController = this.sceneController;
+    const clickedResizeHandle = sceneController.sceneModel.clickedResizeHandle;
+
+    // The following may seem wrong to you, but it's correct.
+    // Because we say for example bottom-left and not left-bottom. Y-X order.
+    const resizeHandlePositionX = clickedResizeHandle.split("-")[1];
+    const resizeHandlePositionY = clickedResizeHandle.split("-")[0];
```
justvanrossum, line 416: You are repeating yourself too much: you call `.split("-")` twice. Easy fix:
```js
    const [resizeHandlePositionY, resizeHandlePositionX] = clickedResizeHandle.split("-");
```

src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -346,6 +405,179 @@ export class PointerTool extends BaseTool {
...
+
+        const deepEditChanges = [];
+        for (const layer of layerInfo) {
+          const layerGlyph = layer.layerGlyphController.instance;
+
+          // NOTE: calculate the scale based on selection width per layer.
+          let scaleX = (layer.selectionWidth + delta.x) / layer.selectionWidth;
+          let scaleY = (layer.selectionHeight + delta.y) / layer.selectionHeight;
+
+          if (clickedResizeHandle.includes("middle")) {
```
justvanrossum, line 482: This if block plus the two following if blocks should be if / else if / else if, no?

src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -346,6 +405,179 @@ export class PointerTool extends BaseTool {
...
+          if (event.shiftKey) {
+            if (scaleX > scaleY) {
+              scaleY = scaleX;
+            } else {
+              scaleX = scaleY;
+            }
+          }
+          let transformation = new Transform().scale(scaleX, scaleY);
+
+          let pinPoint = layer.regularPinPoint;
```
justvanrossum, line 507: Please write this as an `?` expression instead of overwriting a redundant assignment. Like this:
```js
   const pinPoint = event.altKey ? layer.altPinPoint : layer.regularPinPoint;
```
This style is almost always to be preferred.

justvanrossum (commented):
I left some comments.

src/fontra/views/editor/edit-tools-pointer.js
ollimeier: I am happy to change it, but have you heard of the 'return early' code concept? This was my intention of putting it at the top:
https://medium.com/swlh/return-early-pattern-3d18a41bba8

src/fontra/views/editor/edit-tools-pointer.js
ollimeier: Thanks. Done.

src/fontra/views/editor/edit-tools-pointer.js
justvanrossum: Of course I have, and I use it plenty.

In this case it is pretty much a switch between three cases. Two cases already existed, you added a third while not following the style of the code that was already there. Hence my feedback.

src/fontra/views/editor/edit-tools-pointer.js
ollimeier: Thanks. I moved it to getResizeHandle.

src/fontra/views/editor/edit-tools-pointer.js
ollimeier: done.

src/fontra/views/editor/edit-tools-pointer.js
ollimeier: Thanks. done.

src/fontra/views/editor/edit-tools-pointer.js
ollimeier: I would say no, because we have calculated scaleX and scaleY already before. An `else` would fallback to what we have already in line 479 and 480. But let me think about it a bit more.

src/fontra/views/editor/edit-tools-pointer.js
ollimeier: Thanks. done.

src/fontra/views/editor/edit-tools-pointer.js
justvanrossum: Aren't the three blocks mutually exclusive? If one "if" triggers, none of the others will. Or do I miss something?

justvanrossum: If "middle", there will not be "center", and the last shiftKey "if" only needs to be done if there is no "middle" and no "center".

ollimeier: Done + further code improvements

justvanrossum (commented):
Regarding, #1492, we actually have to address this here after all.

You call getSelectionBounds() as part of the draw code as well as on each mouse move. This is fundamentally too much computation for fast drawing, and does a lot of redundant work.

Instead, we should:
- subscribe to selection changes
- subscribe to glyph edits
- subscribe to change of edited-glyph

Then, use a throttled callback that will do the bounds computation, and store the computed bounds in a property of the scene model, and request a canvas update.

(Throttled: limit the number of callback calls per second, there's a helper func for that)

ollimeier:
Further note from Just:
We will separate:
- computation of the selection bounds
- using the selection bounds

Computation stores the current selection bounds in a scene model attribute. The rest of the code will use that attribute, unless it’s undefined.

justvanrossum (approved):

justvanrossum:
Maybe add those notes to #1492 instead.

PR #1516, opened 2024-07-03
Bugfix: Bounds Resize on component vs. info panel glitch

### LINKED ISSUE #1512: Bounds Resize on component vs. info panel glitch
- Have the info panel open
- Select a component
- Notice the selected component in the info panel
- Click on a resize handle, but don't drag yet (but don't release the mouse button yet)
- Notice the selected component in the info panel is gone, despite the fact that the component is still selected
- (Once you drag, the info panel correctly shows the selected component again)

This does not happen when normally dragging a component.

ollimeier:
Fixes #1512 

I moved the whole code for 'bounds resize' up to the top of `handleDrag`. If `resizeHandle` exists, there is no need to go through the rest of the code. This fixes the glitch with the component vs. info panel. 
Plus: It is more close to what it was before we added the 'resize bounds'-feature.

### DIFF
--- a/src/fontra/views/editor/edit-tools-pointer.js
+++ b/src/fontra/views/editor/edit-tools-pointer.js
@@ -95,6 +95,15 @@ export class PointerTool extends BaseTool {
 
   async handleDrag(eventStream, initialEvent) {
     const sceneController = this.sceneController;
+    const initialSelection = sceneController.selection;
+    const resizeHandle = this.getResizeHandle(initialEvent, initialSelection);
+    if (resizeHandle) {
+      sceneController.sceneModel.clickedResizeHandle = resizeHandle;
+      await this.handleBoundsResize(initialSelection, eventStream, initialEvent);
+      delete sceneController.sceneModel.clickedResizeHandle;
+      return;
+    }
+
     const point = sceneController.localPoint(initialEvent);
     const size = sceneController.mouseClickMargin;
     const { selection, pathHit } = this.sceneModel.selectionAtPoint(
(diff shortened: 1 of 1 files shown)


justvanrossum:
Hm, the disadvantage that I think this has (I haven't tried it yet) is that if a resize handle overlaps a (non-selected) point, now the resize handle "wins" when clicking, where the point would have won before. Can you verify that? If that's the case, I'm not quite comfortable with this solution.

ollimeier:
> Hm, the disadvantage that I think this has (I haven't tried it yet) is that if a resize handle overlaps a (non-selected) point, now the resize handle "wins" when clicking, where the point would have won before. Can you verify that? If that's the case, I'm not quite comfortable with this solution.

Yes, I can confirm this.

ollimeier:
> > Hm, the disadvantage that I think this has (I haven't tried it yet) is that if a resize handle overlaps a (non-selected) point, now the resize handle "wins" when clicking, where the point would have won before. Can you verify that? If that's the case, I'm not quite comfortable with this solution.
> 
> Yes, I can confirm this.

I double checked with the main branch and this behaviour is there from the beginning: the resize handle 'wins'. 
For me this does not feel unnatural. I see the bounds and if I want to get the point below the handle I can unselect and then click the point or hide the 'resize selection bounds' via the 'User Settings' panel. 
But I am happy to get some advice how to change it in any way.

justvanrossum:
Ok, thanks for checking. In that case, let's leave that behavior, and reconsider if people complain.

justvanrossum (approved):

PR #1521, opened 2024-07-03
Add context menu 'Close Contour'

### LINKED ISSUE #1515: "Close contour" function
The pen tool allows to close a contour, but it would be good to have a (contextual?) menu for that.

- If the end points of an open contour are selected, close the contour

Should it also work if the end points of multiple contours are selected?

Menu title: "Close contour" (possibly plural)
Shortcut: command-J (for "join")

ollimeier:
Fixes #1515 

I understood #1515 that start and end points should be selected. This check is done by `doCloseContour`. 

But I am wondering if I probably misunderstood the task? Because it would be possible to do it with any kind of selection, does not matter which point. At least one point (or many) should be selected -> then: close the contour. And if multiple contours are selected, close all of them? And If nothing is selected, close all open contours?

Thanks a lot in advance for your feedback.


justvanrossum:
Good point, maybe indeed the selection doesn't need to be so precise.

src/fontra/views/editor/scene-controller.js
```diff
@@ -641,6 +641,12 @@ export class SceneController {
 
   getContextMenuItems(event) {
     const contextMenuItems = [
+      {
+        title: translate("action.close-contour"),
+        enabled: () => this.doCloseContour(),
```
justvanrossum, line 646: The naming convention is (you can see that in all examples):
- enabled: canBlaBla
- callback: doBlaBla

justvanrossum: Btw. please look into `translatePlural`, as used with `action.decompose-component`. We need it here, too.

src/fontra/views/editor/scene-controller.js
```diff
@@ -637,10 +637,28 @@ export class SceneController {
       parseSelection(relevantSelection);
     this.contextMenuState.pointSelection = pointSelection;
     this.contextMenuState.componentSelection = componentSelection;
+
+    // relevantSelection does not include cmd + A selection,
```
justvanrossum, line 641: There is a bit of misunderstanding here: the rest of the context menus only get enable of you right-click _on_ part of the selection or on something selectable. If you click anywhere else, the context menu items will not be enabled.

So it's not that "relevantSelection does not include cmd-A selection", it totally does, as long as you right-click on something that is selected or is selectable.

This is by design. We could argue about that design, but that's how all other context menu items here behave, so this one should also behave like that.

This means you can get rid of the additional parseSelection call, and just use pointSelection instead of pointIndices.

justvanrossum (commented):
Mostly looks good and works well, thanks.

Edge cases to solve:
- it should not close a path that consists of a single point
- perhaps it should not close a path that only has a single _on-curve_ point
- if you draw a path, and there's a "dangling" off curve point, closing this path will cause a quadratic segment to be added. This is undesirable unless we are in quad drawing mode. The pen tool, in a similar situation, will convert such a quad segment to cubic.

src/fontra/views/editor/scene-controller.js
justvanrossum: Sorry about the first comment: it missed that that convention is only held in editor.js, not elsewhere, for better or for worse.

src/fontra/views/editor/scene-controller.js
ollimeier: Thanks for the explanation. I was not aware of that. I changed it like suggested.

ollimeier:
> Mostly looks good and works well, thanks.
> 
> Edge cases to solve:
> 
> * it should not close a path that consists of a single point
> * perhaps it should not close a path that only has a single _on-curve_ point
> * if you draw a path, and there's a "dangling" off curve point, closing this path will cause a quadratic segment to be added. This is undesirable unless we are in quad drawing mode. The pen tool, in a similar situation, will convert such a quad segment to cubic.

Thank you very much for the valid feedback. I managed to find a solution for all mentioned edge cases.

src/fontra/views/editor/scene-controller.js
```diff
@@ -1148,6 +1168,49 @@ export class SceneController {
...
+        const path = layerGlyph.path;
+        for (const contourIndex of openContours) {
+          // close open contour
+          path.contourInfo[contourIndex].isClosed = true;
+          if (curveType === "cubic") {
+            this.closeContourEnsureCubicOffCurves(path, contourIndex);
+          }
+        }
+      }
+      return "Close Contour" + (openContours.length > 1 ? "s" : "");
```
justvanrossum, line 1185: Not sure if we do this elsewhere, but why not use the same translation as for the menu item.

src/fontra/views/editor/scene-controller.js
```diff
@@ -1148,6 +1168,49 @@ export class SceneController {
...
+    });
+  }
+
+  closeContourEnsureCubicOffCurves(path, contourIndex) {
```
justvanrossum, line 1189: This should be a helper function, not a method. It does not depend on any state of the `this` object, it doesn't even reference `this`.

(My general rule-of-thumb: if it doesn't use `this`, most likely it should not be a method.)

src/fontra/views/editor/scene-controller.js
```diff
@@ -1333,6 +1396,32 @@ function getSelectedContours(path, pointSelection) {
...
+    return [];
+  }
+  const selectedContours = new Set();
+  for (const contourIndex of getSelectedContours(path, pointSelection)) {
+    if (path.contourInfo[contourIndex].isClosed) {
+      // skip if contour is closed already
+      continue;
+    }
+    const contour = path.getContour(contourIndex);
+    const numOnCourvePoints = contour.pointTypes.filter((x) => x == 0).length;
```
justvanrossum, line 1414: It is wasteful to build an array and then only use it for its length. You can do this with `.reduce()` instead.

Typo: `numOnCourvePoints` -> `numOnCurvePoints`

src/fontra/views/editor/scene-controller.js
```diff
@@ -1148,6 +1168,49 @@ export class SceneController {
...
+    const prevEndPoint = path.getPoint(contourInfo.endPoint - 1);
+    const endPoint = path.getPoint(contourInfo.endPoint);
+    const startPoint = path.getPoint(contourInfo.endPoint - numContourPoints + 1);
+
+    if (prevEndPoint.type || !endPoint.type || startPoint.type) {
+      // Sanity check: we expect on-curve/off-curve/on-curve
+      return;
+    }
+
+    // Compute handles for a cubic segment that will look the same as the
```
justvanrossum, line 1202: This code currently only works if the dangling point is at the end. But it should also work if it's at the start. Reverse the contour then "close contour" should work just the same.

src/fontra/views/editor/scene-controller.js
```diff
@@ -1148,6 +1168,49 @@ export class SceneController {
     });
   }
 
+  async closeContour() {
```
justvanrossum, line 1171: Somehow closing doesn't work when the open path starts with a curve:
image
Or maybe it's the smooth flag that prevents it to work.

justvanrossum (commented):
More edge cases :)

src/fontra/views/editor/scene-controller.js
ollimeier: Thanks. Done.

src/fontra/views/editor/scene-controller.js
ollimeier: thanks. done.

src/fontra/views/editor/scene-controller.js
ollimeier: thanks and thanks. done.

justvanrossum:
General comment: to check if a point is on-curve, you have to do a little more than comparing with `0`, as this doesn't work for "smooth" points, which are on-curve, yet have the smooth bit flag set, so the pointType isn't zero. You have to use `VarPackedPath.POINT_TYPE_MASK` and `VarPackedPath.ON_CURVE`, for example like this:

```js
    if ((pointType & VarPackedPath.POINT_TYPE_MASK) === VarPackedPath.ON_CURVE) {
      // on-curve
    } else {
      // off-curve
    }
```

src/fontra/views/editor/scene-controller.js
ollimeier: I fixed it, with the help of the below comment.

ollimeier:
> General comment: to check if a point is on-curve, you have to do a little more than comparing with `0`, as this doesn't work for "smooth" points, which are on-curve, yet have the smooth bit flag set, so the pointType isn't zero. You have to use `VarPackedPath.POINT_TYPE_MASK` and `VarPackedPath.ON_CURVE`, for example like this:
> 
> ```js
>     if ((pointType & VarPackedPath.POINT_TYPE_MASK) === VarPackedPath.ON_CURVE) {
>       // on-curve
>     } else {
>       // off-curve
>     }
> ```

This fixed the other issue.

src/fontra/views/editor/scene-controller.js
ollimeier: Great edge cases. done.

src/fontra/views/editor/scene-controller.js
```diff
@@ -1333,6 +1371,75 @@ function getSelectedContours(path, pointSelection) {
...
+  const contourInfo = path.contourInfo[contourIndex];
+  const numContourPoints = path.getContour(contourIndex).pointTypes.length;
+
+  const prevEndPoint = path.getPoint(contourInfo.endPoint - 1);
+  const endPoint = path.getPoint(contourInfo.endPoint);
+  const startPoint = path.getPoint(contourInfo.endPoint - numContourPoints + 1);
+
+  if (startPoint.type && !endPoint.type) {
+    // contour starts with off-curve and ends with on-curve point,
+    // reverse the contour, first.
```
justvanrossum, line 1417: You can't just reverse the winding direction behind the user's back... Please find a better solution.

ollimeier: OK, I am happy to find a different solution. But I am now wondering if I misunderstood your feedback? Because you wrote this: _Reverse the contour then "close contour" should work just the same._

justvanrossum: Ah, I can see how that can be misinterpreted. What I meant was: "if you have an open contour with a dangling off curve point at the end, and then you (the user) reverse it, 'close open contour' should also work". That was a roundabout way of saying "if the dangling off-curve point is at the _start_ of the open contour, 'close open contour' should also work."

ollimeier: Makes of cause totally sense. Sorry for that misunderstanding. 
I pushed a solution. Here a short demo of different open paths:

https://github.com/googlefonts/fontra/assets/21055547/5de96654-8b8c-4cb4-9eca-c05e9b280270

I won't close this review/feedback, because I am not sure if you will like the solution.

src/fontra/views/editor/scene-controller.js
```diff
@@ -1333,6 +1371,73 @@ function getSelectedContours(path, pointSelection) {
...
+      continue;
+    }
+    selectedContours.add(contourIndex);
+  }
+
+  return [...selectedContours];
+}
+
+function closeContourEnsureCubicOffCurves(path, contourIndex) {
+  const contourInfo = path.contourInfo[contourIndex];
```
justvanrossum, line 1408: You can use `path.getContourPoint(contourIndex, contourPointIndex)` to get points more conveniently. `contourPointIndex` can be negative to get points from the end.

There is also `path.getNumPointsOfContour(contourIndex)` to more conveniently get number of points of a contour.

Your "situation" variable can be named better. We have two situations:
- dangling off-curve at the end
- dangling off-curve at the start
Perhaps "offCurveAtEnd" or "offCurveAtStart" is a better name?

ollimeier: Thank you very much. done.

src/fontra/views/editor/scene-controller.js
```diff
@@ -1333,6 +1371,70 @@ function getSelectedContours(path, pointSelection) {
...
+  // one-off-curve quad segment we have.
+  const [handle1, handle2] = [firstPoint, lastPoint].map((point) => {
+    return {
+      ...vector.roundVector(scalePoint(point, middlePoint, 2 / 3)),
+      type: "cubic",
+    };
+  });
+
+  const numContourPoints = path.getNumPointsOfContour(contourIndex);
+  const handle1ContourIndex = offCurveAtStart ? 0 : numContourPoints - 1;
```
justvanrossum, line 1433: What you call `handle1ContourIndex` is in fact a contour point index.

Btw. setContourPoint and insertPoint also take negative point indices, so I'm 99% sure you don't need the numContourPoints here.

Ah, but you want to append at the end of the contour, which won't work with a negative index. However, there is path.appendPoint().

ollimeier: Thank you very much for your caring feedback. I removed numContourPoints and used path.appendPoint().

justvanrossum (commented):
I found one more edge case: I don't think an open path that only has two on-curve points (a straight line without handles) should be closable.

ollimeier:
> I found one more edge case: I don't think an open path that only has two on-curve points (a straight line without handles) should be closable.

Thanks once again. 
I am not sure about the changes I have made, because first I made 'getSelectedOpenContours' as general as possible, so we may can recuse it someday (at least that was my intention), but now it includes already two edge cases. I am not sure, if this still a good idea or if we may should change the name of the function.
I would much appreciate your thoughts on that. Thanks a lot in advance.

src/fontra/views/editor/scene-controller.js
```diff
@@ -1333,6 +1371,70 @@ function getSelectedContours(path, pointSelection) {
...
+    const contour = path.getContour(contourIndex);
+    const numOnCurvePoints = contour.pointTypes.reduce(
+      (acc, pointType) =>
+        acc +
+        ((pointType & VarPackedPath.POINT_TYPE_MASK) === VarPackedPath.ON_CURVE
+          ? 1
+          : 0),
+      0
+    );
+    if (numOnCurvePoints === 1) {
```
justvanrossum, line 1398: Better safe than sorry:
```suggestion
    if (numOnCurvePoints <= 1) {
```

PR #1528, opened 2024-07-04
Fix: getResizeHandle error on incompatible source

### LINKED ISSUE #1524: PointerTool.getResizeHandle can error on incompatible source
Reproduce:
- go to a glyph source
- select all
- delete (makes the glyph incompatible)
- go to one another source, that still has an outline
- select all
- go to the source where you deleted the outline
- error:
image

ollimeier:
Fixes #1524 

The issues lives more deep into 
`getSelectionBounds` -> 
`filterPathByPointIndices` -> 
`getSelectionByContour` -> 
`getContourAndPointIndex`, 
when the selection does not match with the actual path/components. 
I first thought it would be good to fix it somewhere in getSelectionBounds, but probably this would change the behaviour too much???

Why I added a try/catch to getResizeBounds. 
I am happy to do it differently if you have an idea how to fix it in more proper way.

### DIFF
--- a/src/fontra/client/core/glyph-controller.js
+++ b/src/fontra/client/core/glyph-controller.js
@@ -615,9 +615,12 @@ export class StaticGlyphController {
 
     const selectionRects = [];
     if (pointIndices.length) {
-      const path = filterPathByPointIndices(this.instance.path, pointIndices);
-      if (path.getBounds()) {
-        selectionRects.push(path.getBounds());
+      const pathBounds = filterPathByPointIndices(
+        this.instance.path,
+        pointIndices
+      ).getBounds();
+      if (pathBounds) {
+        selectionRects.push(pathBounds);
       }
     }
 
--- a/src/fontra/client/core/glyph-controller.js
+++ b/src/fontra/client/core/glyph-controller.js
@@ -616,7 +616,7 @@ export class StaticGlyphController {
     const selectionRects = [];
     if (pointIndices.length) {
       const path = filterPathByPointIndices(this.instance.path, pointIndices);
-      if (path) {
+      if (path.getBounds()) {
         selectionRects.push(path.getBounds());
       }
     }
--- a/src/fontra/client/core/path-functions.js
+++ b/src/fontra/client/core/path-functions.js
@@ -626,11 +626,14 @@ export function scalePoint(pinPoint, point, factor) {
 export function getSelectionByContour(path, pointIndices) {
   const selectionByContour = new Map();
   for (const pointIndex of pointIndices) {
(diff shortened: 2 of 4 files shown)


justvanrossum:
Hm, I think it is better to fix it deeper.

One way would be to add a check to the `pointIndex` loop in `getSelectionByContour()` so that `pointIndex` doesn't exceed `path.numPoints`.

src/fontra/client/core/path-functions.js
```diff
@@ -626,11 +626,14 @@ export function scalePoint(pinPoint, point, factor) {
 export function getSelectionByContour(path, pointIndices) {
   const selectionByContour = new Map();
   for (const pointIndex of pointIndices) {
-    const [contourIndex, contourPointIndex] = path.getContourAndPointIndex(pointIndex);
-    if (!selectionByContour.has(contourIndex)) {
-      selectionByContour.set(contourIndex, []);
+    if (pointIndex !== undefined && pointIndex < path.numPoints) {
```
justvanrossum, line 629: I don't think `pointIndex` can be undefined.

But please rewrite in this form:

```js
    if (condition not met) {
      continue;
    }
```

src/fontra/client/core/glyph-controller.js
```diff
@@ -616,7 +616,7 @@ export class StaticGlyphController {
     const selectionRects = [];
     if (pointIndices.length) {
       const path = filterPathByPointIndices(this.instance.path, pointIndices);
-      if (path) {
+      if (path.getBounds()) {
```
justvanrossum, line 619: You're calling `path.getBounds()` twice, which is wasteful. Please use a variable to prevent this.

src/fontra/client/core/path-functions.js
ollimeier: About `pointIndex` == 'undefined': I copied it from here:
https://github.com/googlefonts/fontra/blob/31cffdbd507d027c9724512c1f36b4d6aa21eb6d/src/fontra/views/editor/edit-tools-pen.js#L623

ollimeier: I removed the `pointIndex !== undefined` + `continue` if `pointIndex >= path.numPoints`

src/fontra/client/core/glyph-controller.js
ollimeier: Thanks. done.

justvanrossum (approved):

src/fontra/client/core/path-functions.js
justvanrossum: > I copied it from here

Why, though? That situation is not comparable at all. Please don't blindly copy/paste/copilot.

PR #1536, opened 2024-07-09
Fix #1523: Change context menu positioning strategy to avoid edge case

### LINKED ISSUE #1523: Literal edge case for the context menu: control-click can immediately trigger unwanted menu
Normally, when you right-click or control-click, the context menu will appear next to the pointer.

However, if you control-click / right-click near the right edge or bottom edge, the pointer will be somewhere inside the contextual menu.

Normally, when you right-click / control-click and release the pointer, the menu stays up and you can navigate around to find the menu item you need.

However, in the edge case, this will _immediately_ choose the item the pointer happens to be over. This is clearly undesirable.

hwoongkang:
If you see the code below, before, if the calculated bounding rect of the context menu exceeded the bounding rect of the canvas, the strategy was to modify x only by the "exceeding amount":

```Javascript
if (thisRect.right > containerRect.right) {
        position.x -= thisRect.right - containerRect.right + 2;
        // ...
}
```
More clearyly, when you view the positioning of the context menu, "the right margin" was fixed: if you click the menu close enough from the right border of the canvas, the context menu is always positioned 2 px apart from the right border of the canvas.

So is the case with the bottom positioning too.

That means if you click more than 2px apart, but still close enough from the right border, it can end up making your pointer located inside the context menu.

Therefore I made the context menu appear on the "left side" of a pointer by adjusting its position by its width. And since the "original right side" position was 1px to the right and 1px to the bottom, I adjusted 2 more pixels to make it 1px to the left of the pointer.

This fixes #1523

### DIFF
--- a/src/fontra/client/web-components/menu-panel.js
+++ b/src/fontra/client/web-components/menu-panel.js
@@ -208,7 +208,7 @@ export class MenuPanel extends SimpleElement {
       const containerRect = this.positionContainer.getBoundingClientRect();
       const thisRect = this.getBoundingClientRect();
       if (thisRect.right > containerRect.right) {
-        position.x -= thisRect.right - containerRect.right + 2;
+        position.x -= thisRect.width + 2;
       }
       if (thisRect.bottom > containerRect.bottom) {
         position.y -= thisRect.bottom - containerRect.bottom + 2;


hwoongkang:
Actually by this fix, I wanted to fix the problem by the smallest amount of change, but I have another idea too.

I couldn't find the code firing "contextmenu" custom event, but it seems like opening context menu happens on "pointerdown".

But since handling the callbacks of menu item happens on "pointerup", its where the original issue happens: You open the menu, pointer is inside the menu, and it triggers the menu item as soon as let pointer go.

So, if its preferable to maintain the positioning strategy of the context menu --clipping the menu to the rightmost side--, we may introduce a auxiliary flag system to notify the menu that it "just opened" and should not trigger the menu items.

justvanrossum:
I think your solution is good, thank you! I will merge it.

Currently, we keep the context menu tightly within the canvas area. But if the right sidebar is visible, there isn't really a reason why it couldn't go on top of the sidebar. Feel free to propose further improvements.

PR #1551, opened 2024-07-18
Code quality: use model.shapeToolCtrlKey instead of model.event.ctrlKey

### LINKED ISSUE #1548: [shape tool] Code improvements
In edit-tools-shape.js, the code assigns a vale to the scene model with the name "event". First off, this is too generic: any tool that writes (private) data to the scene model should use a prefix. In this case, probable "shapeTool". Secondly, the "event" that is being assigned is only used for its ctrlKey flag, so I suggest to use that instead of assigning the whole event.

ollimeier:
Fixes #1548

### DIFF
--- a/src/fontra/views/editor/edit-tools-shape.js
+++ b/src/fontra/views/editor/edit-tools-shape.js
@@ -59,12 +59,12 @@ export class ShapeToolRect extends BaseTool {
       const drawPath = new Path2D();
       this.drawShapePath2D(drawPath, mouseRect, event);
       this.sceneModel.shapeToolShapePath = drawPath;
-      this.sceneModel.shapeToolCtrlKey = event.ctrlKey;
+      this.sceneModel.shapeToolShowFill = event.ctrlKey;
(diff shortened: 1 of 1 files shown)


src/fontra/views/editor/edit-tools-shape.js
```diff
@@ -59,12 +59,12 @@ export class ShapeToolRect extends BaseTool {
       const drawPath = new Path2D();
       this.drawShapePath2D(drawPath, mouseRect, event);
       this.sceneModel.shapeToolShapePath = drawPath;
-      this.sceneModel.event = event;
+      this.sceneModel.shapeToolCtrlKey = event.ctrlKey;
```
justvanrossum, line 62: I would name this after what this flag is supposed to do, say `.shapeToolShowFill`.

ollimeier: I changed it.

justvanrossum (approved):

PR #1560, opened 2024-07-19
Improve UX of interactive resize tool

### LINKED ISSUE #1556: Resize transformation box UX improvements (drag off-curve)
From @JeremieHornus 
_another note: I find i slightly annoying to see the transformation box popping out every time I drag an off-curve point_

https://github.com/user-attachments/assets/afe65e4f-d04a-4db6-bbb4-1d20f482ebb8

ollimeier:
Fixes #1556

### DIFF
--- a/src/fontra/views/editor/edit-tools-pointer.js
+++ b/src/fontra/views/editor/edit-tools-pointer.js
@@ -54,6 +54,7 @@ export class PointerTool extends BaseTool {
       sceneController.hoveredGlyph = this.sceneModel.glyphAtPoint(point);
     }
 
+    this.sceneController.sceneModel.showResizeSelection = true;
     const resizeHandle = this.getResizeHandle(event, sceneController.selection);
     if (this.sceneController.sceneModel.hoverResizeHandle != resizeHandle) {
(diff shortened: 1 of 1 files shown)


src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -293,6 +293,11 @@ export class PointerTool extends BaseTool {
   }
 
   async handleDragSelection(eventStream, initialEvent) {
+    const resizeSelectionViewStatus =
```
justvanrossum, line 296: Please open a second window, look at the user settings, then go back to the first and drag some points. Observe the "resize selection" checkbox go on and off.

I don't think this is the best way to do it. Isn't there a property you already use for the bounds resize handles that you can temporarily reset?

ollimeier: Oh damn. Sorry. I totally ignored that. You're absolutely right. 
Sadly there is not property I can temporarily reset, but I added one. I hope that's better?!
Thanks for your quick feedback.

justvanrossum (approved):

PR #1561, opened 2024-07-19
Adding rotation to interactive resize tool

### LINKED ISSUE #1486: Followup resize transformation: include rotation
As a follow up + up for discussion:

Do we want to add rotation functionality to resize transformation? 
If so, maybe rename 'resize' to 'transform'? Or any other suggestions?

The code for the hover cursor may look like that:

```
  setCursorForResizeHandle(handleName, rotationKey = undefined) {
    if (handleName && rotationKey) {
      this.setCursor("url('/tabler-icons/rotate.svg') 12 12, auto");
      return;
    }
```

And the code for the rotation transformation may look like that:
```
          // rotation based on pinPoint
          if (event.metaKey) {
            const angleInitial = Math.atan2(
              pinPoint.y - initialPoint.y,
              pinPoint.x - initialPoint.x
            );
            const angle = Math.atan2(
              pinPoint.y - currentPoint.y,
              pinPoint.x - currentPoint.x
            );
            // Do not do scale, because together with rotation it will be a mess
            transformation = new Transform().rotate(angle - angleInitial);
          }
```

Useful links:
https://helpx.adobe.com/indesign/using/transforming-objects.html#transform_objects_with_the_free_transform_tool
https://www.youtube.com/watch?v=HRbvuv_baUo

ollimeier:
Fixes #1486 

As discussed:
- I created some custom rotation cursor icons as SVG for each handle.
- The metaKey is not in use anymore. Instead, the user need to get a little bit off the handle to get the rotation icon.
- I renamed a lot from 'resize' to transformation. 

+ I also added 'shift constrain' for rotation, so it snaps to 45 degrees.


ollimeier:
@JeremieHornus Here is a 'work in progress' preview of the rotation:

https://github.com/user-attachments/assets/0e4c0799-8d90-41a5-ae24-4d6778f8d4e5

You can rotate a selection using the MetaKey (the cmd key on the Mac).
As written in the description above, there are still a lot of open questions that I have to discuss with Just, but I wanted to give you a little teaser so that you can be curious ;)

justvanrossum:
I don't think a modifier key is the solution. Also: the meta key doesn't exist on Windows, although we could use the control key in that case (as we do with general shortcuts).

However, I think a solution that doesn't require a keystroke is possible. Let's discuss tomorrow.

ollimeier:
@justvanrossum Here a short demo of how it looks currently:

https://github.com/user-attachments/assets/29617d50-3813-4627-94fe-1ca6634c4081

src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -54,19 +55,45 @@ export class PointerTool extends BaseTool {
...
+      this.setCursorForRotationHandle(rotationHandle);
+    } else if (resizeHandle) {
       this.setCursorForResizeHandle(resizeHandle);
     } else {
       this.setCursor();
     }
   }
 
+  setCursorForRotationHandle(handleName) {
+    if (handleName === "bottom-left") {
```
justvanrossum, line 76: If you'd name the cursors according to the handle names (I mean, why not???), then the below could be simplified to:

```js
    const cursorPositions = {
      "bottom-left": "14 14",
      "bottom-right": "16 16",
      "top-left": "16 12",
      "top-right": "16 10",
      "middle-left": "20 14",
      "middle-right": "12 14",
      "top-center": "16 18",
      "bottom-center": "16 10",
    }
    this.setCursor(`url('/images/cursor-rotate-${handleName}.svg') ${cursorPositions[handleName]}, auto`);
```

icons/export-icons-as-svg.py
```diff
@@ -10,9 +10,9 @@ def numToString(number):
     return str(otRound(number))
 
 
-def makeSVG(pathString, width, height):
+def makeSVG(pathString, width, height, scaling=""):
```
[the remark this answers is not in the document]
ollimeier: flake8 failed with `f-string is missing placeholders` why I removed the f for the first line.
Done.

src/fontra/views/editor/edit-tools-pointer.js
ollimeier: My first naming orientation was based on icon-tabler names. But I am totally with you. I'll change it in a bit.

ollimeier: Thanks. Done.

justvanrossum (commented):
The rotation icons don't work (well) in dark mode. I think they need a white under layer. Not sure if/how we could integrate that into the ufo-based custom icon toolchain.

justvanrossum (commented):
Multi-source rotation: the angle of rotation should be determined by the edit layer. Right now all layers rotate at different angles, which makes no sense at all.

ollimeier:
> The rotation icons don't work (well) in dark mode. I think they need a white under layer. Not sure if/how we could integrate that into the ufo-based custom icon toolchain.

I added a compromise with adding a white `stroke` to the svg.

icons/fontra-icons.ufo/glyphs/_cursor-rotate-temp.glif
```diff
@@ -0,0 +1,95 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<glyph name="_cursor-rotate-temp" format="2">
```
justvanrossum, line 2: Is this glyph still needed? It is named "temp" and doesn't seem to be referenced as a component.

ollimeier: I decomposed the original arrow, to be able to add a white stroke for dark mode. I wanted to keep the original outline with overlapping/separate outlines for possible design modifications why I added a `cursor-rotate-temp`with an underscore at the beginning, so it gets ignored via the python script. I saw that underscore in a different glyph and thought it's a possible way to keep that glyph for reference without exporting it.

justvanrossum:
> I added a compromise with adding a white stroke to the svg.

Why is it a compromise?

I find the white stroke a little too thin at the moment. Also: you matched the design to the other icons, but here we have to blend in with system cursors. I find the soft corners out of place.

We can make a special function that creates SVG for cursors. Say "makeCursorSVG", then the original makeSVG can stay unchanged. What you can then do is layer the same path twice: once with a white stroke, and once without stroke in black. This gives us full control over the black shape, as well as the stroke thickness.

ollimeier:
> Multi-source rotation: the angle of rotation should be determined by the edit layer. Right now all layers rotate at different angles, which makes no sense at all.

I missed it, thanks a lot for finding it. I pushed a solution.

ollimeier:
> > I added a compromise with adding a white stroke to the svg.
> 
> Why is it a compromise?
> 
> I find the white stroke a little too thin at the moment. Also: you matched the design to the other icons, but here we have to blend in with system cursors. I find the soft corners out of place.
> 
> We can make a special function that creates SVG for cursors. Say "makeCursorSVG", then the original makeSVG can stay unchanged. What you can then do is layer the same path twice: once with a white stroke, and once without stroke in black. This gives us full control over the black shape, as well as the stroke thickness.

Great idea. done.

icons/export-icons-as-svg.py
```diff
@@ -12,9 +12,18 @@ def numToString(number):
...
+        + f'viewBox="0 0 {width} {height}">'
+        + f'<path d="{pathString}"/></svg>\n'
+    )
+
+
+def makeCursorSVG(pathString, width, height):
+    return (
+        '<svg xmlns="http://www.w3.org/2000/svg" width="32" height="32" '
+        + f'viewBox="0 0 {width} {height}">'
+        + f'<path stroke="rgb(255,255,255)" stroke-width="55" d="{pathString}"/>'
```
justvanrossum, line 25: To be good web citizens, we should avoid outputting the path twice. This should work (untested):

```
<svg ... >
<defs><path id="icon-path" d="{pathString}"/></defs>
<use href="#icon-path" stroke="rgb(...)" stroke-width="55"/>
<use href="#icon-path"/>
</svg>
```

https://developer.mozilla.org/en-US/docs/Web/SVG/Element/defs#example

justvanrossum:
There's one thing I keep forgetting to mention: if you hover over a resize corner, then move outside towards the "rotation" area, the cursor briefly goes back to normal before we reach the rotation area. So it seems there's a gap between the two, which is a little distracting.

icons/export-icons-as-svg.py
```diff
@@ -38,4 +47,10 @@ def makeSVG(pathString, width, height):
     glyph.draw(TransformPen(pen, (1, 0, 0, -1, 0, 800)))
     svgPath = pen.getCommands()
     iconPath = imagesDir / f"{iconName}.svg"
-    iconPath.write_text(makeSVG(svgPath, glyph.width, 1000))
+
+    # For cursors, the maximum size seems to be 32:
```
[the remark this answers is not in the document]
ollimeier: right. done.

icons/export-icons-as-svg.py
```diff
@@ -38,4 +47,10 @@ def makeSVG(pathString, width, height):
...
+    # https://stackoverflow.com/questions/6648279/cursor-256x256-px-size#answer-6648759
+    if iconName.startswith("cursor"):
```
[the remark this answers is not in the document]
ollimeier: done.

ollimeier:
> There's one thing I keep forgetting to mention: if you hover over a resize corner, then move outside towards the "rotation" area, the cursor briefly goes back to normal before we reach the rotation area. So it seems there's a gap between the two, which is a little distracting.

One reason for that is, that the distance of a corner is more far away than the distance on the sides, because the diagonale of the square is greater than the width or height. 
As a demonstration of the issue I visualized the 'handles' for resize + rotation. This makes it obvious, that the gap in the corner is greater than on the side.

Screenshot 2024-07-25 at 13 12 09

justvanrossum:
> One reason for that is

Let's focus on a solution. Remember what I sketched during our call earlier this week? 

This is an even worse rendering of that idea:

image

Circle: resize handle
Square: rotate handle
Where they overlap: resize

You can use pointInRect for the square.

ollimeier:
> > One reason for that is
> 
> Let's focus on a solution. Remember what I sketched during our call earlier this week?
> 
> This is an even worse rendering of that idea:
> 
> image
> 
> Circle: resize handle Square: rotate handle Where they overlap: resize
> 
> You can use pointInRect for the square.

OK, I'll give it a try.

icons/export-icons-as-svg.py
ollimeier: done

src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -25,10 +26,13 @@ import { equalGlyphSelection } from "./scene-controller.js";
...
   registerVisualizationLayerDefinition,
   strokeRoundNode,
+  strokeSquareNode,
 } from "./visualization-layer-definitions.js";
 import { copyComponent } from "/core/var-glyph.js";
 
-const handleMarginValue = 10;
+const transformationHandleMargin = 6;
+const transformationHandleSize = 8;
+const rotationHandleSizeFactor = 1.5;
```
justvanrossum, line 35:
```suggestion
const rotationHandleSizeFactor = 1.2;
```
For me, this feels more balanced between resize and rotate.

PR #1570, opened 2024-07-24
"Close contours" vs. join ends of two open contours

### LINKED ISSUE #1566: "Close contours" vs. join ends of two open contours
The "Close contours" context menu currently closes any contour that has at least one point selected.

We also need to be able to join two separate open paths, by selecting an end point of each contour.

I suggest to extend the feature like this:

- If the selection is exactly two points, and each are the end points of different open contours, then join these contours
- Else: do current behavior

The title of the context menu may have to differentiate between these cases. Say "Close contours" vs "Join contours".

ollimeier:
Fixes #1566

### DIFF
--- a/src/fontra/views/editor/scene-controller.js
+++ b/src/fontra/views/editor/scene-controller.js
@@ -1532,7 +1532,7 @@ export function joinContours(path, firstPointIndex, secondPointIndex) {
   );
   assert(
     path.isStartOrEndPoint(firstPointIndex) && path.isStartOrEndPoint(secondPointIndex),
-    "firstPointIndex and/or secondPointIndex is not a start or end point, but both must be start and/or end point."
+    "firstPointIndex and secondPointIndex must be start or end points"
   );
   assert(
     firstPointIndex < secondPointIndex,
--- a/src/fontra/views/editor/scene-controller.js
+++ b/src/fontra/views/editor/scene-controller.js
@@ -1528,7 +1528,7 @@ export function joinContours(path, firstPointIndex, secondPointIndex) {
   let selectedPointIndices = [];
   assert(
     firstPointIndex != secondPointIndex,
-    "firstPointIndex and secondPointIndex are the same, but must be different."
+    "firstPointIndex and secondPointIndex must be different"
   );
   assert(
     path.isStartOrEndPoint(firstPointIndex) && path.isStartOrEndPoint(secondPointIndex),
@@ -1543,6 +1543,11 @@ export function joinContours(path, firstPointIndex, secondPointIndex) {
     path.getContourAndPointIndex(firstPointIndex);
   const [secondContourIndex, secondContourPointIndex] =
     path.getContourAndPointIndex(secondPointIndex);
+
+  assert(
+    firstContourIndex != secondContourIndex,
+    "firstContourIndex and secondContourIndex must be different"
+  );
 
   let firstContour = path.getUnpackedContour(firstContourIndex);
   let secondContour = path.getUnpackedContour(secondContourIndex);
--- a/src/fontra/views/editor/scene-controller.js
+++ b/src/fontra/views/editor/scene-controller.js
@@ -1552,7 +1552,8 @@ export function connectTwoDistinctContours(path, firstPointIndex, secondPointInd
   if (!firstContourPointIndex) {
     [firstContour, secondContour] = [secondContour, firstContour];
   }
-
+  let selectedContourPointIndex1 = firstContour.points.length - 1;
+  let selectedContourPointIndex2 = selectedContourPointIndex1 + 1;
   let loneCubicHandle;
   const lastPointFirstContour = firstContour.points.at(-1);
   if (lastPointFirstContour.type) {
@@ -1575,6 +1576,7 @@ export function connectTwoDistinctContours(path, firstPointIndex, secondPointInd
     );
     firstContour.points.push(handle1);
     firstContour.points.push(handle2);
+    selectedContourPointIndex2 += 1;
   }
 
   const newContour = {
@@ -1585,6 +1587,13 @@ export function connectTwoDistinctContours(path, firstPointIndex, secondPointInd
   path.deleteContour(firstContourIndex);
   path.insertUnpackedContour(firstContourIndex, newContour);
   path.deleteContour(secondContourIndex);
+
+  selectedPointIndices.push(
+    path.getAbsolutePointIndex(firstContourIndex, selectedContourPointIndex1)
+  );
+  selectedPointIndices.push(
+    path.getAbsolutePointIndex(firstContourIndex, selectedContourPointIndex2)
+  );
 
   return selectedPointIndices;
 }
--- a/src/fontra/views/editor/scene-controller.js
+++ b/src/fontra/views/editor/scene-controller.js
@@ -1525,6 +1525,22 @@ function splitLocation(location, glyphAxes) {
 
 export function joinContours(path, firstPointIndex, secondPointIndex) {
   let selectedPointIndices = [];
+  if (firstPointIndex === secondPointIndex) {
+    throw new Error(
+      "firstPointIndex and secondPointIndex are the same, but must be different."
+    );
+  }
+  if (
+    !path.isStartOrEndPoint(firstPointIndex) ||
+    !path.isStartOrEndPoint(secondPointIndex)
+  ) {
+    throw new Error(
+      "firstPointIndex and/or secondPointIndex is not a start or end point, but both must be start and/or end point."
+    );
+  }
+  if (firstPointIndex > secondPointIndex) {
+    [firstPointIndex, secondPointIndex] = [secondPointIndex, firstPointIndex];
+  }
 
   const [firstContourIndex, firstContourPointIndex] =
     path.getContourAndPointIndex(firstPointIndex);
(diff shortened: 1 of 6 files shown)


ollimeier:
Here is a short demo:

https://github.com/user-attachments/assets/2c6c340d-0856-4887-be51-9014f1f219ba

justvanrossum (commented):
You are repurposing the connectContours() function for something it was not designed to do.

connectContours(): connects contours if we already have an end point that overlaps with a start point. That is a very specific task!

What you already have from the Close Contours PR: close an open path.

What you need: connect two _distinct_ contours (joining the end points of a single contour is already handled by your existing code), taking into account the possibility that there's a dangling off curve point and turn it into a proper two-handle cubic.

src/fontra/views/editor/scene-controller.js
```diff
@@ -1462,3 +1533,67 @@ function splitLocation(location, glyphAxes) {
...
+  }
+
+  if (!firstContourPointIndex) {
+    [firstContour, secondContour] = [secondContour, firstContour];
+  }
+  let selectedContourPointIndex1 = firstContour.points.length - 1;
+  let selectedContourPointIndex2 = selectedContourPointIndex1 + 1;
+  let loneCubicHandle;
+  const lastPointFirstContour = firstContour.points.at(-1);
+  if (lastPointFirstContour.type) {
```
justvanrossum, line 1559: The way this is written allows `loneCubicHandle` to be written to twice. Which I think why the following case doesn't work correctly:

- contour 1 ends with a dangling off-curve point
- contour 2 starts with a dangling off-curve point

This case should _not_ result more off-curves being added

This if and the next if should be an if / else if instead, not just to prevent the above bug, but also to make the intention clearer.

src/fontra/views/editor/scene-controller.js
```diff
@@ -1462,3 +1533,67 @@ function splitLocation(location, glyphAxes) {
...
+    loneCubicHandle = firstContour.points.pop();
+  }
+
+  const firstPointSecondContour = secondContour.points.at(0);
+  if (firstPointSecondContour.type && !lastPointFirstContour.type) {
+    loneCubicHandle = secondContour.points.shift();
+  }
+
+  if (loneCubicHandle) {
+    const [handle1, handle2] = [lastPointFirstContour, firstPointSecondContour].map(
```
justvanrossum, line 1569: This can't be right, because one of these you have just popped/shiffed from the contour. And indeed: I don't see the correct curve appear: one of the handles stays in the same place where I expect it to move.

src/fontra/views/editor/scene-controller.js
```diff
@@ -647,18 +647,37 @@ export class SceneController {
...
       {
-        title: () =>
-          translatePlural(
-            "action.close-contour",
-            this.contextMenuState.openContourSelection?.length
-          ),
-        enabled: () => this.contextMenuState.openContourSelection?.length,
-        callback: () => this.doCloseSelectedOpenContours(),
+        title: () => {
+          if (this.contextMenuState.joinContourSelection?.length === 2) {
```
justvanrossum, line 662: This also triggers when the two selected points are part of the same contour. Which leads to the wrong title.

Also, this should be a `condition ? A : B` expression.

src/fontra/views/editor/scene-controller.js
ollimeier: Thanks for finding this. I hope I understood you right. Here is a short demo video as a reference for the fix:

https://github.com/user-attachments/assets/ab05b8e2-94b0-4541-8f07-cb11a7925b84

I mark this as resolved.

src/fontra/views/editor/scene-controller.js
ollimeier: Thanks once again. Should be fixed now.

src/fontra/views/editor/scene-controller.js
ollimeier: Makes absolutely sense, with the 'same contour' issue. The was even duplicated/redundant code related to that, I was able to remove. Sorry. The condition is done, too.

src/fontra/views/editor/scene-controller.js
```diff
@@ -1168,6 +1181,28 @@ export class SceneController {
...
+          layerGlyph.path,
+          pointIndex1,
+          pointIndex2
+        );
+
+        for (const pointIndex of selectionPointIndices) {
+          newSelection.add(`point/${pointIndex}`);
+        }
+      }
+      if (newSelection) {
```
justvanrossum, line 1199: Since you initialize `newSelection` with an empty set, this will always be true. Do you intend to do this only on a non-empty selection, or is it a safe-guard against undefined/null?

src/fontra/views/editor/scene-controller.js
```diff
@@ -1462,3 +1524,66 @@ function splitLocation(location, glyphAxes) {
 
   return { fontLocation, glyphLocation };
 }
+
+export function connectTwoDistinctContours(path, firstPointIndex, secondPointIndex) {
```
justvanrossum, line 1528: Suggest to call this function `joinContours`.

src/fontra/views/editor/scene-controller.js
```diff
@@ -1462,3 +1524,66 @@ function splitLocation(location, glyphAxes) {
...
+  let selectedPointIndices = [];
+
+  const [firstContourIndex, firstContourPointIndex] =
+    path.getContourAndPointIndex(firstPointIndex);
+  const [secondContourIndex, secondContourPointIndex] =
+    path.getContourAndPointIndex(secondPointIndex);
+
```
justvanrossum, line 1535: Maybe we need some sanity checking here:
- assert that the contour indices are indeed different
- assert that .isStartOrEndPoint() is true for both indices
- assert that firstPointIndex < secondPointIndex

Throw an exception if any of the above does not hold.

src/fontra/views/editor/scene-controller.js
ollimeier: I removed the unnecessary check and also made it `const`. done.

src/fontra/views/editor/scene-controller.js
ollimeier: done.

src/fontra/views/editor/scene-controller.js
ollimeier: The first two items are covered in `getSelectedJoinContoursPointIndices`, therefore we know what we pass into `joinContours` is valid. Let me add the last item to with a sorting to getSelectedJoinContoursPointIndices and we are done. Or do you want to be on the save side if we may want to use joinContours somewhere else in future?

ollimeier: OK, done. Except the last point I made the point indices swap. I hope you find it also a valid solution – if not, I'll change it.

ollimeier: OK, sorry. Changed it as requested.

src/fontra/views/editor/scene-controller.js
```diff
@@ -1462,3 +1523,78 @@ function splitLocation(location, glyphAxes) {
 
   return { fontLocation, glyphLocation };
 }
+
+export function joinContours(path, firstPointIndex, secondPointIndex) {
+  let selectedPointIndices = [];
+  assert(
+    firstPointIndex != secondPointIndex,
+    "firstPointIndex and secondPointIndex are the same, but must be different."
```
justvanrossum, line 1531: Actually, [what I asked](https://github.com/googlefonts/fontra/pull/1570/files#r1689941939) was about the *contour* index. That would also catch the case where the point indices are the same.

justvanrossum: Also, please formulate more compactly, like this:
```
"firstPointIndex and secondPointIndex must be different."
```

src/fontra/views/editor/scene-controller.js
```diff
@@ -1462,3 +1523,78 @@ function splitLocation(location, glyphAxes) {
...
+  );
+  assert(
+    path.isStartOrEndPoint(firstPointIndex) && path.isStartOrEndPoint(secondPointIndex),
+    "firstPointIndex and/or secondPointIndex is not a start or end point, but both must be start and/or end point."
```
justvanrossum, line 1535:
```suggestion
    "firstPointIndex and secondPointIndex must be start or end points"
```

src/fontra/views/editor/scene-controller.js
```diff
@@ -1462,3 +1523,78 @@ function splitLocation(location, glyphAxes) {
...
+  );
+  assert(
+    firstPointIndex < secondPointIndex,
+    "firstPointIndex must be smaller than secondPointIndex, but it is not."
```
justvanrossum, line 1539:
```suggestion
    "firstPointIndex must be less than secondPointIndex"
```

src/fontra/views/editor/scene-controller.js
ollimeier: I am sorry for that. I changed the message + added a sanity check for contour indices. I am closing this as resolved.

justvanrossum: The point index assert is now redundant. As I wrote, contour index assert also catches that case. Please remove it.

ollimeier: done.

justvanrossum (approved):

PR #1571, opened 2024-07-24
Run tests on 3.12

justvanrossum:
~(But let's also try on 3.13, will undo if it causes trouble)~

We can't use 3.13 yet, it is blocked by https://github.com/samuelcolvin/watchfiles/issues/276


EwoutH:
An important PR for Python 3.13 compatibility in watchfiles is now merged, a new release should be created today or tomorrow.

PR #1578, opened 2024-07-26
Hide transform selection tool if pointer-tool is not selected

### LINKED ISSUE #1567: Interactive resize tool: hide bounding box in other tools
Currently the bounding box with handles are shown in all other tools, although it is not possible to use it other than in the pointer tool. We should hide the resize bounding box in all other tools except the pointer tool (and maybe the hand tool).

ollimeier:
Fixes #1567

### DIFF
--- a/src/fontra/views/editor/edit-tools-pointer.js
+++ b/src/fontra/views/editor/edit-tools-pointer.js
@@ -648,6 +648,17 @@ export class PointerTool extends BaseTool {
     }
     return undefined;
   }
+
+  activate() {
+    this.setCursor();
+    this.sceneController.sceneModel.showTransformSelection = true;
+    this.canvasController.requestUpdate();
+  }
+
+  deactivate() {
+    this.sceneController.sceneModel.showTransformSelection = false;
+    this.canvasController.requestUpdate();
+  }
 }
 
 function pointInSquareHandle(point, handle, handleSize) {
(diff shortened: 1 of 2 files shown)


justvanrossum (commented):
This is the wrong place. Tools have activate() and deactivate() methods that you can override. I'm pretty sure you can do this by overriding these on the pointer tool class, and set showTransfromSelection there.

https://github.com/googlefonts/fontra/blob/fff62d67f95bc7ce3507d45b4a791dc68f0f0612/src/fontra/views/editor/edit-tools-base.js#L15-L21

ollimeier:
> This is the wrong place. Tools have activate() and deactivate() methods that you can override. I'm pretty sure you can do this by overriding these on the pointer tool class, and set showTransfromSelection there.
> 
> https://github.com/googlefonts/fontra/blob/fff62d67f95bc7ce3507d45b4a791dc68f0f0612/src/fontra/views/editor/edit-tools-base.js#L15-L21

Thanks, I changed it. But it's weird, that when I have something selected and go back and forth the transform selection bounding box does not show up again when the pointer tool is selected. It does show up, when hovering over the glyph, but not immediately when changing the tool.

justvanrossum:
Very basic things:
1. If you override a method, you must call super.
2. You could have debugged this: activate() was simply never called.

PR #1582, opened 2024-07-31
Add remove overlaps

### LINKED ISSUE #1514: Add "remove overlap" function
- It should take the selection into account:
  - Only work on selected contours
  - Could we make it work on a _partially_ selected contour? This may be hard to do.
- We do not have JS implementation (long term I hope Kurbo + JS bindings will be the solution) so we are for now stuck with doing it in Python:
  - Send the selected outlines to the server, perform remove overlap, send result back (as we do with clipboard conversion and glyph name / code point lookups)
- Use skia-pathops rather than booleanOperations. The former supports quadratics. But perhaps we should be able to use both, as the tradeoffs between skia-pathops and booleanOperations aren't super clear-cut.

ollimeier:
Fixes #1514.


ollimeier:
Here is a short demo:

https://github.com/user-attachments/assets/5011ace2-6102-45c8-b9be-1d3282874b8f

src/fontra/client/core/server-utils.js
```diff
@@ -24,3 +24,7 @@ export const getCodePointFromGlyphName = memoize(async (glyphName) => {
 export async function parseClipboard(data) {
   return await callServerAPI("parseClipboard", { data });
 }
+
+export async function unionPath(data, contourIndices) {
```
justvanrossum, line 28: I think any path subsetting (which contours to even consider) should be done at the front end, and you should not pass contourIndices. Also, you should pass a _path_, not a _glyph_, just like I explicitly suggested in https://github.com/googlefonts/fontra/issues/1514#issuecomment-2258450907.

src/fontra/core/serverutils.py
```diff
@@ -22,3 +27,68 @@ def getCodePointFromGlyphName(glyphName):
...
     return unstructure(clipboard.parseClipboard(data))
+
+
+# I have the feeling that I am overcomplicating things here, but even if not,
+# the following code should likely be moved to a different location.
+def skiaPathFromStaticGlyph(
+    staticGlyph: StaticGlyph, contourIndices: list
+) -> pathops.Path:
+    builder = pathops.OpBuilder()
+    for i, contour in enumerate(staticGlyph.path.asPath().contours):
```
justvanrossum, line 38: You should use the pen protocols and helper classes instead of rolling your own pen stuff.

The PackedPath class has a drawPoints(pen) method, that you can use. The skia-pathops path.getPen() returns a segment pen, whereas PackedPath.drawPoints() expects a point pen. Convert like this.

Use the fontTools.pens.pointPen.PointToSegmentPen and fontTools.pens.pointPen.SegmentToPointPen adapter classes to adapt one protocol to the other. Into a pathops path:

```python
path = pathops.Path()
fontraPath.drawPoints(PointToSegmentPen(path.getPen())
```

To _construct_ a PackedPath you should use the PackedPathPointPen.

justvanrossum (commented):
Thanks!

I'm not sure about the icon choice: let's fine tune later.

I left some comments.

requirements.txt
```diff
@@ -6,3 +6,4 @@ pyyaml==6.0.1
 ufomerge==1.7.1
 unicodedata2==15.1.0
 watchfiles==0.22.0
+skia-pathops==0.8.0
```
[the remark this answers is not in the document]
justvanrossum: Btw. this should be `skia-pathops==0.8.0.post1`. This may solve the test failure. Always start with the lastest.

src/fontra/client/core/server-utils.js
ollimeier: Thanks. done.

requirements.txt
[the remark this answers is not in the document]
ollimeier: done

src/fontra/core/serverutils.py
```diff
@@ -22,3 +30,52 @@ def getCodePointFromGlyphName(glyphName):
...
+
+
+@api
+def intersectPath(pathA, pathB):
+    return skiaPathOperations(pathA, pathB, pathops.PathOp.INTERSECTION)
+
+
+@api
+def excludePath(pathA, pathB):
+    return skiaPathOperations(pathA, pathB, pathops.PathOp.DIFFERENCE)
```
justvanrossum, line 61: This should be `pathops.PathOp.XOR`.

ollimeier:
So, I added the other path operations as well. For me it's counter-intuitive, that the path operations interactive with none selected contours.

https://github.com/user-attachments/assets/eacc8188-cb0e-42db-b65a-4e1178bd81ec


For reference: in Figma subtract works the way that the bottom layer of a selection remains:
https://www.youtube.com/watch?v=8ZaX6qEcUCU

justvanrossum (commented):
Please create a new module `pathops.py`, with the four functions plus support code. 

serverutils.py will expose these as apis, as thin wrappers around the actual functionality. This follows how the other serverutils are implemented.

justvanrossum:
Please use `layers-intersect-2.svg` for "intersect" / `INTERSECTION` , and use "layers-difference.svg" for "exclude" / `XOR`.

justvanrossum:
> For me it's counter-intuitive, that the path operations interactive with none selected contours.

The Figma example is simple, it only uses two shapes. And it may have a different notion of "layers". With our approach, we can do operations on collections of shapes, including counters:

https://github.com/user-attachments/assets/1825ef60-32e3-4f1c-9891-9b6bd1d7aca9

This is not possible with "operate on the selection" alone.

justvanrossum:
I think we may need a fifth option:
- Remove overlaps (remove overlaps on selected contours)
- **Union selection with unselected path**
- Subtract selection from unselected path
- Intersect selection with unselected path
- Symmetric difference (XOR) selection with unselected path

justvanrossum:
> Union selection with unselected path

Playing with this: it is not different from "select all -> remove overlap", so it is redundant after all.

justvanrossum:
The problem with the other operators is that we need two explicit groups of contours. We don't have that option in Fontra, so the next bet thing (IMO) is to treat the selection as one group, and the rest as the other group. To me that feels very natural, although I'll admit that it also has its limitations. Still: it is directly much more useful and explainable then doing things based on the contour order.

justvanrossum:
Use case:
https://github.com/user-attachments/assets/a867503d-1cf7-427c-8e99-ffdade33c5c3

justvanrossum:
A big downside is that it implicitly performs remove overlap on the unselected contours, which likely is undesirable.

src/fontra/core/serverutils.py
ollimeier: Right. Thanks. done.

justvanrossum:
> downside

https://github.com/user-attachments/assets/5f38f955-dec0-416e-879c-bc754fc06c77

On the other hand, I don't really see a way around that: pathops just seems to imply remove overlap on both sides of the operator.

ollimeier:
> Please create a new module `pathops.py`, with the four functions plus support code.
> 
> serverutils.py will expose these as apis, as thin wrappers around the actual functionality. This follows how the other serverutils are implemented.

Thanks. Done.

src/fontra/views/editor/panel-transformation.js
```diff
@@ -451,6 +530,77 @@ export default class TransformationPanel extends Panel {
...
+  async doPathOperations(pathOperationFunc, undoLabel) {
+    let { point: pointIndices } = parseSelection(this.sceneController.selection);
+    pointIndices = pointIndices || [];
+
+    const positionedGlyph =
+      this.sceneController.sceneModel.getSelectedPositionedGlyph();
+
+    const allContourIndices = [...range(positionedGlyph.glyph.path.numContours)];
+    const selectedContourIndices =
+      !pointIndices.length && undoLabel === "Remove overlaps"
```
justvanrossum, line 542: I don't like to switch things based on a label. You could compare `pathOperationFunc` with `unionPath` instead. But maybe do that once at the top of this function, and use a variable for that flag.

justvanrossum: Say `doUnion`

src/fontra/views/editor/panel-transformation.js
```diff
@@ -451,6 +530,77 @@ export default class TransformationPanel extends Panel {
...
+        ? allContourIndices
+        : [...getSelectionByContour(positionedGlyph.glyph.path, pointIndices).keys()];
```
justvanrossum, line 544: I want you to rewrite this, closer to how I originally suggested.

Idea:
```js
  const selectedContourIndicesMap = getSelectionByContour(positionedGlyph.glyph.path, pointIndices)
  const isContourSelected = pointIndices.length || !doUnion ? (i) => selectedContourIndicesMap.has(i) : (i) => true;
```

Then use`isContourSelected()` directly in the loop where you split the path in two.

src/fontra/views/editor/panel-transformation.js
```diff
@@ -451,6 +530,77 @@ export default class TransformationPanel extends Panel {
...
+
+        for (const contourIndex of range(path.numContours)) {
+          if (selectedContourIndices.includes(contourIndex)) {
+            selectedContoursPath.appendContour(path.getContour(contourIndex));
+          } else {
+            unselectedContoursPath.appendContour(path.getContour(contourIndex));
+          }
+        }
+
+        return await pathOperationFunc(selectedContoursPath, unselectedContoursPath);
```
justvanrossum, line 576: `unionPath()` only takes one argument, so this looks misleading.

src/fontra/views/editor/panel-transformation.js
```diff
@@ -451,6 +530,77 @@ export default class TransformationPanel extends Panel {
...
+      }
+    );
+
+    await this.sceneController.editGlyphAndRecordChanges(
+      (glyph) => {
+        for (const [layerName, newPath] of Object.entries(layerPaths)) {
+          const path = glyph.layers[layerName].glyph.path;
```
justvanrossum, line 583: I would restructure:
```js
  const glyph = glyph.layers[layerName]
  if (doUnion) {
    ...delete selected contours...
    append new path to glyph.path
  } else {
    glyph.path = ...thenewpath...  // just replace the whole thing
  }
```

src/fontra/views/editor/panel-transformation.js
justvanrossum: I would also love you to reverse the arguments: subtractPaths is now backwards (hence you use REVERSE_DIFFERENCE in the implementation). subtractPaths(pathA, pathB) should subtract pathB from pathA.

justvanrossum (commented):
I bet one of the icons is now no longer being used.

ollimeier:
> I bet one of the icons is now no longer being used.

Right. I removed it. Thanks.

justvanrossum:
We'll have to experiment with this method of doing pathops, instead of using OpBuilder:

https://github.com/fonttools/skia-pathops/blob/815070e964334ec6cdf453f2d859fe2db49d9b06/src/python/pathops/operations.py#L40-L59

(As kindly suggested in https://github.com/fonttools/skia-pathops/issues/74)

src/fontra/views/editor/panel-transformation.js
ollimeier: I used `glyph.layers[layerName].glyph.path = layerPath;` but it fails and I don't know why. I compared the code you did for the knife tool, but I cannot see what causes it to fail.

The contours outlines are gone, but displayed in light grey and an error pops up in the console:

Screenshot 2024-08-01 at 15 18 08

ollimeier: Please see: https://github.com/googlefonts/fontra/pull/1582/commits/0ae336495c127ddbf38df68d6d3494a001b19406

src/fontra/views/editor/panel-transformation.js
```diff
@@ -531,63 +531,62 @@ export default class TransformationPanel extends Panel {
...
-        const path = layerGlyph.path;
+        const path = layerGlyph.path.copy();
         const selectedContoursPath = new VarPackedPath();
         const unselectedContoursPath = new VarPackedPath();
 
         for (const contourIndex of range(path.numContours)) {
-          if (selectedContourIndices.includes(contourIndex)) {
+          if (isContourSelected(contourIndex)) {
             selectedContoursPath.appendContour(path.getContour(contourIndex));
+            contoursToBeDeleted.push(contourIndex);
```
justvanrossum, line 565: You are doing this for each layer, so this likely breaks multi source editing. Also, it is literally the list of keys of `selectedContourIndicesMap`, so there is really no need to do this here.

src/fontra/views/editor/panel-transformation.js
ollimeier: done.

src/fontra/core/pathops.py
```diff
@@ -0,0 +1,54 @@
...
+    PointToSegmentPen,
+    SegmentToPointPen,
+)
+
+from .classes import structure, unstructure
+from .path import PackedPath, PackedPathPointPen
+
+
+def skiaPathOperations(pathA, pathB, pathOperation):
+    fontraPathA = structure(pathA, PackedPath)
```
justvanrossum, line 13: The structure/unstructure parts belong in serverutils.py: it's a consequence of how the web api funcs work, and should not affect this library module.

src/fontra/core/pathops.py
```diff
@@ -0,0 +1,54 @@
...
+    skiaPathA = pathops.Path()
```
justvanrossum, line 14: Please write helper functions:
- `skiaPathToFontraPath()`
- `fontraPathToSkiaPath()`

src/fontra/views/editor/panel-transformation.js
ollimeier: done, please see: https://github.com/googlefonts/fontra/pull/1582/commits/dc01cb28bae3b80e17cbf6243667aacca2e5f57f

src/fontra/views/editor/panel-transformation.js
ollimeier: Fixed it: 
[Bugfix: layerPath was not a VarPackedPath](https://github.com/googlefonts/fontra/pull/1582/commits/c363c2084cb87a05864abe94ce2e3bbd1e0a5d64)

src/fontra/core/pathops.py
ollimeier: done, see: https://github.com/googlefonts/fontra/pull/1582/commits/2c37dc2d96eee1a6f0fc6fcb71cbcc2497a51d3b

src/fontra/views/editor/panel-transformation.js
```diff
@@ -580,16 +580,17 @@ export default class TransformationPanel extends Panel {
     await this.sceneController.editGlyphAndRecordChanges(
       (glyph) => {
         for (const [layerName, layerPath] of Object.entries(layerPaths)) {
+          const newPackedPath = VarPackedPath.fromObject(layerPath);
```
justvanrossum, line 583: Suggest to do the conversion in server-utils.js.

justvanrossum: (In retrospect, parseClipboard() should do the same, there's now a conversion method in editor.js. For later maybe.)

src/fontra/views/editor/panel-transformation.js
ollimeier: ok, done.

justvanrossum:
> We'll have to experiment with this method of doing pathops, instead of using OpBuilder:

I did: doesn't make a difference.

ollimeier:
> > We'll have to experiment with this method of doing pathops, instead of using OpBuilder:
> 
> I did: doesn't make a difference.

Thanks for helping me out with this.

src/fontra/client/lang/zh-CN.json
```diff
@@ -158,6 +158,11 @@
   "sidebar.selection-transformation.origin.top.center": "原点为上侧中心",
   "sidebar.selection-transformation.origin.top.left": "原点为左上角",
   "sidebar.selection-transformation.origin.top.right": "原点为右上角",
+  "sidebar.selection-transformation.pathOperations": "Path Operations",
```
justvanrossum, line 161: The style for keys is hyphenated, please use `sidebar.selection-transformation.path-operations`. Same below.

justvanrossum: (And in English as well of course)

src/fontra/views/editor/panel-transformation.js
```diff
@@ -434,6 +448,78 @@ export default class TransformationPanel extends Panel {
...
+      "sidebar.selection-transformation.pathOperations.exclude"
+    );
+    formContents.push({
+      type: "universal-row",
+      field1: {
+        type: "auxiliaryElement",
+        key: "removeOverlaps",
+        auxiliaryElement: html.createDomElement("icon-button", {
+          "src": "/tabler-icons/layers-union.svg",
+          "onclick": (event) => this.doPathOperations(unionPath, labelUnion),
```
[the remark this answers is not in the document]
ollimeier: done.

src/fontra/client/lang/zh-CN.json
ollimeier: ok. done.

PR #1623, opened 2024-08-28
Show dialog message if URL is malformed.

### LINKED ISSUE #1616: Malformed URL fragment does not give any feedback except an empty canvas
For example: a truncated URL causes an "unexpected EOF" error during decompression.

Suggested fix:
- catch decoding/deserialize error, show dialog saying something like "The URL is malformed: the UI settings could not be restored"

ollimeier:
Fixes #1616

### DIFF
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -518,7 +518,7 @@ export function loadURLFragment(fragment) {
   try {
     return JSON.parse(strFromU8(unzlibSync(base64ToBytes(fragment.slice(1)))));
   } catch {
-    return false;
+    return null;
   }
 }
 
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -515,7 +515,11 @@ export function loadURLFragment(fragment) {
   if (fragment[0] != "#") {
     throw new Error("assert -- invalid fragment");
   }
-  return JSON.parse(strFromU8(unzlibSync(base64ToBytes(fragment.slice(1)))));
+  try {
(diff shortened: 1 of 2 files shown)


src/fontra/client/core/utils.js
```diff
@@ -515,7 +515,11 @@ export function loadURLFragment(fragment) {
   if (fragment[0] != "#") {
     throw new Error("assert -- invalid fragment");
   }
-  return JSON.parse(strFromU8(unzlibSync(base64ToBytes(fragment.slice(1)))));
+  try {
+    return JSON.parse(strFromU8(unzlibSync(base64ToBytes(fragment.slice(1)))));
+  } catch {
+    return false;
```
justvanrossum, line 521:
```suggestion
    return null;
```

justvanrossum (commented):
One small comment. Otherwise looks fine.

justvanrossum (approved):

justvanrossum (changes requested):
I just now tried it myself, but while there is a dialog now, there is a different error after this, prevnting the URL to be updated later:

image

Paste a malfored URL in a new tab, dismiss the dialog, then add text to the text entry field. Notice how the URL does not get updated: it stays at the malformed URL.

ollimeier:
> I just now tried it myself, but while there is a dialog now, there is a different error after this, prevnting the URL to be updated later:
> 
> image
> Paste a malfored URL in a new tab, dismiss the dialog, then add text to the text entry field. Notice how the URL does not get updated: it stays at the malformed URL.

Thanks for finding this. I fixed it.

justvanrossum (approved):

Issue #1461, opened 2024-06-16
Hope Fontra has an official Chinese name


buwuzhengyedexiaoji:
This software has a new nickname, “字挼 (zìruá)” among the Chinese font designer community. Within this context, the term “字” serves as a free translation for “font”, while “挼” is a transliteration of “-ra”. The whole translation means “to touch the glyphs”. I believe it’s worth considering this name as the official Chinese translation. Hope you’ll take my suggestion into account.

LittleWhiteDoge:
+1
That will be a great translation 😘

justvanrossum:
After internal discussion weve decided not to go with this name.

PR #1625, opened 2024-08-29
Add shortcuts panel

ollimeier:
This is a draft PR for early feedback.

Based on our meetings (yesterday and today), I reworked a couple items:
- undo/replace ui-form code changes with custom made code
- set default shortcuts for tools back to 1234567
- remove editing of globalOverride in UI and dialog
- a different solution for the "delay recoding" of a shortcut
- add export/import of shortcuts (via json files)
- merge the two 'undo' shortcuts into one, even though there are internally two (one with and one without globalOverride)
- open dialog via double click on input element (not on label as it was before)
- added delete and refresh button next to the input element (only visible if shortcut input is focused)

Open issues (I still have not found a solution, yet):
- the refresh and delete svg icons are not shown (currently displayed with green boxes)
- the shortcuts are not displayed correct in the menu (this seems to be a loading issue)
- tooltip does not work for label nor for input element, but it works for the icon-buttons, which is weird.


src/fontra/client/lang/zh-CN.json
```diff
@@ -86,6 +91,8 @@
   "menubar.file.open": "打开……",
   "menubar.font": "字体",
   "menubar.font.edit": "编辑字体信息、参数轴和源",
+  "menubar.fontra": "字挼 (zìruá)",
```
justvanrossum, line 94: Where did this translation come from?

src/fontra/client/web-components/inline-svg.js
```diff
@@ -3,11 +3,16 @@
 import { htmlToElement } from "/core/html-utils.js";
 
 export class InlineSVG extends HTMLElement {
-  constructor(src) {
+  constructor(src, options) {
```
justvanrossum, line 6: Instead of hacking workarounds into existing code, let's figure out the problem.

src/fontra/client/lang/zh-CN.json
ollimeier: Here: https://github.com/googlefonts/fontra/issues/1461

src/fontra/client/web-components/inline-svg.js
ollimeier: I reverted the latest changes.

src/fontra/client/lang/zh-CN.json
justvanrossum: Ah, I never closed this issue, but we decided back then (with Mark and other Chinese speaker at Black Foundry) that we'll not go with that name. It was even suggested the issue may be trolling.

ollimeier: OK, thanks. I was not aware of that. I removed the translation.

ollimeier:
I am closing this, because I will open a new draft PR in a bit (based on the new 'actions'-approach).

PR #1636, opened 2024-09-04
Add customizable shortcuts panel

### LINKED ISSUE #721: Custom shortcuts for tools
It would be great to be able to customize shortcuts for the main tools (Selection, Pen, Measurement, etc.), for example assigning them to letter keys instead of number keys.

Currently there are a few issues with having the number keys as shortcuts:

1. If we first adjust an axis by typing a value or using arrow keys, and want to immediately switch to a different tool, pressing a number key directly would actually enter the number into the axis value box, and thus changing the axis value and the component. Currently we often find ourselves messing up axis values this way and having to undo the action, or need to be extra careful to make sure to click outside the axis value box first before switching tools. (It seems pressing a letter key would not enter the letter into the box.)

2. On standard French AZERTY keyboards, the number keys are shift-functions, so the number shortcuts require 2 keys.

Then some designers have expressed they are more familiar with letter keys as shortcuts for tools, similar to Glyphs or some other softwares (V for selection, P for pen, etc.)

### LINKED ISSUE #1589: Specify keyboard shortcuts for tools
Currently the 'main' tools keyboard shortcuts go from 1 to 6. But whenever a new tool gets added the shortcuts may change depending on its position, therefore the users requested unique keyboard shortcuts which should be ideally be customizable.

- Pointer Tool: (need to be specified)
- Pen Tool: (need to be specified)
- Knife Tool: (need to be specified)
- Shape Tool: (need to be specified)
- Power Ruler Tool: (need to be specified)
- Handtool Tool: (need to be specified)

### LINKED ISSUE #1594: New 'application settings' view
Today we discussed that we need a new section for general settings for any kind of setting which do not belong to the font. For example: shortcuts or appearance settings like text sizes.

We may need a new menu bar item 'Fontra' at the beginning or at the end of the main menu, similar to other application menubars on macOS.

At the beginning this will not be displayed, because this is in preparation for customizable shortcuts #721

Fontra
-> Application settings
--> Shortcuts
--> Editor appearance
--> Extensions
--> Server info

All the items should be more or less placeholders, because at the current stage it's meant as preparation.
The content is more similar to the 'font info sections'.

Notes related to 'Shortcuts':
A mock-up which contains a list of items with shortcuts (customizable). The list should be grouped, eg. one group could be 'Tools'. At that stage it's not necessary to make it foldable. 
A custom design search box is not required, because the build-in browser search is good enough. At least for now.

This view will partially replace the "gear" sidebar, and therefore this issue relates to (or is part of) #1588.

### LINKED ISSUE #1642: Customizable shortcuts for visualization layers
In todays meeting we discussed to implement shortcuts for visualization layers. 
This might be a good follow up after merging the new shortcuts panel.

ollimeier:
This fixes #1594, fixes #1589, fixes #721, and fixes #1642.

Known issues:
- `event.preventDefault()` does not work for default shortcuts like cmd+N (event.stopImmediatePropagation() does not work either) 
-> **some default shortcuts cannot be overwritten**, please see: [bugs.chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=33056&no_tracker_redirect=1)
-> [Good comment related to this issue]( https://github.com/liftoff/GateOne/issues/290#issuecomment-23121578)


src/fontra/client/lang/en.json
```diff
@@ -60,6 +73,7 @@
   "editor.pen-tool": "Pen Tool",
   "editor.pointer-tool": "Pointer Tool",
   "editor.power-ruler-tool": "Power Ruler Tool",
+  "editor.shape-tool": "Shape Tool",
   "editor.shape-tool-ellipse": "Shape Tool Ellipsis",
```
justvanrossum, line 77: This typo must have existed for a while...
```suggestion
  "editor.shape-tool-ellipse": "Shape Tool Ellipse",
```

src/fontra/views/applicationsettings/panel-base.js
```diff
@@ -0,0 +1,91 @@
...
+    if (onOff && !this.initialized) {
+      this.initializePanel();
+      this.initialized = true;
+    }
+  }
+
+  initializePanel() {
+    this.undoStack = new UndoStack();
+
+    // const subscribePattern = Object.fromEntries(
```
justvanrossum, line 23: These things should be deleted: it is when a view deals with a font. This view doesn't.

justvanrossum (commented):
This isn't a proper review yet, I just noticed some small things.

src/fontra/views/applicationsettings/panel-shortcuts.js
```diff
@@ -0,0 +1,659 @@
...
+    const shortCutsDataCustom = {};
+    for (const actionIdentifier of getActionIdentifiers()) {
+      const actionInfo = getActionInfo(actionIdentifier);
+      const shortCuts = actionInfo.customShortCuts;
+      if (!shortCuts) {
+        continue;
+      }
+      shortCutsDataCustom[actionIdentifier] = shortCuts;
+    }
+    const data = JSON.stringify(shortCutsDataCustom);
```
justvanrossum, line 140: Use indentation of 2 spaces:
```suggestion
    const data = JSON.stringify(shortCutsDataCustom, null, 2);
```

src/fontra/views/applicationsettings/panel-shortcuts.js
```diff
@@ -0,0 +1,659 @@
...
+    const blob = new Blob([data], { type: "application/json" });
+    const url = URL.createObjectURL(blob);
+    const a = document.createElement("a");
+    a.href = url;
+    a.download = "shortcuts.json";
```
justvanrossum, line 145:
```suggestion
    a.download = "fontra-shortcuts.json";
```

justvanrossum (commented):
Two more small things.

justvanrossum:
I noticed that if you assign a shortcut to the ellipse tool, the tool will switch, but the icon doesn't. This is technically unrelated to this PR, but perhaps it stll makes sense to fix it here.

ollimeier:
> I noticed that if you assign a shortcut to the ellipse tool, the tool will switch, but the icon doesn't. This is technically unrelated to this PR, but perhaps it stll makes sense to fix it here.

I agree. I fixed it here already: https://github.com/googlefonts/fontra/pull/1604/files
It's related to `collapseSubTools`.

justvanrossum:
> I fixed it here already: https://github.com/googlefonts/fontra/pull/1604/files

Ah, but that PR got merged into a branch which is now abandoned, right?

ollimeier:
> > I fixed it here already: https://github.com/googlefonts/fontra/pull/1604/files
> 
> Ah, but that PR got merged into a branch which is now abandoned, right?

yes.

src/fontra/client/core/actions.js
```diff
@@ -187,6 +223,9 @@ function loadActionsByKeyOrCode() {
   actionsByKeyOrCode = {};
   for (const [actionIdentifier, action] of Object.entries(actionInfoController.model)) {
     for (const shortCut of action.customShortCuts || action.defaultShortCuts || []) {
+      if (!shortCut) {
```
justvanrossum, line 226: How can this happen?

src/fontra/client/core/utils.js
```diff
@@ -120,14 +120,18 @@ export function hyphenatedToLabel(s) {
   return capitalizeFirstLetter(s).replaceAll("-", " ");
 }
 
+// platform is deprecated, please see:
+// https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform
+// export const isMac = typeof navigator !== "undefined" && navigator.platform.toLowerCase().indexOf("mac") >= 0
+
+// Therefore use window.navigator https://developer.mozilla.org/en-US/docs/Web/API/Window/navigator
+export const isMac =
+  typeof navigator !== "undefined" && navigator.userAgent.indexOf("Mac") != -1;
```
justvanrossum, line 129: I don't think we support browsers that don't support `navigator`, so I think that first check is redundant.

src/fontra/views/applicationsettings/panel-shortcuts.js
```diff
@@ -0,0 +1,682 @@
...
+        "onclick": (event) => this.resetShortCut(id),
+        "data-tooltip": "Reset to default",
+        "data-tooltipposition": "top",
+      })
+    );
+
+    this.append(
+      html.createDomElement("icon-button", {
+        "class": "fontra-ui-shortcuts-panel-icon",
+        "src": "/tabler-icons/trash.svg",
```
justvanrossum, line 673: I think we should use an "X" icon instead: https://tabler.io/icons/icon/x

src/fontra/client/core/actions.js
```diff
@@ -27,19 +28,22 @@ actionInfoController.addListener((event) => {
   actionsByKeyOrCode = undefined;
 });
 
+let sortIndex = 0;
 export function registerAction(actionIdentifier, actionInfo, callback, enabled = null) {
-  registerActionInfo(actionIdentifier, actionInfo);
+  registerActionInfo(actionIdentifier, actionInfo, sortIndex);
   registerActionCallbacks(actionIdentifier, callback, enabled);
+  sortIndex++;
```
justvanrossum, line 35: I think the sortIndex logic should be done by registerActionInfo() itself. And it would not need the sortIndex argument.

src/fontra/client/core/actions.js
```diff
@@ -27,19 +28,22 @@ actionInfoController.addListener((event) => {
...
-export function registerActionInfo(actionIdentifier, actionInfo) {
+export function registerActionInfo(actionIdentifier, actionInfo, sortIndex) {
   actionInfoController.synchronizeItemWithLocalStorage(actionIdentifier, actionInfo);
   // We only want customShortCuts to be changable, so we'll reset everything
   // except customShortCuts
   const storedActionInfo = getActionInfo(actionIdentifier);
   actionInfoController.model[actionIdentifier] = {
     ...actionInfo,
     customShortCuts: storedActionInfo.customShortCuts,
+    sortIndex: sortIndex,
```
justvanrossum, line 46: I think the caller should (optionally) be able to set the sortIndex itself. I would set `actionInfo.sortIndex` only when `actionInfo.sortIndex` is undefined.

src/fontra/client/core/actions.js
justvanrossum: Perhaps each topic should have its own sortIndex counter.

I assume the topic in the UI are currently also sorted based on the sortIndex?

justvanrossum (commented):
Some feedback, it's not a complete review.

This doesn't support multiple shortcuts for a single action yet, right?

justvanrossum:
I added a help func named `getBaseKeyFromKeyEvent(event)` to actions.js. We should use its return value to:
- match against what it currently called `keyOrCode`, but should perhaps then be called `baseKey`
- display the "base" key as part of a menu item shortcut string, and in the custom shortcut UI

src/fontra/client/core/utils.js
ollimeier: `npm test` fails without it: _Exception during run: ReferenceError: navigator is not defined_

src/fontra/views/applicationsettings/panel-shortcuts.js
ollimeier: done.

src/fontra/client/core/actions.js
ollimeier: I moved the sortIndex logic out of registerAction into registerActionInfo.

src/fontra/client/core/actions.js
ollimeier: I added it `sortIndex: actionInfo.sortIndex === undefined ? sortIndex : actionInfo.sortIndex,` and used it for "action.add-component" -> so that "Add Component" and "Decompose Component" are below each other.

ollimeier:
> I added a helper func named `getBaseKeyFromKeyEvent(event)` to actions.js. We should use its return value to:
> 
> * match against what it currently called `keyOrCode`, but should perhaps then be called `baseKey`
> * display the "base" key as part of a menu item shortcut string, and in the custom shortcut UI

Thanks a lot for the helper function. I used it in the shortcuts panel to record the correct `keyOrCode`. But I don't fully understand your two bullet points. I don't know what else I should do with it. 

For example:
`match against what it currently called `keyOrCode`, but should perhaps then be called `baseKey``
-> Do you want me to replace `keyOrCode` with `baseKey` everywhere in the code? Or do you have a specific line of code mind which can be improved with `getBaseKeyFromKeyEvent`?

`display the "base" key as part of a menu item shortcut string, and in the custom shortcut UI`
-> In my opinion, this is already happening, isn't it? The keyOrCode (which should be the same as the baseKey) is used to get the correct icon/symbol via shortCutKeyMap.

I apologize that I am having trouble understanding these instructions.

src/fontra/client/core/utils.js
justvanrossum: Perfect, makes sense!

src/fontra/client/core/actions.js
ollimeier: That was related to customShortCuts being `null`. I changed it to be an empty list instead. The check `if (!shortCut)` is not needed anymore.

src/fontra/client/core/actions.js
ollimeier: As discussed, I moved the sorting out of actions into the UI code (shortcuts panel) – including the discussed refinements to topic indices (00X0-action-topics.XXX) etc.

ollimeier:
> I added a helper func named `getBaseKeyFromKeyEvent(event)` to actions.js. We should use its return value to:
> 
> * match against what it currently called `keyOrCode`, but should perhaps then be called `baseKey`
> * display the "base" key as part of a menu item shortcut string, and in the custom shortcut UI



> I added a helper func named `getBaseKeyFromKeyEvent(event)` to actions.js. We should use its return value to:
> 
> * match against what it currently called `keyOrCode`, but should perhaps then be called `baseKey`
> * display the "base" key as part of a menu item shortcut string, and in the custom shortcut UI

Done.

src/fontra/views/applicationsettings/panel-base.js
ollimeier: I removed the commented out code.

ollimeier:
> Some feedback, it's not a complete review.
> 
> This doesn't support multiple shortcuts for a single action yet, right?



> Some feedback, it's not a complete review.
> 
> This doesn't support multiple shortcuts for a single action yet, right?

There is no UI for adding multiple shortcuts to a single action, BUT you can add multiple shortcuts to a json file and import it. It will then support multiple shortcuts (but it will only display the first one). 

Is this enough for now or do you want me to add a UI for being able to add and visualize multiple shortcuts? Or is this better for a follow up?

ollimeier:
> I noticed that if you assign a shortcut to the ellipse tool, the tool will switch, but the icon doesn't. This is technically unrelated to this PR, but perhaps it stll makes sense to fix it here.



> I noticed that if you assign a shortcut to the ellipse tool, the tool will switch, but the icon doesn't. This is technically unrelated to this PR, but perhaps it stll makes sense to fix it here.



> I noticed that if you assign a shortcut to the ellipse tool, the tool will switch, but the icon doesn't. This is technically unrelated to this PR, but perhaps it stll makes sense to fix it here.

done.

src/fontra/client/core/actions.js
```diff
@@ -61,55 +73,94 @@ export function getActionInfo(actionIdentifier) {
...
-    return "";
-  }
+// add A-Z keys
+for (const key of new Array(26).fill(1).map((_, i) => String.fromCharCode(65 + i))) {
+  shortCutKeyMap[`Key${key}`] = key;
+}
+// add 0-9 keys
+for (let i = 0; i <= 9; i++) {
+  shortCutKeyMap[`Digit${i}`] = `${i}`;
+  shortCutKeyMap[`Numpad${i}`] = `${i}`;
```
justvanrossum, line 115: I think we should do something like this:
```suggestion
  shortCutKeyMap[`Numpad${i}`] = `Numpad-${i}`;
```
The mechanism distinquishes between "regular" digits and "numpad" digits, and I think we should show that difference.

I would also like to discuss the other Numpad codes/char mapping, such as NumpadMultiply.

justvanrossum (commented):
I can't immediately find where the "record" button width is defined, but we need more space...
image

Also, it probably shouldn't overflow like that.

I would also like to discuss the "to blur" or "not to blur" after we've recorded a shortcut. While the field currently does not blur (stays in focus) the contents stay gray, which is confusing.

justvanrossum:
We have layout issue with long labels:

image

I don't think we can use the content size, as this will break alignment across sections. But:
- The label could wrap
- We could make the label column wider

(I think we should do both)

src/fontra/views/applicationsettings/applicationsettings.js
```diff
@@ -0,0 +1,116 @@
+import * as html from "../core/html-utils.js";
+import { ServerInfoPanel } from "./panel-server-info.js";
+import { ShortCutsPanel } from "./panel-shortcuts.js";
+import { ensureLanguageHasLoaded, translate } from "/core/localization.js";
+import { message } from "/web-components/modal-dialog.js";
+
+export class ApplicationSettingsController {
+  static async fromWebSocket() {
```
justvanrossum, line 8: This method name makes no sense, as there is no websocket involved. 

Perhaps the html script should call `.start()`, and the `start()` method should do `await ensureLanguageHasLoaded`.

src/fontra/views/applicationsettings/applicationsettings.js
```diff
@@ -0,0 +1,116 @@
...
+    document.title = `Fontra Application Settings`;
```
justvanrossum, line 9: The title is set in the HTML, and isn't dynamic. Is this line needed?

src/fontra/views/applicationsettings/applicationsettings.js
```diff
@@ -0,0 +1,116 @@
...
+
+    window.addEventListener("keydown", (event) => this.handleKeyDown(event));
+  }
+
+  handleKeyDown(event) {
+    const panel = this.panels[this.selectedPanel];
+    panel?.handleKeyDown?.(event);
+  }
+
+  async messageFromServer(headline, msg) {
```
justvanrossum, line 85: This method can be deleted: we have no live websocket connection to the server.

src/fontra/views/applicationsettings/applicationsettings.js
```diff
@@ -0,0 +1,116 @@
...
+    // don't await the dialog result, the server doesn't need an answer
+    message(headline, msg);
+  }
+
+  handleRemoteClose(event) {
```
justvanrossum, line 90: Same.

src/fontra/views/applicationsettings/applicationsettings.js
```diff
@@ -0,0 +1,116 @@
...
+    //
+  }
+
+  handleRemoteError(event) {
```
justvanrossum, line 94: Same.

src/fontra/views/applicationsettings/panel-base.js
```diff
@@ -0,0 +1,57 @@
...
+    if (event[commandKeyProperty]) {
+      if (event.key == "z") {
+        event.stopImmediatePropagation();
+        event.preventDefault();
+        this.doUndoRedo(event.shiftKey);
+      }
+    }
+  }
+
+  async doUndoRedo(isRedo) {
```
justvanrossum, line 34: Is undo actually hooked up?

src/fontra/views/applicationsettings/panel-base.js
```diff
@@ -0,0 +1,57 @@
...
+      return;
+    }
+    if (isRedo) {
+      undoRecord = reverseUndoRecord(undoRecord);
+    }
+
+    this.setupUI();
+  }
+
+  async postChange(change, rollbackChange, undoLabel) {
```
justvanrossum, line 46: I don't think this is used.

ollimeier:
> We have a layout issue with long labels:
> 
> image
> I don't think we can use the content size, as this will break alignment across sections. But:
> 
> * The label could wrap
> * We could make the label column wider
> 
> (I think we should do both)

Thanks for the feedback. done.

src/fontra/views/applicationsettings/applicationsettings.js
ollimeier: Thanks. done.

src/fontra/views/applicationsettings/applicationsettings.js
ollimeier: done.

src/fontra/views/applicationsettings/applicationsettings.js
ollimeier: done.

src/fontra/views/applicationsettings/panel-base.js
ollimeier: done.

src/fontra/views/applicationsettings/applicationsettings.js
ollimeier: done.

src/fontra/views/applicationsettings/panel-base.js
ollimeier: It's not hooked up. It was pretty much a copy of the fontinfo panel-base.js. I left it almost unchanged, because I was not sure if the plan is to implement such functionality or not. 
I removed it.

src/fontra/views/applicationsettings/panel-base.js
ollimeier: I removed it.

src/fontra/views/applicationsettings/applicationsettings.html
```diff
@@ -0,0 +1,75 @@
...
+    <div class="main-container">
+      <div id="header-container"></div>
+      <div id="panel-container"></div>
+    </div>
+  </body>
+  <script type="module">
+    import { ApplicationSettingsController } from "/applicationsettings/applicationsettings.js";
+
+    async function startApp() {
+      window.applicationSettingsController =
```
justvanrossum, line 69: `start()` doesn't return anything, so this statement should be done in two parts:
- assign `applicationSettingsController`
- call `start()` on `window.applicationSettingsController()`

src/fontra/views/applicationsettings/panel-shortcuts.js
```diff
@@ -0,0 +1,462 @@
...
+}
+.fontra-ui-shortcuts-panel-buttons {
+  display: grid;
+  grid-template-columns: max-content max-content max-content;
+  gap: 1em;
+}
+`);
+
+export class ShortCutsPanel extends BaseInfoPanel {
+  static title = "shortcuts.title";
```
justvanrossum, line 60: I know we're not (yet) doing this in the fontinfo view, but let's use a prefix for all translatable string keys for this view.

```suggestion
  static title = "application-settings.shortcuts.title";
```

(This obviously also requires changes to the language files)

src/fontra/views/applicationsettings/panel-shortcuts.js
```diff
@@ -0,0 +1,462 @@
...
+    input.onchange = async (event) => {
+      const file = event.target.files[0];
+      if (!file) {
+        return;
+      }
+      const reader = new FileReader();
+      reader.onload = async (event) => {
+        const data = JSON.parse(event.target.result);
+        for (const actionIdentifier in data) {
+          setCustomShortCuts(actionIdentifier, data[actionIdentifier]);
```
justvanrossum, line 169: Please add a check here to ensure `actionIdentifier` is an existing one. Else `setCustomShortCuts()` will fail.

src/fontra/views/applicationsettings/panel-shortcuts.js
```diff
@@ -0,0 +1,462 @@
...
+      if (!this.saveShortCuts([shortCutDefinition])) {
+        // if the shortcut is invalid, reset the input field
+        element.innerHTML = getShortCutRepresentation(this.shortCutDefinition);
+      }
+      element.blur(); // remove focus
+      this.pressedKeys = new Set();
+    }
+  }
+
+  recordShortCutKeyup(id, event) {
```
justvanrossum, line 373:
```suggestion
  recordShortCutKeyUp(id, event) {
```

src/fontra/views/applicationsettings/applicationsettings.html
ollimeier: I changed it, but I am not 100% sure if it's right – why I keep this feedback open.

src/fontra/views/applicationsettings/panel-shortcuts.js
ollimeier: Didn't you say we won't check that because it's covered in `setCustomShortCuts` anyway (and good to know if there is something wrong):
`  assert(actionInfo, `unknown actionIdentifier: ${actionIdentifier}`);`
Please don't get me wrong, I am happy to change it. I only want to double check with you if I misunderstood something.

src/fontra/views/applicationsettings/panel-shortcuts.js
ollimeier: done.

src/fontra/views/applicationsettings/panel-shortcuts.js
```diff
@@ -0,0 +1,462 @@
...
+  }
+
+  deleteShortCut(id) {
+    this.saveShortCuts([]);
+    const element = document.getElementById(id);
+    element.innerHTML = "";
+    element.focus();
+  }
+
+  outOfFocus(id) {
```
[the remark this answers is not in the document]
ollimeier: done.

src/fontra/views/editor/editor.js
```diff
@@ -1186,7 +1220,15 @@ export class EditorController {
     };
   }
 
-  setSelectedTool(toolIdentifier) {
+  collapseSubTools(editToolsElement) {
```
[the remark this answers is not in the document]
ollimeier: ok. done.

src/fontra/views/applicationsettings/panel-shortcuts.js
justvanrossum: No. In the current state import will fail completely if it uses a non-existent key. This would happen if we'd rename or delete an action. So, the assert in `setCustomShortCuts()` is right, but the _caller_ can (and should in this case) filter for existing action identifiers.

src/fontra/views/applicationsettings/applicationsettings.html
justvanrossum: You got the await wrong. The constructor is (by definition) not async, but `start()` is (as per `async` keyword.

src/fontra/client/lang/en.json
```diff
@@ -99,6 +115,7 @@
   "menubar.view.select-next-source": "Select next source",
   "menubar.view.select-previous-source": "Select previous source",
   "selection.none": "(No selection)",
+  "server-info.title": "Server Info",
```
justvanrossum, line 118: This one should also get the "application-settings" prefix.

src/fontra/views/applicationsettings/panel-shortcuts.js
ollimeier: Thanks. done.

src/fontra/views/applicationsettings/applicationsettings.html
ollimeier: sorry and thanks. done.

justvanrossum (approved):

PR #1641, opened 2024-09-12
Show nodes and handles from component

ollimeier:
Fixes #1630

This is how it looks like:

Screenshot 2024-09-12 at 09 01 23
Screenshot 2024-09-12 at 09 01 26

### DIFF
--- a/src/fontra/views/editor/visualization-layer-definitions.js
+++ b/src/fontra/views/editor/visualization-layer-definitions.js
@@ -1126,7 +1126,6 @@ registerVisualizationLayerDefinition({
     }
 
     context.fillStyle = parameters.color;
-    context.strokeStyle = parameters.strokeColor;
     for (const pt of glyph.componentsPath.iterPoints()) {
       fillNode(context, pt, cornerSize, smoothSize, handleSize);
       strokeNode(context, pt, cornerSize, smoothSize, handleSize);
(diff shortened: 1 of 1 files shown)


justvanrossum:
Thanks!

Looking at the screenshot, I'm not sure about the stroke treatment: it causes these nodes/handles to be _more_ visible somehow. I would love the nodes to be drawn solid, but quite dimmed, so that they are _just_ visible, but clearly different from the editable nodes/handles.

ollimeier:
> Thanks!
> 
> Looking at the screenshot, I'm not sure about the stroke treatment: it causes these nodes/handles to be _more_ visible somehow. I would love the nodes to be drawn solid, but quite dimmed, so that they are _just_ visible, but clearly different from the editable nodes/handles.

What about this:
Screenshot 2024-09-12 at 10 29 42
Screenshot 2024-09-12 at 10 29 46

justvanrossum:
If you're worried about how visible they are on top of a zone: maybe we should use opacity.

ollimeier:
> If you're worried about how visible they are on top of a zone: maybe we should use opacity.

Thanks for your feedback. I added a minor opacity, added a language key and made some adjustments to the zIndex. 
What do you think?

justvanrossum (approved):
Done! Feel free to merge.

PR #1643, opened 2024-09-12
Fix visual tooltip bug related to shape tool

### LINKED ISSUE #1638: Tooltip off for shape subtools
Please see:

Screenshot 2024-09-11 at 15 52 33

ollimeier:
Fixes #1638

Befor:
Screenshot 2024-09-11 at 15 52 33

After:

https://github.com/user-attachments/assets/d3231c59-c0c2-4101-9614-1bc788620695

### DIFF
--- a/src/fontra/client/core/localization.js
+++ b/src/fontra/client/core/localization.js
@@ -87,6 +87,14 @@ export function translatePlural(key, quantity = 0) {
 export function localizePage() {
   // Translate all elements' data-tooltip that have data-tool attribute
   document.querySelectorAll("[data-tool]").forEach((el) => {
+    if (
+      el.classList.contains("tool-button") ||
+      el.classList.contains("subtool-button")
+    ) {
+      // Skip tool buttons, because they have their own tooltips.
+      // This function causes a visual bug otherwiese.
+      return;
+    }
     const key = el.getAttribute("data-tool");
     el.setAttribute("data-tooltip", translate(key));
   });


justvanrossum:
I wonder if `localizePage()` should leave tooltips alone, and we should perhaps use explicit `translate()` calls for all `data-tooltip` fields.

justvanrossum:
Looking and thinking further: "data-tool" (as opposed to "data-tooltip") is only used by tools in editor.js. We only have to fix those to use data-tooltip with explicit `translate()` calls, and rip this bit of dodgy code out of `localizePage()`.

Maybe `localizePage()` can be deleted completely: it also does some magic with `data-translate-key`, which I don't see used anywhere.

ollimeier:
> Looking and thinking further: "data-tool" (as opposed to "data-tooltip") is only used by tools in editor.js. We only have to fix those to use data-tooltip with explicit `translate()` calls, and rip this bit of dodgy code out of `localizePage()`.
> 
> Maybe `localizePage()` can be deleted completely: it also does some magic with `data-translate-key`, which I don't see used anywhere.

I agree. I was actually very confused to figure out, that translation.js modifies the tooltips.

ollimeier:
> Looking and thinking further: "data-tool" (as opposed to "data-tooltip") is only used by tools in editor.js. We only have to fix those to use data-tooltip with explicit `translate()` calls, and rip this bit of dodgy code out of `localizePage()`.
> 
> Maybe `localizePage()` can be deleted completely: it also does some magic with `data-translate-key`, which I don't see used anywhere.

I removed localizePage + related code and infos.

justvanrossum (approved):

PR #1645, opened 2024-09-12
Bugfix: missing resize handles for one selected component

### LINKED ISSUE #1644: Missing resize handles if only one component is selected
If only a single component is selected, we don’t get the resize handles. They only appear when there are two or more selected.

ollimeier:
Fixes #1644

### DIFF
--- a/src/fontra/views/editor/edit-tools-pointer.js
+++ b/src/fontra/views/editor/edit-tools-pointer.js
@@ -797,9 +797,6 @@ function getTransformHandles(transformBounds, margin) {
 }
 
 function getTransformSelectionBounds(glyph, selection) {
-  if (selection.size <= 1) {
-    return undefined;
-  }
   const selectionBounds = glyph.getSelectionBounds(selection);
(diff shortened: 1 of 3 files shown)


src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -797,9 +797,6 @@ function getTransformHandles(transformBounds, margin) {
 }
 
 function getTransformSelectionBounds(glyph, selection) {
-  if (selection.size <= 1) {
```
justvanrossum, line 800: The reason this is there is to avoid showing resize handles for a single-point selection. You need to distinguish between component and point selection.

ollimeier: A single point will be covered anyway a view lines below.

ollimeier:
```
  if (width == 0 && height == 0) {
    // return undefined if for example only one point is selected
```

PR #1669, opened 2024-09-24
Migrate server info to application settings

### LINKED ISSUE #1668: Migrate Server info to Application settings

ollimeier:
Fixes #1668

### DIFF
--- a/src/fontra/views/applicationsettings/applicationsettings.js
+++ b/src/fontra/views/applicationsettings/applicationsettings.js
@@ -19,8 +19,8 @@ export class ApplicationSettingsController {
 
     for (const panelClass of [
       ShortCutsPanel,
+      ServerInfoPanel,
       // TODO: Add more panels here:
-      // ServerInfoPanel,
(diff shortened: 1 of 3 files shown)


justvanrossum:
Can you post a screenshot of that Fontra menu submenu? I’m away from my computer but would love to see.

ollimeier:
> Can you post a screenshot of that Fontra menu submenu? I’m away from my computer but would love to see.

Sure:
Screenshot 2024-09-24 at 17 32 53

justvanrossum:
This raises the question about how we handle the same situation under “Font”. 

We don’t have many items under Fontra, so perhaps this should be a flat list instead of a submenu.

ollimeier:
> This raises the question about how we handle the same situation under “Font”.
> 
> We don’t have many items under Fontra, so perhaps this should be a flat list instead of a submenu.

Ok, I removed the submenu:

Screenshot 2024-09-25 at 08 55 39

Issue #1650, opened 2024-09-16
non-editor views do not respond to light/dark mode override


justvanrossum:
While they do respond to the system setting, our theme override setting is ignored.

This goes for:
- Font info
- Plugins
- Application settings (#1636)

justvanrossum:
The app settings view now does respond to the theme override, and I think that's because it now contains the theme settings UI...

I think all view html files should contain
```html
    <script type="module" src="/core/theme-settings.js"></script>
```
as part of `<head>`.

PR #1674, opened 2024-09-25
Add 'User settings' to application-settings

### LINKED ISSUE #1673: Migrate Clipboard export user settings to Application settings

ollimeier:
Fixes #1673

We decided to have the following categories as separate section and not bundled:
- Theme settings
- Display Language
- Clipboard export format


ollimeier:
@justvanrossum This is how it looks like:
Screenshot 2024-09-25 at 15 49 03

src/fontra/views/editor/editor.js
```diff
@@ -128,8 +129,7 @@ export class EditorController {
       async (...args) => await this.editListenerCallback(...args)
     );
 
-    this.clipboardFormatController = new ObservableController({ format: "glif" });
-    this.clipboardFormatController.synchronizeWithLocalStorage("fontra-clipboard-");
+    this.clipboardFormatController = clipboardFormatController;
```
justvanrossum, line 132: Suggest to just use `clipboardFormatController` instead of `this.clipboardFormatController`.

justvanrossum (commented):
I'm not super happy with all the "card" repetitions in the panel code and CSS: let's create a follow-up issue to fix that.

Other than that, one small comment, the rest is fine.

src/fontra/views/editor/editor.js
ollimeier: But will the change not break other parts?

justvanrossum: > But will the change not break other parts?

You can do a global search for `.clipboardFormatController`. It looks like it's used by the gear panel and editor.js, and nothing else.

ollimeier: ok, done.

justvanrossum (commented):
LGTM, except I'd like the "Server info" panel to be the last in both the menu and the panel items sidebar, as it's the least important one.

src/fontra/views/applicationsettings/applicationsettings.html
```diff
@@ -6,6 +6,7 @@
     <link href="/css/tooltip.css" rel="stylesheet" />
     <title>Fontra Application Settings</title>
     <script type="module" src="/web-components/modal-dialog.js"></script>
+    <script type="module" src="/web-components/grouped-settings.js"></script>
```
justvanrossum, line 9: Please add
```
<script type="module" src="/core/theme-settings.js"></script>
```
after this line, as per #1650.

(While it works without this, given the inclusion of the theme panel, it is good practice and a necessary example for other views.)

ollimeier: ok, done.

PR #1678, opened 2024-09-25
Adding editor-behavior-panel to application settings

### LINKED ISSUE #1677: Rect-select live modifier keys should go to a new "Editor behavior" panel under Application settings

ollimeier:
Fixes #1677


src/fontra/client/core/experimental-features.js
```diff
@@ -0,0 +1,11 @@
+import { ObservableController } from "./observable-object.js";
+
+export const experimentalFeaturesController = new ObservableController({
```
justvanrossum, line 3: I would love to have this bundled with the clipboard settings as `applicationSettingsController`, even if that means we're changing the localStorage.

And either way, I'd like to drop the "Experimental" bit.

So, `clipboard-format.js` and `experimental-features.js` should together become `application-settings.js`, which should export a single `applicationSettingsController` object.

src/fontra/client/core/experimental-features.js
```diff
@@ -0,0 +1,11 @@
...
+  scalingEditBehavior: false,
```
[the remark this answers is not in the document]
justvanrossum: `scalingEditBehavior` can be a property on both tools objects, its value being `false` on `PointerTool`, and `true` on `PointerToolScale`.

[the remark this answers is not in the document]
justvanrossum: I assume the same can be done for `quadPenTool` (maybe renamed as `drawQuadratic`) for the two pen tools.

src/fontra/client/core/experimental-features.js
ollimeier: Thanks for the great feedback. I am totally with you. I don't wanted to change too much, while we keep the 'gear'-sidebar. But now this is done.

src/fontra/client/core/experimental-features.js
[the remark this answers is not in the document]
ollimeier: I changed the way 'scalingEditBehavior' works and removed scalingEditBehavior and quadPenTool from controller and gear panel. I did not rename quadPenTool, because it's redundant anyway as it's a new separate tool. 

Even though it feels like I'm done with it, I keep this feedback open because I'm not 100% sure I've managed to change it the way you want it to be.

src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -770,11 +771,11 @@ export class PointerToolScale extends PointerTool {
 
   activate() {
     this.setCursor();
-    this.sceneController.experimentalFeatures.scalingEditBehavior = true;
+    this.scalingEditBehavior = true;
```
justvanrossum, line 774: This should not be a dynamic property: in `PointerTool` `this.scalingEditBehavior` should always be false, and in `PointerToolScale` it should always be true. Since we should not write to this property, perhaps it should be a getter, like how it's done in the `curveType` getter in the pen tools.

src/fontra/views/editor/panel-user-settings.js
```diff
@@ -92,18 +93,18 @@ export default class UserSettingsPanel extends Panel {
...
-          key: "scalingEditBehavior",
-          displayName: "Scaling edit tool behavior",
-          ui: "checkbox",
-        },
-        {
-          key: "quadPenTool",
-          displayName: "Pen tool draws quadratics",
-          ui: "checkbox",
-        },
+        // {
```
justvanrossum, line 98: This can now be cleaned up, right?

src/fontra/views/editor/edit-tools-pointer.js
ollimeier: Ah, yes, you're right. That's much better. Done.

src/fontra/views/editor/panel-user-settings.js
ollimeier: I removed it.

src/fontra/client/core/experimental-features.js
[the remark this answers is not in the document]
ollimeier: I assume it's save to close this as we have solved it with [ee0f88c](https://github.com/googlefonts/fontra/pull/1678/commits/ee0f88c15d8e58f49f22e8022f0daae59cb5e187)

justvanrossum (approved):

PR #1687, opened 2024-10-01
Add actions to transformation panel

### LINKED ISSUE #1683: Some transformation panel buttons should use actions.js, so they become "shortcuttable"
Specifically:
- the align objects buttons
- the distribute objects buttons
- the path operations (remove overlap, etc.)

ollimeier:
Fixes #1683

### DIFF
--- a/src/fontra/views/editor/panel-transformation.js
+++ b/src/fontra/views/editor/panel-transformation.js
@@ -413,7 +413,7 @@ export default class TransformationPanel extends Panel {
 
     const topic = "0070-action-topics.selection-transformations";
 
-    const registerActionsDistribute = [
+    const registerActions = [
       ["align.left", alignLeft],
       ["align.center", alignCenter],
       ["align.right", alignRight],
@@ -423,7 +423,7 @@ export default class TransformationPanel extends Panel {
       ["distribute.horizontally", distributeHorizontally],
       ["distribute.vertically", distributeVertically],
     ];
-    for (const [keyPart, moveDescriptor] of registerActionsDistribute) {
+    for (const [keyPart, moveDescriptor] of registerActions) {
       registerAction(
         `action.selection-transformation.${keyPart}`,
         { topic, titleKey: `sidebar.selection-transformation.${keyPart}` },
--- a/src/fontra/views/editor/panel-transformation.js
+++ b/src/fontra/views/editor/panel-transformation.js
@@ -1,3 +1,4 @@
+import { registerAction } from "../core/actions.js";
 import { ChangeCollector, applyChange, consolidateChanges } from "../core/changes.js";
 import { EditBehaviorFactory } from "./edit-behavior.js";
 import Panel from "./panel.js";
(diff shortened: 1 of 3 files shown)


justvanrossum:
We have a fundamental problem here: the actions only get registered once the _panel_ gets updated/shown. This means that if you _hide_ the panel, then reload, and shortcut does not work.

We somehow need to refactor this, but I need to think a bit more about how to best approach it.

Also: there is an existing bug where you get a console error message if you use one of the actions while there is no selection:
image

ollimeier:
> We have a fundamental problem here: the actions only get registered once the _panel_ gets updated/shown. This means that if you _hide_ the panel, then reload, any shortcuts do not work.
> 
> We somehow need to refactor this, but I need to think a bit more about how to best approach it.
> 
> Also: there is an existing bug where you get a console error message if you use one of the actions while there is no selection: image

I am not able to reproduce the error. Can you please let me know if this error still exists after my latest changes?

justvanrossum:
Yes, the error still exists, but only in the "remove overlaps" action/button: make sure there is _no_ selected glyph in the canvas, then push the "remove overlaps" button. (All other buttons appear to be fine in this case.)

ollimeier:
> Yes, the error still exists, but only in the "remove overlaps" action/button: make sure there is _no_ selected glyph in the canvas, then push the "remove overlaps" button. (All other buttons appear to be fine in this case.)

Ah, I see. Thanks. I fixed it.

src/fontra/views/editor/panel-transformation.js
```diff
@@ -101,6 +102,46 @@ export default class TransformationPanel extends Panel {
       skewY: 0,
       customDistributionSpacing: null,
     };
+    this.registerActions();
+  }
+
+  registerActions() {
+    const topic = "0070-action-topics.selection-transformations";
+
+    const registerActions = [
```
justvanrossum, line 111: I'd prefer a different vairable name here. I think I'd like `moveActions`.

src/fontra/views/editor/panel-transformation.js
```diff
@@ -101,6 +102,46 @@ export default class TransformationPanel extends Panel {
...
+    ];
+    for (const [keyPart, moveDescriptor] of registerActions) {
+      registerAction(
+        `action.selection-transformation.${keyPart}`,
+        { topic, titleKey: `sidebar.selection-transformation.${keyPart}` },
+        () => this.moveObjects(moveDescriptor)
+      );
+    }
+
+    const registerActionsPathOperations = [
```
justvanrossum, line 129: And for this one I'd prefer `pathActions`.

justvanrossum (commented):
Two minor nits, then this is done I think.

src/fontra/views/editor/panel-transformation.js
ollimeier: done.

src/fontra/views/editor/panel-transformation.js
ollimeier: done.

justvanrossum (approved):

PR #1691, opened 2024-10-03
Increase zIndex of point visualization to bring it above anchor names

### LINKED ISSUE #1688: Anchors labels hide points and contours
when anchors label are turn on they are above the contours and points and hide them, would be nice to put them in the background or in between the fill grey and the contour/points
image (6)

ollimeier:
Fixes #1688 

Still as draft PR, because I am not sure about this change.

It now looks like this:
Screenshot 2024-10-03 at 15 43 55


justvanrossum:
Hm yes, I agree this doesn't look optimal, either. Let's discuss next time we talk.

justvanrossum:
I'll close this for now, we might come back to it later with a better solution.

PR #1692, opened 2024-10-03
Use overshootColor from cjk.design.frame

### LINKED ISSUE #1690: Let's use a non-gray color for the alignment zones
Let's try using the light blue we use for the overshoot area in the CJK design frame.

ollimeier:
Fixes #1690 

Still draft PR, because I like the grey more. 

This is how it looks with the cjk.design.frame overshootColor:
Screenshot 2024-10-03 at 15 54 35

Screenshot 2024-10-03 at 15 54 41

### DIFF
--- a/src/fontra/views/editor/visualization-layer-definitions.js
+++ b/src/fontra/views/editor/visualization-layer-definitions.js
@@ -179,8 +179,8 @@ registerVisualizationLayerDefinition({
   defaultOn: true,
   zIndex: 100,
   screenParameters: { strokeWidth: 1 },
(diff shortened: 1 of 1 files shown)


justvanrossum:
What do we think about the color? Is it perhaps too bright, or is it ok?

justvanrossum:
(To be honest, I think this is a huge improvement, I like the blue color a lot in this context.)

ollimeier:
> (To be honest, I think this is a huge improvement, I like the blue color a lot in this context.)

OK then. I have marked it as 'ready for review'. Feel free to merge it.

justvanrossum (approved):

PR #1695, opened 2024-10-03
Remove extensions menu and delete plugins view

### LINKED ISSUE #1694: Remove "Extensions" menu, delete `plugins` view

ollimeier:
Fixes #1694 

I kept "menubar.extensions" in the translation json files, because I thought we may add it again to the menu bar one day – and if so, why not keep the Chinese translation? Or should I remove it?


justvanrossum (commented):
You need to update pyproject.toml as well.

PR #1704, opened 2024-10-09
Magic contour selection

GaetanBaehr:
A pull request for magic contour selection with the pointer tool.

By holding the cmd (metaKey) modifier key:

- Clicking anywhere in the glyph editor view selects the nearest contour.
- Holding shift and clicking somewhere else adds the nearest contour to the selection.
- Dragging the mouse changes the selection to the nearest contour.
- Dragging while holding shift adds the nearest contours to the selection.

Note:
In `path-hit-tester.js` -> `PathHitTester` -> `findNearest`, I commented out line 65:
`results = results.filter((hit) => hit.t != 0 && hit.t != 1);`

This line restricts findNearest to only work when, in the case of a rectangular shape, the cursor's x value is between the rectangle's xmin and xmax, or the y value is between its ymin and ymax. 
This function is used exclusively with the power ruler tool and now the pointer tool. Commenting out this line seems to have no effect on the power ruler tool.

### DIFF
--- a/src/fontra/client/core/path-hit-tester.js
+++ b/src/fontra/client/core/path-hit-tester.js
@@ -62,7 +62,7 @@ export class PathHitTester {
       }
     }
 
-    results = results.filter((hit) => hit.t != 0 && hit.t != 1);
+    // results = results.filter((hit) => hit.t != 0 && hit.t != 1);
     results.sort((a, b) => a.d - b.d);
     return results[0];
   }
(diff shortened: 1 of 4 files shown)


justvanrossum (commented):
This is not yet a solid code review, but I found one small issue.

However, there is a bigger thing missing: whenever the "magic behavior" is active (regardless of how we decide to activate it), the user should see what will be selected once they click. In other words, this is missing "hover" behavior. It should set `sceneController.hoverSelection` appropriately while hovering.

davelab6:
@justvanrossum is this stale?

justvanrossum:
Yeah, it needs work to be revived. @GaetanBaehr: I will close this now, but feel free to reopen or open a new one once you get back to it.

PR #1705, opened 2024-10-09
Don't show transformation handles if readOnly or glyph locked

### LINKED ISSUE #1702: Dialog's default button doesn't respond to enter key in some cases
Open a glyph in read-only mode (either lock it, or open a .ttf or .glyphs file).

Select some points. If you try to drag the points, you get the `Can’t edit glyph “xxx”` dialog. You can dismiss it with the Escape key or the Enter key.

Now, with the same selection, try to scale or rotate the selection with the bounding box handles. The same dialog appears. You can dismiss it with the Escape key, but _not_ with the Enter key, despite the Okay button being the default.

Does the dialog not get the proper focus in this case?

Or is some other event handler stealing the event?

Is this issue specific to the transform-with-bounds-handles, or is there a more general cause?

ollimeier:
Fixes #1702

Even though I don't know what the issue was caused by, it's already unnecessary to show the transformation handles if the file is readOnly or the glyph is locked – therefore I added code which don't show the handles if readOnly or glyph is locked.

### DIFF
--- a/src/fontra/views/editor/edit-tools-pointer.js
+++ b/src/fontra/views/editor/edit-tools-pointer.js
@@ -720,6 +720,14 @@ registerVisualizationLayerDefinition({
     if (!model.showTransformSelection) {
       return;
     }
+
+    if (
+      !!positionedGlyph.varGlyph?.glyph.customData["fontra.glyph.locked"] ||
+      model.fontController.readOnly
+    ) {
+      return;
+    }
+
     const transformBounds = getTransformSelectionBounds(
       positionedGlyph.glyph,
       model.selection
--- a/src/fontra/views/editor/edit-tools-pointer.js
+++ b/src/fontra/views/editor/edit-tools-pointer.js
@@ -128,6 +128,7 @@ export class PointerTool extends BaseTool {
         !!rotationHandle
       );
       delete sceneController.sceneModel.clickedTransformSelectionHandle;
+      initialEvent.preventDefault();
       return;
     }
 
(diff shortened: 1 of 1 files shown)


justvanrossum:
Even though I agree this fix is probably what we want, I still want to know what causes the behavior described in the issue.

ollimeier:
@justvanrossum I sadly cannot explain why, but [89792aa](https://github.com/googlefonts/fontra/pull/1705/commits/89792aa65a01fb23fce1747c0d55a9ddbbbcd9ec) fixes the enter key issue.

For reference: 

https://github.com/user-attachments/assets/33da292a-bae6-4527-9636-2808bcbc888b

src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -720,6 +721,14 @@ registerVisualizationLayerDefinition({
     if (!model.showTransformSelection) {
       return;
     }
+
+    if (
+      !!positionedGlyph.varGlyph?.glyph.customData["fontra.glyph.locked"] ||
```
justvanrossum, line 726: I played with this, but this is not the right thing: it only hides the visualization, but you still see the cursor change, and you can still try to drag the (now invisible) handles.

justvanrossum (commented):
Let's only do the `initialEvent.preventDefault()` change, as fixing the other thing I noticed is probably more work than it's worth.

src/fontra/views/editor/edit-tools-pointer.js
ollimeier: Thanks for the feedback and testing. I did not notice that behavior. I removed this part of the code.

PR #1726, opened 2024-10-16
new pen tool icon

GaetanBaehr:
Here is a new icon for the pen tool using a pencil
image


justvanrossum:
If you could run `python icons/export-icons-as-svg.py` that would be great (it rebuilds the SVG icons from the .ufo)

GaetanBaehr:
I ran it

justvanrossum:
With this new design, I think we should rename the glyphs/svg files to:

- `pencil`
- `pencilquad`

(This would also be a change in the source code.)

GaetanBaehr:
should I rename them?

justvanrossum:
Let's leave it for now. I'll merge, and we can rename later if we want.

Issue #1466, opened 2024-06-18
Improve UI of font sources panel


ollimeier:
It works, looks good for now and is consistent with other parts of the panel, but we may want to consider to improve it in some points. If possible keep this in mind while working on other items. Also keep in mind that this may also be used for avar2 editing.

justvanrossum:
After #1711, I think we should at least increase the minimum height of the source boxes somewhat.

Other ideas:
- start with all boxes expanded, and have a way for the user to say "collapse everything" (say: alt-click on the chevron)

ollimeier:
@justvanrossum @JeremieHornus:

Here a quick an dirty idea for this issue above, a 'one-liner' with the most important infos when the cards are folded:

https://github.com/user-attachments/assets/09e3f093-b1b3-41a3-bc7d-a26a43c99a4e


It's far from a PR, but if you're interested in the code, it's this branch:
https://github.com/googlefonts/fontra/tree/issue-1466-improve-ui-font-sources-panel

justvanrossum:
Good thinking. This sounds like a good path forward.

(I'd still like to see all panels expanded by default.)

Maybe only the name and the location should remain. I'm not sure why we'd include the italic angle but none of the line metrics.

The location could be formatted like this:
```
Vertical=100, Horizontal=200, Weight=400
```

We should probably make the location "dense" before displaying: ensure it includes axes that are at the default.

Perhaps we can leave out the labels: "Regular" says enough, we don't need "Name: Regular".

For the location string: we could format axes that are at the default differently, maybe make them bold or italic. Or the opposite: make them gray.

ollimeier:
Some further notes from a discussion with @justvanrossum : Make the whole menu of the panel sticky in both directions: vertically AND horizontally. Only the right content with the cards should move in both directions.

ollimeier:
@justvanrossum 

This is what I changed:
- Make the menu stick to the top left corner
- refactor the current code
- remove italic info from oneliner. Only keep name and the location.
- location formatted like this: Vertical=100, Horizontal=200, Weight=400
- it now includes axes that are at the default
- I removed the labels 'Name' and 'Location'
- I made the default values of the location gray

This is the draft PR: https://github.com/googlefonts/fontra/pull/1739

PR #1729, opened 2024-10-16
Add cross axis mapping panel

### LINKED ISSUE #1657: [avar-2] Design, implement and test basic authoring UI
We need a new panel under Font Info, named "Cross-axis mapping".

It will be "card-based", like the axis panel, and like the axis panel, each card should be draggable to manually tweak the order of cards.

Each card has two text labels:
- description
- groupDescription

There are two location slider sets, one for "inputLocation" and one for "outputLocation. I think they should be lined up next to each other.

Like the source panel cards, they should be collapsable (needed when where are many axes).

This is an insane font to test an insane number of axes: https://github.com/googlefonts/amstelvar-avar2/blob/main/Sources/Roman/AmstelvarA2-Roman_avar2.designspace

We probably need a checkbox for each axis to explicitly say "this axis participates in the mapping". Maybe this is hard to align properly unless it would be an optional feature of the `<designspace-location>` webcomponent. Perhaps we need a different collection of axis sliders instead of `<designspace-location>`, as per axis we need:
- a label (the axis name)
- an input slider
- an output slider
- a checkbox

All on a single row in the layout.

Perhaps it is useful to first make a simple wireframe sketch.

ollimeier:
Fixes #1657 

I guess this is good enough for a first draft PR to get feedback.

Note:
- you can open/close all cards with altKey + click the arrow.
- by default all cards are open (this differs from the sources panel, but is based on: https://github.com/googlefonts/fontra/issues/1466#issuecomment-2405928546)

There are some known issue:
- Reordering a card shows a wrong height of the card if closed (maybe a good follow up, because I am not sure if this should be fixed within `setupSortableList`)
- when editing a cross-axis mapping, the UI gets reloaded. This is not a big deal, BUT if one or more cards are closed, it's irritating, because these closed cards will be open afterwards.


ollimeier:
Here a screenshot of how it looks currently:
Bildschirmfoto 2024-10-16 um 16 38 42

And this is not a mockup anymore ;)

justvanrossum:
> Here a screenshot of how it looks currently

Great!

Some visual feedback:
- Please reduce the space between the slider and its checkbox (could be similar to the space between the numeric field and the slider). This should make it clearer that the checkbox belongs to the slider to its left
- Let's leave out the "incl." label (if could get a tooltip, like I suggested earlier)
- Please increase the space between the "Input/output location" label and the sliders
- The axis label should be `axis.name`, not `axis.tag` (see the Axes panel, but also the designspace-navigation panel)

Behavior:
- When the slider gets moved back to the default, this should _not_ imply unchecking the checkbox. The checkbox is needed _because_ there is not a 1-on-1 relationship between "axis is at default" and "axis participates".

Question:
- Undo doesn't seem to work?

> when editing a cross-axis mapping, the UI gets reloaded. This is not a big deal, BUT if one or more cards are closed, it's irritating, because these closed cards will be open afterwards.

That actually does make it a big deal. Why is it necessary to reload the UI on every change?

justvanrossum:
Layout structure:

I'm not confortable with reusing designspace-location this way, and the changes to the designspace-location web-component. The labels and checkboxes now have their own grid, and it will be hard to keep this perfectly in sync with the sliders grid. It should be a single grid:

```
+-----------+----------------+----------+-----------------+----------+
|           | Input location            | Output location            |
+-----------+----------------+----------+-----------------+----------+
| axis name | range-slider   | checkbox | range-slider    | checkbox |
+-----------+----------------+----------+-----------------+----------+
| axis name | range-slider   | checkbox | range-slider    | checkbox |
+-----------+----------------+----------+-----------------+----------+
| etc.      |                |          |                 |          |
```

ollimeier:
> +-----------+----------------+----------+-----------------+----------+
> |           | Input location            | Output location            |
> +-----------+----------------+----------+-----------------+----------+
> | axis name | range-slider   | checkbox | range-slider    | checkbox |
> +-----------+----------------+----------+-----------------+----------+
> | axis name | range-slider   | checkbox | range-slider    | checkbox |
> +-----------+----------------+----------+-----------------+----------+
> | etc.      |                |          |                 |          |

mmmhhh, I am not sure if I understand this correctly. 
Currently it looks like this:

```
+------+-----------+-----------------+----------+-----------------+----------+------+
| Icon |           | Description     |          | Group Descript. |          | Icon |
+------+-----------+-----------------+----------+-----------------+----------+------+
|      |           | Input location  |          | Output location |          |      |
+------+-----------+-----------------+----------+-----------------+----------+------+
|      | axis name | number | slider | checkbox | number | slider | checkbox |      |
+------+-----------+-----------------+----------+-----------------+----------+------+
|      | axis name | number | slider | checkbox | number | slider | checkbox |      |
+------+-----------+-----------------+----------+-----------------+----------+------+
| etc. |           |                 |          |                 |          |      |

```
I don't fully understand how it should look like. When we want to keep the same look and feeling. Sorry.

ollimeier:
We also discuss an idea to have one or two preview glyphs for each card to get a visual representation of what we specify with input and output location. Should be a follow up.

ollimeier:
Hi @justvanrossum, I will set this PR to 'Ready for review' in a bit.

I was able to address all feedback:
- reduce the space between the slider and its checkbox -> done.
- leave out the "incl." -> done. 
- increase the space between the "Input/output location" label and the sliders -> done.
- axis label should be axis.name, not axis.tag -> done.
- moving slider back to default, should not imply unchecking the checkbox -> done.
- Undo doesn't seem to work? -> works now. done.
- any editing caused a reload of the UI -> not anymore. done.
- don't use designspace-location -> done.
- folded card without input/output headlines -> done.
- wrong draggable card height if a card is folded -> works now.
- remove 'Unnamed' string for new added mapping -> done.
- fix tooltip for checkboxes -> done.

justvanrossum:
Great, thank you!

justvanrossum:
I think found a bug: if you move a slider that is not yet participating *once*, then
1. the default value (or 0?) is written to the file instead of the slider value
2. undo seems to contain an empty step? you have to undo twice before the undo takes place

(FWIW, this is not related to the designspace backend: if I edit a .fontra project, the same thing happens)

justvanrossum:
Request: please swap the Group description with the Description, so the Group description comes first.

ollimeier:
> Request: please swap the Group description with the Description, so the Group description comes first.

Agree. I changed it.

ollimeier:
> I think found a bug: if you move a slider that is not yet participating _once_, then
> 
> 1. the default value (or 0?) is written to the file instead of the slider value
> 2. undo seems to contain an empty step? you have to undo twice before the undo takes place
> 
> (FWIW, this is not related to the designspace backend: if I edit a .fontra project, the same thing happens)

Thanks a lot for finding this. I can reproduce it. I'll have a look.

ollimeier:
> I think found a bug: if you move a slider that is not yet participating _once_, then
> 
> 1. the default value (or 0?) is written to the file instead of the slider value
> 2. undo seems to contain an empty step? you have to undo twice before the undo takes place
> 
> (FWIW, this is not related to the designspace backend: if I edit a .fontra project, the same thing happens)

Thanks once again for finding these issues. They were related to each other. The last commit fixes both.

ollimeier:
@justvanrossum I noticed, that the value written to the file has quite a lot of digits after the decimal point:
Screenshot 2024-10-22 at 09 12 21

Should we limit this to two?

justvanrossum:
> Should we limit this to two?

I don't think it's a problem, let's keep the digits.

justvanrossum:
It seems the width of the slider column seems to depend on the with of the description label:

image

It would be better if the sliders had the same widths.

ollimeier:
> It seems the width of the slider column seems to depend on the with of the description label:
> 
> image
> It would be better if the sliders had the same widths.

Once you see it, you cannot unsee it – right?! Thanks for finding this. I simply have not seen it till now. I made it `minmax(200px, 300px)`.

ollimeier:
While editing the width of the column, I was wondering if it would make sense to use 'Group' instead of 'Group description' for the label, because 'Group description' is quite long. Any thoughts about that?

justvanrossum:
> use 'Group' instead of 'Group description'

Hm, not sure. Let me think about it. The whole thing is a bit weird in the .designspace format.

What helps a little bit is to change the `grid-template-columns` to `min-content auto` in `.fontra-ui-font-info-cross-axis-mapping-panel-column`: that keeps the space for the label to a minimum.

At the same time, the text entry box should be allowed to go over the checkbox columns, so I think each description label+input could be two columns wide.

justvanrossum:
With a narrow viewport, the left text input bleeds into the next column instead of shrinking:

image

Perhaps the label should also be allowed to shrink.

justvanrossum:
The label + input baseline should be aligned:
image

justvanrossum:
image

This phrasing is awkward, but I'm also not sure what this tooltip is helping at all. Maybe remove it? But if you want to keep it, it would be slightly better as "Specify axes and their values for the input location". (Same for output.)

image

Suggest: "When checked, this axis participates in the mapping"

ollimeier:
> image
> This phrasing is awkward, but I'm also not sure what this tooltip is helping at all. Maybe remove it? But if you want to keep it, it would be slightly better as "Specify axes and their values for the input location". (Same for output.)
> 
> image
> Suggest: "When checked, this axis participates in the mapping"

Thanks. I removed and change it.

ollimeier:
> The label + input baseline should be aligned: image

Thanks. That's weird because I use `labeledTextInput` I wouldn't expect to have to touch it.

ollimeier:
> > The label + input baseline should be aligned: image
> 
> Thanks. That's weird because I use `labeledTextInput` I wouldn't expect to have to touch it.

`display: grid;` changes the visual position of the baseline :(

ollimeier:
> > > The label + input baseline should be aligned: image
> > 
> > 
> > Thanks. That's weird because I use `labeledTextInput` I wouldn't expect to have to touch it.
> 
> `display: grid;` changes the visual position of the baseline :(

I looked into the axes-panel, and it looks like you fixed this 'baseline-position'-issue with:
```
  align-items: start;
  align-content: start;
```
 It is better than before but still not perfect. But the question is, is this good enough?

src/fontra/client/core/ui-utils.js
```diff
@@ -94,13 +94,19 @@ function didReorder(a, b) {
   return false;
 }
 
+export function checkboxWithoutLabel(controller, key, options) {
```
justvanrossum, line 97: Please get rid of this function, and just call `labeledCheckbox()` with `null` as the first argument (*not* `false`).

Unless you really prefer a separate function, but then please call it `unlabeledCheckbox` instead. And same: please use `null` instead of `false`.

(I've seen you do this before (pass "false" to say "no value"), but that's never nice or conventional. I'd even say it's incorrect. If value is optional, it should be replacable with `null` or `undefined`, but never with `false`.)

ollimeier:
> > I think this is now good to go. Let's merge! 🎉
> 
> Sadly not. There is still a visual bug: Screenshot 2024-10-23 at 16 53 20

I fixed it. I will merge it now :)

PR #1740, opened 2024-10-23
Make the sidebar navigation match the width of the content

### LINKED ISSUE #1717: App settings (and Font Info?) Title sidebar width should be flexible
Its current width is too tight for some translations:

Image

ollimeier:
Fixes #1717

### DIFF
--- a/src/fontra/views/applicationsettings/applicationsettings.html
+++ b/src/fontra/views/applicationsettings/applicationsettings.html
@@ -19,6 +19,8 @@
         padding: 0em;
         box-sizing: border-box;
         gap: 0em;
+        width: 100vw;
+        height: 100vh;
       }
 
       #header-container {
@@ -54,7 +56,7 @@
       }
 
       #panel-container {
-        /* overflow: scroll; */
+        overflow: auto;
       }
     </style>
   </head>
--- a/src/fontra/views/fontinfo/fontinfo.html
+++ b/src/fontra/views/fontinfo/fontinfo.html
@@ -13,10 +13,12 @@
       }
       .main-container {
         display: grid;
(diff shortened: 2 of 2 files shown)


ollimeier:
@justvanrossum  I played around with the CSS and I don't think it's possible to find a solution with CSS only (to make the left sidebar navigation fixed in position while the right content is moved to the right depending on the navigation width). I tried it first with a reduce demo page, here attached as a zip file:
[test_sticky_nav.zip](https://github.com/user-attachments/files/17486805/test_sticky_nav.zip)

The code above works, but is a little hacky. To figure out the actual width of the sidebar navigation we need to execute our code at the very end if everything has been rendered already. Usually 'onload' would work. But in our case, with the translation, this does not work. Why I added a timeout.

ollimeier:
@justvanrossum Here a demo:

https://github.com/user-attachments/assets/2607d5f6-a57f-4b44-94d1-1c733530431c

justvanrossum:
Here is my attempt:
[test_sticky_nav_just.zip](https://github.com/user-attachments/files/17491483/test_sticky_nav_just.zip)
This works but I wonder how well it will work in the real Fontra context.

It must be possible to do this with 100% CSS, using CSS grid as the basis.

ollimeier:
@justvanrossum Thanks a lot for your zip file with the necessary CSS code. I feel so bad, I was not able to solve it. However: It works with pure CSS, here a short demo:

https://github.com/user-attachments/assets/b7609019-94d1-42f8-9b5a-5f300093396a

I tested it in Chrome, Safari and Firefox and I am going to set it to 'Ready for review' in a second.

PR #1741, opened 2024-10-23
Rebuild languages with arg --wip

ollimeier:
Draft PR -> because this is not an issue nor discussed with you @justvanrossum 

While working on several different issues, I noticed, that when I want to update the languages it takes every row from the google sheet, even those that are not yet fully discussed, because it's still 'WIP'.

This is why I added the possibility to add the argument `--wip`:
`python ./scripts/rebuild_languages.py --wip`

Only if `--wip` is set, it will take the rows which are not finally discussed. 
Or in other words, without `--wip` it will ignore the WIP rows.


scripts/rebuild_languages.py
```diff
@@ -9,7 +10,8 @@
 from urllib.request import urlopen
 
 
-def prettier(path):
+def prettier(path, repoDir):
```
justvanrossum, line 13: If you leave `repoDir` as a global var, this chnge isn't necessary.

Suggest to leave all constants as global vars, I think this is about:
- repoDir
- langDir
- localizationJSPath
- languagesBlock, but rename to languagesBlockTemplate, and use `languagesBlockTemplate.format(...)`
- languageSourceTemplate

scripts/rebuild_languages.py
```diff
@@ -33,104 +35,120 @@ def downloadSheet(url):
...
+                langEn=languageInEnglish,
+                langLang=languageInLanguage,
+                status=languageStatus,
+            )
         )
+
+        languageStrings[languageCode] = strings = {}
+
+        for row in rows:
+            if not wip and row[0] == "WIP":
```
justvanrossum, line 81:
```suggestion
            if wip or "WIP" not in row[0]:
```
1. make the meaning clearer
2. allow the comment collumn to contain info beyond WIP

We could also say "WIP" must be at the beginning of the cell, if general containment is too loose.

scripts/rebuild_languages.py
```diff
@@ -33,104 +35,120 @@ def downloadSheet(url):
     return list(reader)
 
 
-languageSpreadsheetURL = (
-    "https://docs.google.com/"
-    "spreadsheets/d/1woTU8dZCHJh7yvdk-N1kgQBUj4Sn3SdRsbKgn6ltJQs/"
-)
+def main(wip):
```
justvanrossum, line 38: Please rename this variable to `includeWorkInProgress`.

scripts/rebuild_languages.py
```diff
@@ -33,104 +35,120 @@ def downloadSheet(url):
...
-    print("writing", languagePath)
-    lines = []
-    for k, v in sorted(strings.items()):
-        lines.append(f"  {jsonDump(k.strip())}: {jsonDump(v)},")
-    stringsBlock = "\n".join(lines)
-    languageSource = languageSourceTemplate.format(
-        stringsBlock=stringsBlock, url=languageSpreadsheetURL
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
```
justvanrossum, line 147: Suggest to move the argparse code to `main()` itself.

scripts/rebuild_languages.py
justvanrossum: Maybe rename the command line flag to `--include-wip` to make this more explicit.

scripts/rebuild_languages.py
```diff
@@ -33,104 +35,120 @@ def downloadSheet(url):
...
+        description="Script to rebuild the language files from the Google Sheet"
+    )
+    parser.add_argument(
+        "--wip", action="store_true", required=False, help="Flag include WIP rows"
```
justvanrossum, line 151: I think you can omit `required=False`, 99% sure that's the default.

Please make the help text say:
"Include work-in-progress strings, as marked with 'WIP' in the the first column."

scripts/rebuild_languages.py
ollimeier: Thanks. Done.

scripts/rebuild_languages.py
ollimeier: done.

scripts/rebuild_languages.py
ollimeier: done.

scripts/rebuild_languages.py
ollimeier: done.

scripts/rebuild_languages.py
```diff
@@ -33,104 +56,103 @@ def downloadSheet(url):
...
-        if not key.strip():
-            continue
-        string = row[columnIndex]
-        if not string or string == "-":
-            string = languageStrings["en"].get(key, "!missing!")
-        strings[key] = string
+        languageStrings[languageCode] = strings = {}
 
+        for row in rows:
+            if not includeWorkInProgress and "WIP" in row[0]:
```
justvanrossum, line 112: I misread this the first time around. Still I'd like to rephrase for clarity (at least this way it reads more naturally to me):
```suggestion
            if "WIP" in row[0] and not includeWorkInProgress:
```

scripts/rebuild_languages.py
```diff
@@ -8,8 +9,30 @@
...
+// Don't edit this file: it is generated by scripts/rebuild_languages.py
+// The strings are maintained here: {url}
+export const strings = {{
+{stringsBlock}
+}};
+"""
+
 
 def prettier(path):
+
```
justvanrossum, line 35: Why the new blank line?
```suggestion
```

scripts/rebuild_languages.py
ollimeier: I am closing this, as we solved it a little bit different.

justvanrossum (approved):

PR #1742, opened 2024-10-23
Shortcuts for previous/next reference font

### LINKED ISSUE #1720: Enable setting shortcuts for selecting previous/next Reference Font
Since we now have the ability to set customized shortcuts, it would be helpful to be able to set shortcuts for selecting previous/next Reference Font. 

This would fulfill the functions of some other proposed Reference Font related features, like #1067 or #1001 , making those issues less urgent.

ollimeier:
Fixes #1720

This is how it looks like in the menu bar:
Screenshot 2024-10-23 at 13 39 51


justvanrossum (commented):
I would love you to:
- remove the menu bar items
- move the action registrations and implementation to panel-reference-font.js

The reference font panel should be considered a plugin, and the main point is to allow users to assign shortcuts. So all it needs to do is to register the two actions.

Plugins currently cannot provide localized strings themselves, so for now these unfortunately have to live with the global Fontra strings.

So, the changes should only be in:
- the language files
- panel-reference-font.js

This PR should leave editor.js alone.

ollimeier:
> I would love you to:
> 
> * remove the menu bar items
> * move the action registrations and implementation to panel-reference-font.js
> 
> The reference font panel should be considered a plugin, and the main point is to allow users to assign shortcuts. So all it needs to do is to register the two actions.
> 
> Plugins currently cannot provide localized strings themselves, so for now these unfortunately have to live with the global Fontra strings.
> 
> So, the changes should only be in:
> 
> * the language files
> * panel-reference-font.js
> 
> This PR should leave editor.js alone.

Thanks for the quick feedback. I moved the code.

ollimeier:
@justvanrossum Do you have a preference how to sort the actions? Currently it's this:
`const topic = "0020-action-topics.menu.view";`
But I am wondering if it should get its own topic?

justvanrossum:
> But I am wondering if it should get its own topic?

Yes it should, called "Reference font". Maybe we should use a much higher number here, for plugin-like items such as these. Say `0200-action-topics.reference-font`, so for now this will stick to the end, even if we add more topics.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -322,6 +323,51 @@ export default class ReferenceFontPanel extends Panel {
...
+        titleKey: "menubar.view.select-next-reference-font",
+        defaultShortCuts: [],
+      },
+      () => this.doSelectPreviousNextReferenceFont(false)
+    );
+  }
+
+  async doSelectPreviousNextReferenceFont(selectPrevious) {
+    const listLength = this.filesUIList.items.length;
+    if (!listLength) {
```
justvanrossum, line 355:
```suggestion
    if (listLength < 2) {
```

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -322,6 +323,51 @@ export default class ReferenceFontPanel extends Panel {
...
+      return;
+    }
+
+    const index = this.filesUIList.getSelectedItemIndex();
```
justvanrossum, line 359:
```suggestion
    const index = this.filesUIList.getSelectedItemIndex() || 0;
```
Then we don't have to bother with isNaN below.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -322,6 +323,51 @@ export default class ReferenceFontPanel extends Panel {
...
+
+    const newIndex = isNaN(index)
```
justvanrossum, line 361:
```js
  const newIndex = modulo(index + (selectPrevious ? -1 : 1), listLength);
```
`modulo()` from utils.js.

justvanrossum: `doSelectPreviousNextSource()` uses JS's `%` modulo, and adds numSources to make it work for negative numbers. Our `modulo()` function behaves like Python's `%` and does the right thing. `doSelectPreviousNextSource()` should probably follow the same pattern.

```js
$ node
> 14 % 6
2
> -14 % 6
-2
> 
```

```python
$ python
>>> 14 % 6
2
>>> -14 % 6
4
>>> 
```

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -322,6 +323,43 @@ export default class ReferenceFontPanel extends Panel {
...
+    this.initActions();
+  }
+
+  initActions() {
+    const topic = "0200-action-topics.reference-font";
+    registerAction(
+      "action.select-previous-reference-font",
+      {
+        topic,
+        titleKey: "menubar.view.select-previous-reference-font",
```
justvanrossum, line 336: These keys are no longer correct.

How about
- "reference-font.select-previous-reference-font"
- "reference-font.select-next-reference-font"

?

src/fontra/views/editor/panel-reference-font.js
ollimeier: Oh, that's impressive. Thanks. I add it as suggested. And for the doSelectPreviousNextSource I created a new issue. I am closing this as resolved.

src/fontra/views/editor/panel-reference-font.js
ollimeier: Thanks. I update the titleKeys in the .js file, as well as in the google sheet and I also updated the languages files after that.

PR #1744, opened 2024-10-23
Use modulo for doSelectPreviousNextSource

### LINKED ISSUE #1743: Use modulo for `doSelectPreviousNextSource()`
`doSelectPreviousNextSource()` uses JS's `%` modulo, and adds numSources to make it work for negative numbers. Our `modulo()` function behaves like Python's `%` and does the right thing. `doSelectPreviousNextSource()` should probably follow the same pattern.

```js
$ node
> 14 % 6
2
> -14 % 6
-2
> 
```

```python
$ python
>>> 14 % 6
2
>>> -14 % 6
4
>>> 
```

_Originally posted by @justvanrossum in https://github.com/googlefonts/fontra/pull/1742#discussion_r1812707524_

ollimeier:
Fixes #1743


justvanrossum:
You should learn the commands to really undo a commit in a branch (which can work with a forced push), so we don't have these back-and-forth-commits in the file history. Because you'll now have to add the modulo import a second tome.

Here are some tips: https://www.git-tower.com/learn/git/faq/undo-last-commit
But there's a lot more info out there, and it may be difference for GH Desktop and/or VS Code.

Issue #1752, opened 2024-10-25
Some UI strings aren't handled through translate(), so are currently not translatable


justvanrossum:
It seems there are still some items not in the spreadsheet. For example:
- "Characters that decompose with this character" in Related Glyphs panel
- "Glyph note for" in the Glyph Notes
- The button in the shortcuts setting page
- "cut", "copy", "paste", which do exist in the sheet, but the translations are not deployed in the interface

(There may be more)

justvanrossum:
Fixed by #1764.

PR #1764, opened 2024-11-04
Add missing translation keys

### LINKED ISSUE #1763: application-settings.display-language.title maybe always english?
When you accidentally changed the language to a language you cannot read and closed the tab, it's difficult to find back to 'Languages' when you don't know exactly where it is located. 
So, I was wondering if  application-settings.display-language.title should alway be english or if it should include at least english, but then the name get's very long, eg: "Weergavetaal (Language)"

ollimeier:
Fixes #1763 #1752

This is the first part of the rework for missing translations.

I split it up, because it may get messy otherwise. This PR includes commits based on each file (Related Glyphs, Glyph Note, Shortcuts panel, Sources Panel and Status Definitions panel). But the next work might not be structured this nicely anymore, because the strings are all over the place. 
So, I created another branch where I first marked the missing translations:
https://github.com/googlefonts/fontra/tree/issue-1752-missing-translation-keys-2


ollimeier:
@justvanrossum I am wondering if we really want to have a unique translation for undo (always)? 

Here an example: 
We have `undoLabel: "align left",` this would require a new translation key, but we have already a translation for '**Align left**' via key: `sidebar.selection-transformation.align.left`.
This is same for many other undo strings. It's basically just set in 'lowercase'.
Do you think it would make sense to reuse some translation keys and accept it's not lowercase? We could also make it always convert to lowercase, but I am not sure if this would work (always) for other languages except English?
But doubling the keys also does not feel correct.

Would be great to get you feedback and advice. Thanks a lot in advance.

src/fontra/views/fontinfo/panel-development-status-definitions.js
```diff
@@ -3,34 +3,35 @@ import * as html from "../core/html-utils.js";
...
+import { translate } from "/core/localization.js";
 import { message } from "/web-components/modal-dialog.js";
 
 const defaultStatusFieldDefinitions = {
   "fontra.sourceStatusFieldDefinitions": [
     {
       color: [1, 0, 0, 1],
       isDefault: true,
-      label: "In progress",
+      labelKey: "development-status-definitions.default.label.in-progress",
```
justvanrossum, line 14: I don't think these should be translated: the status labels are stored in the font's customData, and I wouldn't want to add "labelKey" to the data structure. Also, when adding new custom statuses, we don't have the option to provide translations, so it'll be a bit half-assed anyway.

ollimeier: Thanks. I reverted the changes relate to that.

ollimeier:
As discussed yesterday, this translation key update is almost complete, but we decided to leave it as it is, so that we do not spend too much time on translations and can focus on other features.

src/fontra/client/web-components/plugin-manager.js
```diff
@@ -50,18 +51,30 @@ export class PluginManager extends SimpleElement {
   }
 
   async promptAddPlugin(text = "") {
-    const newPluginPrompt = await dialogSetup("Add plugin", "", [
-      { title: "Cancel", resultValue: "no", isCancelButton: true },
-      { title: "Create", resultValue: "ok", isDefaultButton: true, disabled: true },
-    ]);
+    const newPluginPrompt = await dialogSetup(
+      translate("plugin-manager.dialog.title"),
```
justvanrossum, line 55: Suggest to use a less generic key prefix for these. What about `plugin-manager.add-plugin-dialog.xxx`?

src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -419,7 +420,12 @@ export class PointerTool extends BaseTool {
         }
       }
       return {
-        undoLabel: "drag selection" + (shouldConnect ? " and connect contours" : ""),
+        undoLabel: translate(
```
justvanrossum, line 423: This looks tricky, esp. with the space added to the argument. Wouldn't two distinct keys be easier?

src/fontra/views/editor/editor.js
```diff
@@ -1005,26 +1009,38 @@ export class EditorController {
   async showDialogGlyphEditCannotEditReadOnly(create = false) {
     const glyphName = this.sceneSettings.selectedGlyphName;
     await message(
-      `Can’t ${create ? "create" : "edit"} glyph “${glyphName}”`,
-      "The font is read-only."
+      create
```
justvanrossum, line 1012: Suggest to do the switching inside the `translate()` call, something like this:
```js
   translate(create ? key1 : key2, glyphName)
```

src/fontra/views/editor/editor.js
```diff
@@ -2272,12 +2288,15 @@ export class EditorController {
...
       const newAnchorIndex = instance.anchors.length - 1;
       this.sceneController.selection = new Set([`anchor/${newAnchorIndex}`]);
-      return "Add Anchor";
+      return translate("action.add-anchor");
     });
   }
 
   async doAddEditAnchorDialog(anchor = undefined, point = undefined) {
-    const titlePrefix = anchor ? "Edit" : "Add";
+    const titleDialog = anchor
```
justvanrossum, line 2296: Similarly, suggest to put the switch inside the `translate()` call, so we don't have to repeat it.

src/fontra/views/editor/editor.js
```diff
@@ -2272,12 +2288,15 @@ export class EditorController {
...
+      ? translate("action.edit-anchor")
+      : translate("action.add-anchor");
+    const defaultButton = anchor ? translate("dialog.edit") : translate("dialog.add");
```
justvanrossum, line 2299: Same: please use only one `translate()` call.

src/fontra/views/editor/editor.js
```diff
@@ -2436,8 +2456,9 @@ export class EditorController {
     const numGuidelines = guidelineSelection?.length || 0;
     // + (fontGuidelineSelection?.length || 0);
 
-    const s = numGuidelines > 1 ? "s" : "";
-    return `${hasLockedGuidelines ? "Unlock" : "Lock"} Guideline${s}`;
+    return hasLockedGuidelines
```
justvanrossum, line 2459: Same, pls put the switch inside the `translatePlural()` call.

src/fontra/views/editor/editor.js
```diff
@@ -2473,7 +2493,9 @@ export class EditorController {
             guideline.locked = locking;
           }
         }
-        return `${identifier} Guideline(s)`;
+        return locking
```
justvanrossum, line 2496: Same, pls put the switch inside the translatePlural() call.

src/fontra/views/editor/editor.js
```diff
@@ -2531,7 +2553,12 @@ export class EditorController {
     point = undefined,
     global = false
   ) {
-    const titlePrefix = guideline ? "Edit" : "Add";
+    const titleDialog = guideline
```
justvanrossum, line 2556: Same, pls put the switch inside the translatePlural() call.

src/fontra/views/editor/editor.js
```diff
@@ -2531,7 +2553,12 @@ export class EditorController {
...
+      ? translate("action.edit-guideline")
+      : translate("action.add-guideline");
+    const defaultButton = guideline
```
justvanrossum, line 2559: Same, pls put the switch inside the translatePlural() call.

src/fontra/views/editor/panel-designspace-navigation.js
```diff
@@ -671,7 +681,12 @@ export default class DesignspaceNavigationPanel extends Panel {
       sourceController.addKeyListener("active", async (event) => {
         await this.sceneController.editGlyphAndRecordChanges((glyph) => {
           glyph.sources[index].inactive = !event.newValue;
-          return `${event.newValue ? "" : "de"}activate ${source.name}`;
+          return event.newValue
```
justvanrossum, line 684: Same, pls put the switch inside the translate() call.

src/fontra/views/editor/panel-designspace-navigation.js
```diff
@@ -796,9 +821,12 @@ export default class DesignspaceNavigationPanel extends Panel {
       let layerMessage = "";
       if (layer !== undefined && deleteLayerCheckBox.checked) {
         delete glyph.layers[source.layerName];
-        layerMessage = " and layer";
+        layerMessage = ` ${translate("sidebar.designspace-navigation.undo.and-layer")}`;
```
justvanrossum, line 824: I think the leading space should be part of the string in the spreadsheet, like we do elsewhere. It depends on the language whether it should be there or not.

src/fontra/views/editor/panel-glyph-note.js
```diff
@@ -140,9 +143,9 @@ async function saveGlyphNote(glyphName, sceneController, newNote) {
     glyph.customData["fontra.glyph.note"] = newNote;
     return oldNote
       ? newNote
-        ? "edit glyph note"
-        : "delete glyph note"
-      : "add glyph note";
+        ? translate("sidebar.glyph-note.undo.edit")
```
justvanrossum, line 146: Please rework to use a single `translate()` call.

src/fontra/views/editor/panel-related-glyphs.js
```diff
@@ -147,8 +154,10 @@ export default class RelatedGlyphPanel extends Panel {
             this.relatedGlyphsHeaderElement.appendChild(
               html.div({ class: "no-related-glyphs" }, [
                 glyphName
-                  ? "(No related glyphs or characters were found)"
-                  : "(No glyph selected)",
+                  ? translate(
```
justvanrossum, line 157: Same, pls put the switch inside the translate() call.

src/fontra/views/editor/panel-selection-info.js
```diff
@@ -177,10 +177,11 @@ export default class SelectionInfoPanel extends Panel {
               : "/tabler-icons/lock-open-2.svg",
           "onclick": (event) => this._toggleGlyphLock(varGlyphController.glyph),
           "data-tooltip": this.fontController.readOnly
-            ? "Glyph is read-only"
+            ? translate("sidebar.selection-info.glyph-locking.tooltip.read-only")
```
justvanrossum, line 180: Same, pls put the switch(es) inside the translate() call.

src/fontra/views/editor/panel-selection-info.js
```diff
@@ -484,7 +491,7 @@ export default class SelectionInfoPanel extends Panel {
       for (const [layerName, layerGlyph] of Object.entries(editLayerGlyphs)) {
         layerGlyph.components[componentIndex].transformation = getDecomposedIdentity();
       }
-      return "reset transformation";
+      return translate("sidebar.selection-info.component.transformation.tooltip");
```
justvanrossum, line 494: I think the key should have the word "reset" in it. And I'm not sure it needs the "tooltip" part.

src/fontra/views/editor/panel-selection-info.js
```diff
@@ -515,7 +522,7 @@ export default class SelectionInfoPanel extends Panel {
           }
         }
       }
-      return "reset axis values";
+      return translate("sidebar.selection-info.component.reset-axis-values.tooltip");
```
justvanrossum, line 525: Here you do include "reset". But likewise, I don't think "tooltip" is needed.

src/fontra/views/fontinfo/panel-development-status-definitions.js
```diff
@@ -186,24 +190,32 @@ class StatusDefinitionBox extends HTMLElement {
...
+        errorMessage
+      );
       return false;
     }
     return true;
   }
 
-  replaceStatusDef(newStatusDef, undoLabel, statusIndex = this.statusIndex) {
+  replaceStatusDef(newStatusDef) {
+    const undoLabel = translate("development-status-definitions.undo.change-color");
```
justvanrossum, line 215: This function is used to change the color *and* the label, so now the label is wrong for the label case.

However, it's quite irresponsible you spent time on this to begin with, as these undo labels aren't even visible anywhere: we only have command-Z and command-shift-Z. No menu bar, no context menu.

Now that this code is there, I think it's fine if both calls to `replaceStatusDef()` share their undo message, it just needs to be generic, and not mention "color" or "label", also not in the key.

src/fontra/client/web-components/plugin-manager.js
ollimeier: thanks. done.

src/fontra/views/editor/edit-tools-pointer.js
ollimeier: Yes, I am totally with you. I tried to be as close as possible to the origin version. I changed it. done.

src/fontra/views/editor/editor.js

src/fontra/views/editor/panel-selection-info.js
ollimeier: done.

src/fontra/views/editor/panel-selection-info.js
ollimeier: done.

src/fontra/views/editor/panel-selection-info.js
ollimeier: done.

src/fontra/views/fontinfo/panel-development-status-definitions.js
ollimeier: I'm sorry. I changed it and updated it with a more generic key and translation.

src/fontra/views/editor/panel-designspace-navigation.js
ollimeier: thanks. done.

src/fontra/views/editor/panel-designspace-navigation.js
ollimeier: Thanks for the info. I changed it.

src/fontra/views/editor/panel-glyph-note.js
ollimeier: done.

src/fontra/views/editor/editor.js
ollimeier: I missed a couple notes from you, because they were hidden. Now, this is done as well. Sorry.

PR #1765, opened 2024-11-04
Add missing translation keys (part 2)

ollimeier:
I make this as a draft PR (not to `main`, but to `issue-1752-missing-translation-keys`), so you see what I am doing.


justvanrossum:
So to recap, please don't translate exception messages in library code like `throw new VariationError("error message")`. Else this PR should probably be merged into the "part 1" branch.

ollimeier:
> So to recap, please don't translate exception messages in library code like `throw new VariationError("error message")`. Else this PR should probably be merged into the "part 1" branch.

I removed the translation keys for VariationError and merged this PR into the 'Part 1'.

PR #1766, opened 2024-11-06
Don't allow guideline selection, if they are hidden.

### LINKED ISSUE #1760: Hidden guidelines can be hovered and selected
I noticed, that guidelines which are not displayed because hidden via 'Glyph editor appearance' can still be hovered, selected and also be changed once selected.

https://github.com/user-attachments/assets/91633a93-de0b-4bd5-8dab-51463a8b88d8

ollimeier:
Fixes #1760 

Hi @justvanrossum , this is a possible solution, but I am not sure if this is the right path.
My thinking here is, that when the visualizationLayersSettings for guidelines is set to be 'hidden', the guidelines should not be part of the general selection. But to check the status of the guidelines visualizationLayersSettings we need `editorController`, but this hasn't been part of `SceneModel`, yet. And I think it may has a reason??? 
Therefore this is just a proposal to get in a conversation with you to get some advice.

### DIFF
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -134,7 +134,8 @@ export class EditorController {
     this.sceneController = new SceneController(
       this.fontController,
       canvasController,
-      applicationSettingsController
+      applicationSettingsController,
+      this
     );
 
     this.sceneSettingsController = this.sceneController.sceneSettingsController;
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -131,11 +131,24 @@ export class EditorController {
       async (...args) => await this.editListenerCallback(...args)
     );
 
+    this.visualizationLayers = new VisualizationLayers(
+      visualizationLayerDefinitions,
+      this.isThemeDark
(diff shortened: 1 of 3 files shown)


justvanrossum:
Giving SceneController a reference to EditorController is definitely not the right solution, as it breaks the abstraction. Will look at it later.

ollimeier:
> > Giving SceneController a reference to EditorController is definitely not the right solution, as it breaks the abstraction. Will look at it later.
> 
> Since you only need `editorController.visualizationLayersSettings`, please pass _that_ instead of the entire `editorController` object. This leaves the abstraction graph intact.
> 
> (Before you say, "we could do one step further, and pass `editorController.visualizationLayersSettings.model`": yes, but not doing that will make it trivial for the `SceneModel` to subscribe to changes in the visualization layer settings, which can be useful in the future. So no.)

Thank you very much for your feedback. I changed it as requested. I also updated the `doSelectAllNone()`, so that you cannot select guidelines via shortcuts, when they are hidden. The PR is now 'Ready for review'.

justvanrossum (approved):

PR #1767, opened 2024-11-07
Fix #1759: Hovering over an icon can trigger the scroll bar

### LINKED ISSUE #1759: Hovering over an icon can trigger the scroll bar
I think this is not intended
Image

hwoongkang:
As @justvanrossum said: https://github.com/googlefonts/fontra/issues/1759#issuecomment-2460640093, animating the svg inside the icon-button is indeed a good solution.

I changed the animating DOM object from the button itself to the svg inside the button.

I also added `will-change: transform` to the svg, in order to not to repeat the issue #1388.

Changing the animating object only didn't work in Windows. I added `contain: content` to the button, to indicate that the transform of its content should not affect the layout: https://developer.mozilla.org/en-US/docs/Web/CSS/contain

This fixes #1759

### DIFF
--- a/src/fontra/client/web-components/icon-button.js
+++ b/src/fontra/client/web-components/icon-button.js
@@ -12,21 +12,28 @@ export class IconButton extends UnlitElement {
       color: var(--foreground-color);
       width: 100%;
       height: 100%;
-      transition: 150ms;
       cursor: pointer;
-      will-change: transform;
+      contain: content;
     }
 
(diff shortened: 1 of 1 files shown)


justvanrossum:
Thank you so much! Will review and test soon.

justvanrossum (approved):
I tested on Windows: I managed to reproduce the original report and can confirm that your fix is working. Thank you so much!

PR #1783, opened 2024-11-13
[background image] Front-End updates (selectable, editable and transformable)

### LINKED ISSUE #1782: [background image] Make the background image selectable, editable, etc.
Make the background image:
- selectable (+ "selected" visualization, I imagine a stroke around the image)
- editable (movable, draggable, via EditBehavior?)
- transformable ("bounds handles" resize/rotate)
- ~deleteable~

_note from Just: I guess some of the code from https://github.com/googlefonts/fontra/pull/1707 can be used._

ollimeier:
This fixes #1782


ollimeier:
@justvanrossum `backgroundImageSelectionAtPoint` currently returns a set, to make it equal to the other selection functions. 
But for now, only one background image can be selected, so it could be a boolean, right? 
But this would be inconsistent with the other selection functions. What do you think?
Do we limit the background image to one image only or do we eventually allow multiple images in future?

justvanrossum:
Please keep it a set, both for consistency with the other methods, as well as for possible future expansion to multiple images.

src/fontra/views/editor/scene-model.js
```diff
@@ -706,6 +713,61 @@ export class SceneModel {
...
+    const image = this.fontController.getBackgroundImageCached(
+      backgroundImage.identifier,
+      () => controller.requestUpdate()
+    );
+
+    if (!image) {
+      return new Set();
+    }
+
+    const x = point.x - positionedGlyph.x;
```
justvanrossum, line 745: Olli, please do read the feedback I wrote on the old PR. Why do you just paste the old code instead of doing it right based on my feedback? You're again reinventing transformation math badly instead of doing the right thing and using a Transform object. Convert the DecomposedTransform to a regular Transform first.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -430,6 +435,10 @@ registerVisualizationLayerDefinition({
       return;
     }
 
+    const { backgroundImage: backgroundImageSelection } = parseSelection(
```
justvanrossum, line 438: You don't need to parse the selection. Just check whether "image/0" is in the selection set.

src/fontra/views/editor/visualization-layer-definitions.js
```diff
@@ -449,6 +458,13 @@ registerVisualizationLayerDefinition({
         context.globalAlpha = backgroundImage.color.alpha;
       }
       context.drawImage(image, 0, 0, image.width, image.height);
+
+      if (backgroundImageSelection) {
```
justvanrossum, line 462: This approach will transform the stroke as well. So if the image is small, the stroke will be thinner. Etc.

So, take this out of the `withSavedState()` block, set up the four corner points, and transform those, using `affine.transformPointObject()`, then use the resulting points to draw a polygon. We may need a helper function `drawPolygon(context, poinsts)`.

Additionally I suggest to use `context.lineJoin = "round";`

src/fontra/views/editor/scene-model.js
justvanrossum: Note that the image should also participate in rect selection, so you will need to write `backgroundImageSelectionAtRect()` as well. The two methods will have the transformed bounding box logic in common.

(Come to think of it, did we consciously decide to let anchors and guidelines _not_ participate in rect select? We may want to revisit that decision.)

justvanrossum:
Will this PR implement all of #1782, or just part of it?

ollimeier:
> Will this PR implement all of #1782, or just part of it?

What do you prefer? Both is still possible: as separate PRs or merge them into one and then have all of #1782 in one PR.

justvanrossum:
Please use a better title than "Make the background image".

justvanrossum:
And use `[background image]` as a prefix for all PRs and issues related to the bg image tasks.

justvanrossum:
So, when I say "separate PRs", I mean "sequential PRs", not "nested PRs", ie. please please please no PRs on PRs, that just brings utter chaos.

src/fontra/views/editor/scene-model.js
ollimeier: About: _(Come to think of it, did we consciously decide to let anchors and guidelines not participate in rect select? We may want to revisit that decision.)_
-> I checked RoboFont and Glyphs and both do not include anchors or guidelines via 'rect-select'.

ollimeier:
> Please use a better title than "Make the background image".

I changed the title to: **[background image] Front-End updates (selectable, editable and transformable)**

src/fontra/views/editor/scene-model.js
justvanrossum: In this case I honestly don't care what RF or Glyphs do. So yes, I want to revisit that decision for anchors and guidelines. But first, yes, I do want `backgroundImageSelectionAtRect()`.

ollimeier: OK, done.
Most of the code for `backgroundImageSelectionAtPoint` and `backgroundImageSelectionAtRect` would be identical, why I ended up to create `backgroundImageSelectionAtPointRect` instead. I hope this is also in your interest. 
I also created this issue: https://github.com/googlefonts/fontra/issues/1785 as a possible follow up for including  anchors and guidelines into rect-select. So we don't forget about it. 
I am closing this as feedback as resolved.

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: Good point. I changed it to `if (model.selection.has("backgroundImage/0")) {` 
BTW: I named it `backgroundImage/0` and not `image/0` as previously, because I saw you named it `backgroundImage` everywhere. I hope this is in your interest.

src/fontra/views/editor/visualization-layer-definitions.js
ollimeier: I changed it as suggested, but instead of `drawPolygon` I named it `strokePolygon` to be in sync with other names like: `strokeCircle`, `strokeLineDashed`, `strokeLine`. I very much hope, this is in your interest as well.

src/fontra/client/core/glyph-controller.js
```diff
@@ -651,6 +655,33 @@ export class StaticGlyphController {
       }
     }
 
+    if (backgroundImageIndices.length) {
```
justvanrossum, line 658: While you are doing the overly generic `backgroundImageIndices`, might as well go all the way, and write it like this:
```js
    for (const i of backgroundImageIndices) {
      assert(i == 0, "we currently only support a single bg image");
      // Then, and check that now makes you go 
      //    return unionRect(...);  // lots of repetition
      // you can do "break;" instead
```
    }

src/fontra/client/core/glyph-controller.js
```diff
@@ -651,6 +655,33 @@ export class StaticGlyphController {
...
+      const image = fontController.getBackgroundImageCached(backgroundImage.identifier);
+      if (!image) {
+        return unionRect(...selectionRects);
+      }
+
+      const affine = decomposedToTransform(backgroundImage.transformation)
+        .translate(0, image.height)
+        .scale(1, -1);
+
+      const pt1 = affine.transformPointObject({ x: 0, y: 0 });
```
justvanrossum, line 675: In rectangle.js, please add a function `rectToPoints()` that returns an array of the four corner points. From the image width and height you can most easily construct a rect with the `rectFromArray([xMin, yMin, xMax, yMax])` function.

src/fontra/views/editor/edit-behavior.js
```diff
@@ -39,13 +40,20 @@ export class EditBehaviorFactory {
...
     this.componentOriginIndices = componentOriginIndices || [];
     this.componentTCenterIndices = componentTCenterSelection || [];
     this.behaviors = {};
     this.enableScalingEdit = enableScalingEdit;
   }
 
-  getBehavior(behaviorName, fullComponentTransform = false) {
+  getBehavior(
+    behaviorName,
+    fullComponentTransform = false,
```
justvanrossum, line 54: Are you sure we need separate flags for fullComponentTransform and fullBackgroundImageTransform? They are the same thing (either both are true, or both are false), so maybe rename `fullComponentTransform` to `doFullTransform`, and you're covered for both cases.

src/fontra/views/editor/scene-model.js
```diff
@@ -716,6 +724,58 @@ export class SceneModel {
   //fontGuidelineSelectionAtPoint(point, size) {
   // }
 
+  backgroundImageSelectionAtPointRect(point = undefined, selRect = undefined) {
```
justvanrossum, line 727: Please make two separate methods, `backgroundImageSelectionAtPoint()` and `backgroundImageSelectionAtRect()`, that then call _this_ function, which should be renamed with an underscore prefix.

src/fontra/client/core/glyph-controller.js
```diff
@@ -610,7 +610,9 @@ export class StaticGlyphController {
     return getRepresentation(this, "flattenedPathHitTester");
   }
 
-  getSelectionBounds(selection) {
+  // TODO: Question: It does not feel right to have fontController as an argument here.
+  // But it is needed to get the backgroundImage. Is there a better way?
+  getSelectionBounds(selection, fontController = undefined) {
```
justvanrossum, line 615: Instead of passing the font controller, please pass a function `getBackgroundImageBounds(imageIdentifier)`.

`getBackgroundImageBounds(imageIdentifier)` should become a new method on FontController. We then pass this as follows:
```js
    glyph.getSelectionBounds(selection, fontController.getBackgroundImageBounds.bind(getBackgroundImageBounds));
```

Although I can imagine we can make that last line a little more convenient. On FontController:
```js
    get getBackgroundImageBoundsFunc () {
      return this.getBackgroundImageBounds.bind(this);
```

src/fontra/views/editor/edit-behavior.js
```diff
@@ -299,6 +346,57 @@ function makeComponentOriginEditFunc(component, componentIndex, roundFunc) {
...
+      return makeBackgroundImageChange(newBackgroundImage);
+    },
+    makeBackgroundImageChange(oldBackgroundImage),
+  ];
+}
+
+// The following causes an error:
+// Uncaught (in promise) RemoteError: TypeError("unhashable type: 'dict'")
+// function makeBackgroundImageChange(image) {
+//   return { f: "=", a: [image] };
```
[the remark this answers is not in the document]
ollimeier: Thanks. done.

src/fontra/views/editor/edit-behavior.js
ollimeier: agree. done.

src/fontra/views/editor/scene-model.js
ollimeier: ok. done.

src/fontra/client/core/glyph-controller.js
ollimeier: ok. I changed it, but named `i` to `imageIndex` to follow the same naming scheme similar to the code above: `anchorIndex` or `componentIndex`.

src/fontra/client/lang/ja.js
```diff
@@ -220,7 +220,9 @@ export const strings = {
   "menubar.view.find-glyphs-that-use": "Find glyphs that use '%0'",
   "menubar.view.remove-selected-glyph-from-canvas": "Remove selected glyph from canvas",
   "menubar.view.replace-selected-glyph-on-canvas": "Replace selected glyph on canvas",
+  "menubar.view.select-next-glyph": "Select next glyph",
```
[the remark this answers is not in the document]
ollimeier: So, the reason is, that these keys are actually really missing in main branch for Japanese. 
'Merge branch 'main' into issue-1752-missing-translation-keys' does not include Japanese, but the keys for previous and next glyph are part of it. Then one commit later 'Merge pull request [#1764] from googlefonts/issue-1752-missing-translation-keys' includes Japanese, but without keys for previous/next glyph.
I suspect, this is due to 'WIP'. During my work on previous/next glyphs the keys work marked as WIP in the google sheet. It was wart of the PR, because I updated the languages files with `--include-wip`. I removed 'WIP' from the google sheet some time (maybe one day) after we merged it the PR.

src/fontra/client/core/glyph-controller.js
ollimeier: done. I also added unittests for rectToPoints.

src/fontra/client/core/glyph-controller.js
ollimeier: done.

src/fontra/client/lang/ja.js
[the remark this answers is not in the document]
ollimeier: I am closing this, as this is not an issue related to this PR.

PR #1784, opened 2024-11-13
Make the background image editable

ollimeier:
movable, draggable, via EditBehavior

This is another draft PR on top of the other, because it's easier to see the changes related to **Make the background image editable**.


justvanrossum:
Please don't do PRs on PRs. This is actually more confusing than helpful in most cases.

This PR now has conflicts and is not reviewable.

PR #1789, opened 2024-11-15
[background image] Add dialog for background image

### LINKED ISSUE #1787: add a dialog for image properties (mostly color, or maybe just opacity/alpha)

ollimeier:
Fixes #1787

Work in progress -> therefore draft PR.

I had to use the other branch as the base branch (not main), to have the selection code included.


ollimeier:
@justvanrossum I found a solution to make the color apply on a png. This is based on the idea to make every white pixel transparent. Reference: https://stackoverflow.com/questions/30438524/algorithm-to-make-image-white-and-transparent

justvanrossum:
It doesn't work for me: I see the whole image as a solid rectangle in the selected color.

justvanrossum:
This should be rebased on main somehow.

ollimeier:
> It doesn't work for me: I see the whole image as a solid rectangle in the selected color.

Does it work now? I made some changes. Search for `function white2transparent(img)`. It's definitely not the right solution, but a demo of the idea 'convert white pixel to transparent pixel'. 
I ask myself in general whether this is a good idea and, if so, whether it should be done in the backend or if in the frontend?
(I do not work today, but it bothered me and I am currently at the airport, where I had some time to look into it)

ollimeier:
@justvanrossum I am also wondering if we should wast our time with the color image issue. Would a user really use that kind of feature? I think the opacity feature is the more important one. 
Having said that, I am wondering if setting the opacity for each image individually is the right thing to do or if we prefer to include such settings more globally in our applications settings? So, that all images have the same opacity, or at least a global default opacity and if some wants to adjust a specific one, they can do this as well?

justvanrossum:
It does work now. But I notice that this colorization is very binary: any grayscale values are gone. So that's definitely not desired.

Let's for now not do colorization, and only provide the opacity slider.

About having a global setting: maybe. Let's start with the individual opacity slider.

(There may be better options with the [Canvas filter property](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/filter), but this isn't supported in Safari.)

justvanrossum:
I have found the solution that doesn't involve pixel hacking. To properly support opacity, we unfortunately need to do the colorization into a temporary canvas. I got the colorization working efficiently (using caching) here: https://github.com/googlefonts/fontra/pull/1815. Will work on the color picker UI in the selection info panel.

ollimeier:
> I have found the solution that doesn't involve pixel hacking. To properly support opacity, we unfortunately need to do the colorization into a temporary canvas. I got the colorization working efficiently (using caching) here: #1815. Will work on the color picker UI in the selection info panel.

This is awesome. Thank you very much.

PR #1841, opened 2024-12-05
Clean up unused background images

### LINKED ISSUE #1813: [background image] Clean up unused background images from .ufo and .fontra

ollimeier:
Fixes #1813 

This is a draft PR, because I know this is not the right approach and this is not a priority, this is just to show you what I did. We can delete it later.

### DIFF
--- a/src/fontra/backends/designspace.py
+++ b/src/fontra/backends/designspace.py
@@ -1200,6 +1200,24 @@ class DesignspaceBackend:
             reader = self.ufoManager.getReader(ufoPath)
             reader.writeImage(imageFileName, data.data, validate=True)
 
+    async def deleteBackgroundImage(self, imageIdentifier: str) -> None:
+        imageInfo = self._imageMapping.reverse.get(imageIdentifier)
+        if imageInfo is None:
+            return
+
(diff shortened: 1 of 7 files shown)


ollimeier:
@justvanrossum Here a short demo, which shows that the data gets removed and restored:


https://github.com/user-attachments/assets/e3c7f369-7fae-43a4-b780-549a43f87307

justvanrossum:
The approach I'm not most seriously considering is pretty much: let the backend do garbage collecting.

The _backend_ keeps track of which image files are actually used, and will put aside the unused images in a temp folder, so they can be "revived" in case undo/redo causes them to be referenced again.

Unreferenced images could be deleted permanently:
- after a certain time they were last referenced
- once the backend closes

Practically, the GlyphDependencies class can be used to track the glyph <-> images relationships. And the initial dependencies could be read in the background once the backend starts up, just like how component references are read.

Once we pick this up, I suggest to _first_ implement it for .fontra, and only tackle .ufo after that, as it's more complex. The .fontra implementation may be simpler, as it is not possible (by design) to reuse a single image in multiple layers/glyphs.

justvanrossum:
This PR is not going in the right direction. Once we pick this up, it'll be better to start from scratch.

Issue #1845, opened 2024-12-10
[font overview] Menu bar for font overview


ollimeier:
> - Some of the top menu bar's *contents* may have to be duplicated, but I suspect that only a fraction from the editor makes sense in a glyph overview view. It is fine to see the menu bar part of this as a separate task. 

 _Originally posted by @justvanrossum in [#102](https://github.com/googlefonts/fontra/issues/102#issuecomment-2526314723)_

PR #1849, opened 2024-12-11
Font overview phase 1

### LINKED ISSUE #102: Font Overview view(s)
Font overview:
- grid view
- list view / spreadsheet

Also affects the project list: it needs a way to either jump to a font overview or to a glyph editor.

ollimeier:
Part of #102 

This implements a very basic font overview. It needs some followups for refinement, but the core functionality is there and mergable.


ollimeier:
@justvanrossum 

TODOs and Questions:
1. I am wondering if it would make sense to refactor `GlyphsSearch` into two web components:
   1. `GlyphsSearch`: Includes the search field, only. Access the list of glyphs with eg. glyphsListItemsController.
   2. `GlyphsSearchForEditor` which uses GlyphsSearch and adds the glyph list.
2. We may want to move `sortedSourceIdentifiers` to a more general location like utils. Follow up task?
3. Do we want to make the sidebar scalable? If so, we may want to refactor `sidebar-resize-gutter` or at least have a look at it. Follow up task?
4. Context menu is implemented in the overview, yet. We may want to add them. Is this a follow up task?
5. Maybe use https://www.npmjs.com/package/unicode-properties for overview sections. Also, how to we handle unencoded glyphs? Follow up task?
6. Add top menu bar, please see: https://github.com/googlefonts/fontra/issues/1845
7. When opening a glyph in the editor via double click, there is an error: It says `error while interpolating font sources '{message: 'objects have incompatible number of entries: 7 != 6', type: 'interpolation-error'}'`. But this is not true. I don't know why this is happening.

ollimeier:
Little demo of the current stage:

https://github.com/user-attachments/assets/001ec9b2-21fc-40d4-8840-b361786a58bc

justvanrossum:
Next steps:
- Figure out "cell view" API, 
  - `getSelectedGlyphNames()`
  - `setSelectedGlyphNames()`
  - `addEventListener("selectionChanged", (event) => ...)`
- Selection behavior:
  - Click to select glyph
  - Shift click to deselect glyph, or add glyph to selection
  - Click-drag, to "paint" selection? Can be followup.
  - Arrow key navigation. Can be followup.
  - Typing a character (and glyphname?) selects the matching glyph. Can be followup.
- Fix cell layout so cells "flow" ("inline block"), see "Related Glyphs" panel
- Fix cell glyph load/draw so it only loads once the cell becomes visible (currently it seems all glyphs are loaded, also the ones outside the scroll area)

Question:
- Should the cell view be an accordion view, so sections can be folded/hidden?

justvanrossum:
Minor bug: double clicking the `/` glyph results in an empty canvas. Since it is the prefix for glyph names, if you want to type a slash, it should be doubled. So I think we should special-case the `/` to become `//` in the text entry.

justvanrossum:
I made a small change to glyph-cell.js, that fixes the lazy loading: d5104697ed1f9dc045fdc4d978a71edcdce51c4b. You should notice a vast improvement in performance, as only glyphs that are in view will be loaded.

One small bug that I don't quite understand: in Egitto, double-clicking on the "tulip" glyph does not open the right glyph. Maybe the source data contains ambiguous code points. Ohh way, the code point is `U+1F337`, so this may again be a UTF-16 issue. It was, and I fixed it.

justvanrossum:
By default, I think the default source should be selected. You can easily get it with the `fontController.fontSourcesInstancer.defaultSourceIdentifier` property.

We should store the selected source identifier in the URL fragment.

justvanrossum (commented):
Arrow keys: good start!

- I expect arrow-key up/down to go to the previous/next row.
- I find the visual difference between "hovered" and "selected" a bit too small. Maybe "hover" could be lighter, closer to the background color.
- There is an annoying delay between clicking and the selection taking effect. I think this should be immediate.

justvanrossum:
One more detail: when arrow-keying a cell that isn't 100% in-view, the container should scroll to make it in view. `UIList` uses `scrollIntoView()` for that. Please check `UIList` for the options and see if they make sense in this case.

justvanrossum:
> There is an annoying delay between clicking and the selection taking effect

I think I see what you're trying to do: to make double click a multiple selection work. But: clicking *within* a multiple selection should not change the selection to begin with. So, double-clicking on a multiple selection should open an editor with the selected glyphs, but it should _not_ reset the selection to the clicked glyph.

justvanrossum:
I noticed the fontoverview isn't yet listening to external changes (edit glyph in editor, come back to font overview: change is not reflected). I guess this will become easier once https://github.com/googlefonts/fontra/pull/1863 gets merged, so please hold off until then.

ollimeier:
> > There is an annoying delay between clicking and the selection taking effect
> 
> I think I see what you're trying to do: to make double click a multiple selection work. But: clicking _within_ a multiple selection should not change the selection to begin with. So, double-clicking on a multiple selection should open an editor with the selected glyphs, but it should _not_ reset the (font overview) selection to the clicked glyph.

There is one case where we change the selection when clicking within a multiple selection: deselect a glyph with click + commandKey.

ollimeier:
> One more detail: when arrow-keying a cell that isn't 100% in-view, the container should scroll to make it in view. `UIList` uses `scrollIntoView()` for that. Please check `UIList` for the options and see if they make sense in this case.

Thank you very much. This is awesome. `scrollIntoView` fixes it.

ollimeier:
> The link From Editor to Font Overview should remove the URL fragment. The font overview now gets the editor URL fragment, which it has no use for.

Fixed it. Thanks for noticing it.

PR #1855, opened 2024-12-15
Move sortedSourceIdentifiers to FontController

### LINKED ISSUE #1850: Move sortedSourceIdentifiers to a more general location
(This is needed for #102, #1849)

ollimeier:
Fixes #1850

It would be great to have unittests for such methods, but I don't know how as we don't have any unittests for FontController, yet – and I don't know how to implement that. 
Also, I was thinking about to refactor `SourceBox` with replacing the arguments `this.fontAxesSourceSpace` and `sources` with `this.fontController` and then call `this.fontController.getSources()` and `mapAxesFromUserSpaceToSourceSpace` within `SourceBox`, but I think this has a negative impact to the performance, as we would do this then for each source while looping through all `sortedSourceIdentifiers`, why I didn't do it. 
So, I hope this PR is good to go.

### DIFF
--- a/src/fontra/client/core/font-controller.js
+++ b/src/fontra/client/core/font-controller.js
@@ -23,7 +23,12 @@ import {
   uniqueID,
 } from "./utils.js";
 import { StaticGlyph, VariableGlyph } from "./var-glyph.js";
-import { locationToString, mapBackward, mapForward } from "./var-model.js";
+import {
+  locationToString,
+  mapAxesFromUserSpaceToSourceSpace,
+  mapBackward,
+  mapForward,
+} from "./var-model.js";
 
 const GLYPH_CACHE_SIZE = 2000;
 const BACKGROUND_IMAGE_CACHE_SIZE = 100;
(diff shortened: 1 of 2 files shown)


justvanrossum (approved):

justvanrossum:
> It would be great to have unittests for such methods, but I don't know how 

I have an idea for starting unittesting FontController and more, and it will be based on the .fontra format: for testing we should implement a class that reads .fontra from disk (Node-only), and is (minimally) compatible with the "remote" font object. We'd need a csv reader for this: we need to check npm. The rest of the data is JSON, and should be trivial to read. (.fontra uses 'unpacked paths' on disk, so we'd need to convert them to PackedPath upon reading, but that's more or less the only diversion from our in-memory object model.)

If you want you can write this up as an issue, so we can turn this into something actionable. Solving this will be very valuable long-term.

justvanrossum:
> Also, I was thinking about to refactor SourceBox with replacing the arguments [ ... ]

Maybe later, we don't need it like we do need `sortedSourceIdentifiers`.

Issue #1768, opened 2024-11-07
Lock button tooltip


navv-1:
I just noticed the lock button tooltip is slightly clipped at the top.

![Image](https://github.com/user-attachments/assets/36228a05-8e53-4e8e-8488-5c0d7095f7aa)

giovanniTramonto:
When removing `overflow-x` + `overflow-y` from `ui-form` and `overflow-y` from `div.ui-form` clipping is gone. 
Overflow could be set in `.selection-info` instead, but would require some general layout adjustments.
Actually, https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_anchor_positioning would be great, because anchor can be placed inside `overflow: hidden`, but the module was only releases this year in Chrome and Edge.

justvanrossum:
Fixed by https://github.com/googlefonts/fontra/pull/1861

justvanrossum:
While I like the concept of anchor positioning, I don't like to depend on a (huge!) polyfill. Maybe we should come back to this once anchor positioning is supported everywhere.

PR #1861, opened 2024-12-16
fix: lock-button-tooltip #1768

giovanniTramonto:
+ handle `ui-form` overflow and padding on `:host` (to prevent clipping)
+ add form modifier class for `selection-info` and `selection-transformation` (keeps flex settings within parent element scope)

### DIFF
--- a/src/fontra/client/web-components/ui-form.js
+++ b/src/fontra/client/web-components/ui-form.js
@@ -14,7 +14,6 @@ export class Form extends SimpleElement {
   static styles = `
     :host {
       --label-column-width: 32%;
-      
       padding: 1em;
       overflow: hidden auto;
     }
--- a/src/fontra/client/web-components/ui-form.js
+++ b/src/fontra/client/web-components/ui-form.js
@@ -14,14 +14,17 @@ export class Form extends SimpleElement {
   static styles = `
     :host {
       --label-column-width: 32%;
+      
+      padding: 1em;
+      overflow: hidden auto;
     }
+
     .ui-form {
       display: grid;
       align-items: center;
       grid-template-columns: var(--label-column-width) auto;
       box-sizing: border-box;
       gap: 0.35rem 0.35rem;
-      overflow-y: auto;
       margin: 0em;
       padding: 0em;
     }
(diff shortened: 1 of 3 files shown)


ollimeier:
@justvanrossum I checke out the branch and tested it. I can confirm that it fixes the issue with the tooltip:

Screenshot 2024-12-16 at 20 55 11

justvanrossum:
Does adding padding to `:host` have any downsides?

ollimeier:
> Does adding padding to `:host` have any downsides?

I sadly don't have experiences with `:host`, but I don't read about any downsides. You can simply style `:host` as any other CSS class.
Please see: https://developer.mozilla.org/en-US/docs/Web/CSS/:host

The following might also be interesting to read: 
https://stackoverflow.com/questions/46448385/use-of-host-selector-vs-container-div#answer-46478233

I am wondering if this should be done in other places as well?

justvanrossum:
My question wasn't about `:host`, but about adding the padding to the ui-info itself instead of where it is being used.

I'm curious if there is a difference in appearance before and after:
- with long axis names in the selection info panel
- the scroll bar of the selection info panel

giovanniTramonto:
The branch fixes the clipping issue. But there is more. I’d propose a component `panel-section` that concerns the overflow and wraps any contents inside. It would allow multiple scrollable boxes in the panel. Then `overflow` does not need to be a part of `ui-form` anymore.

ollimeier:
> My question wasn't about `:host`, but about adding the padding to the ui-info itself instead of where it is being used.
> 
> I'm curious if there is a difference in appearance before and after:
> 
> * with long axis names in the selection info panel
> * the scroll bar of the selection info panel

@justvanrossum I don't see any differences. Here are some screenshots in compare with our main branch:
Screenshot 2024-12-17 at 04 09 34
Screenshot 2024-12-17 at 04 08 59
Screenshot 2024-12-17 at 04 08 56
Screenshot 2024-12-17 at 04 09 37

justvanrossum:
I don't think we should – at this time – go beyong fixing the issue at hand, though. This looks good, and I will play with it a bit, and then merge.

Btw. please sign that CLA thing....

justvanrossum (approved):

giovanniTramonto:
> The branch fixes the clipping issue. But there is more. I’d propose a component `panel-section` that concerns the overflow and wraps any contents inside. It would allow multiple scrollable boxes in the panel. Then `overflow` does not need to be a part of `ui-form` anymore.

https://github.com/googlefonts/fontra/pull/1868

PR #1866, opened 2024-12-17
Move findParentWithClass to ui-utils.js

### LINKED ISSUE #1865: Move findParentWithClass to a more general location
`findParentWithClass` currently lives in **panel-related-glyphs**, but it's now also needed for **font overview**, therefore we may want to move or refactor it.

ollimeier:
Fixes #1865

In preparation for #102, because `findParentWithClass` is also needed for the **font overview** when using the `GlyphCell`.

### DIFF
--- a/src/fontra/client/web-components/ui-form.js
+++ b/src/fontra/client/web-components/ui-form.js
@@ -14,14 +14,16 @@ export class Form extends SimpleElement {
   static styles = `
     :host {
       --label-column-width: 32%;
+      padding: 1em;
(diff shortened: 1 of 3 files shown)


justvanrossum (approved):

justvanrossum:
On the other hand, this function seems to only be used to find some internal container for the Accordion view. Maybe it is better to add a method to the `Accordion` class:

- `toggleAccordionItem(item, onOff)`

Then we have this functionality nicely abstracted where it belongs.

PR #1868, opened 2024-12-17
add section to panel view

giovanniTramonto:
This adds the possibility to wrap panel contents in multiple auto-scrollable sections. So overflow logic is separated from contents.

**Changes**

+ removed `getContentElement` function for wrapper consistency
+ added `getPanelSection` factory to handle overflow of contents (this removed the extraneous styles from `ui-form`)
+ deferred use of `static styles` in panel class extensions 
+ implement panel sections across the application

### DIFF
--- a/src/fontra/client/web-components/ui-accordion.js
+++ b/src/fontra/client/web-components/ui-accordion.js
@@ -5,14 +5,12 @@ import { enumerate } from "/core/utils.js";
 export class Accordion extends UnlitElement {
   static styles = `
   .ui-accordion-contents {
-    padding: 1em;
     display: grid;
     grid-template-rows: auto;
     align-content: start;
     gap: 0.5em;
     text-wrap: wrap;
     width: 100%;
-    height: 100%;
     box-sizing: border-box;
   }
 
--- a/src/fontra/client/web-components/ui-form.js
+++ b/src/fontra/client/web-components/ui-form.js
@@ -14,8 +14,6 @@ export class Form extends SimpleElement {
   static styles = `
     :host {
       --label-column-width: 32%;
-      padding: 1em;
-      overflow: hidden auto;
     }
 
     .ui-form {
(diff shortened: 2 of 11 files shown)


giovanniTramonto:
Yes, double checked.

ollimeier:
@justvanrossum I had a look at the branch and I would say mostly it looks good. There is one change, which is consistent now with the other sidebars, but differs in the user experience in compare to our current main branch: the glyphs-search-field scrolls 'away' and I am not sure if we want that. Please see the recording:


https://github.com/user-attachments/assets/b76f8dc1-49d4-4989-b674-619d3307593d


It would be an exception. I am not sure, how we want to deal with it.

justvanrossum:
> the glyphs-search-field scrolls 'away' and I am not sure if we want that

I am very sure we _don't_ want that. So this is an unacceptable change.

I am not sure of the motivation of this PR. What exactly does this improve and why do we need it?

giovanniTramonto:
Thanks @ollimeier for checking. Search scroll is adjusted to prior style.
For a follow up: would be nicer if the scroll bar would appear not inside the glyph list, but outside in the padding space. So glyph content has a bit more space and would be consistent. 

@justvanrossum Improvements of this MR:
+ consistency in panel
+ isolate panel overflow css from content (i.e. overflow is out of ui-form)
+ css improvements (i.e. in `/panel-related-glyphs.js` the `height: calc(100% - 2em); // Would be nice to do without the calc` is gone)
+ less code

giovanniTramonto:
I guess the changes of this MR are too extensive. See this instead: https://github.com/googlefonts/fontra/pull/1877

justvanrossum:
> I guess the changes of this MR are too extensive

Not necessarily: I just didn't have time to properly test this. But if you feel your new PR is to be preferred anyway, I'll gladly review that one instead.

giovanniTramonto:
> > I guess the changes of this MR are too extensive
> 
> Not necessarily: I just didn't have time to properly test this. But if you feel your new PR is to be preferred anyway, I'll gladly review that one instead.

Yes, thanks, I feel better with a this small change for now. Even though I guess it would be good to make the panel code more consistent. Maybe in a follow up.

PR #1870, opened 2024-12-17
Fix casting of object properties

### LINKED ISSUE #1869: Bug: using knife tool with multiple editor windows
When using the knife tool in one editor window another editor window breaks.
I first noticed it while testing **#1863 Clarify server API**, but it's not related to this branch. It breaks in our **main** branch as well.

https://github.com/user-attachments/assets/d976d577-808d-466c-9f51-d36799b8fa08

Error:
Image

justvanrossum:
This fixes #1869.


ollimeier (approved):
Works well. Thanks for the fix.

PR #1871, opened 2024-12-17
Add method showHideAccordionItem to Accordion class instead of findParentWithClass

ollimeier:
Please see: https://github.com/googlefonts/fontra/pull/1866#issuecomment-2547865463

### DIFF
--- a/src/fontra/views/editor/panel-related-glyphs.js
+++ b/src/fontra/views/editor/panel-related-glyphs.js
@@ -148,6 +148,7 @@ export default class RelatedGlyphPanel extends Panel {
 
     for (const item of this.accordion.items) {
       this._updateAccordionItem(item, glyphName, codePoint).then((hasResult) => {
+        this.accordion.showHideAccordionItem(item, hasResult);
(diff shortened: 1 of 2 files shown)


src/fontra/views/editor/panel-related-glyphs.js
```diff
@@ -148,6 +148,7 @@ export default class RelatedGlyphPanel extends Panel {
 
     for (const item of this.accordion.items) {
       this._updateAccordionItem(item, glyphName, codePoint).then((hasResult) => {
+        this.accordion.showHideAccordionItem(item, hasResult);
```
justvanrossum, line 151: Why is this done in a different place than the original `parent.hidden = hideAccordionItem`? If it was done in `_updateAccordionItem()` before, I think it should still b done there, or you'll change the bahavior.

justvanrossum: Ok, I now see that it is equivalent, because `_updateAccordionItem()` is only called in one spot. Perhaps it is better this way.

PR #1877, opened 2024-12-20
Add scrollable sections to panel-selection-info, panel-transformation

giovanniTramonto:
+ removes overflow settings from `ui-form`
+ moves overflow to panel sections


justvanrossum:
Thanks! In terms of a followup, I'd be most interested in the related glyphs panel, and how to get rid of that `calc()` thing there.

PR #1879, opened 2024-12-20
Declare supported export formats in backend

### LINKED ISSUE #1878: Project manager should declare export formats
Currently if the project manager supplies `export-as` in its info, then Fontra assumes that it supports *all* export formats:

https://github.com/googlefonts/fontra/blob/23f45d4008e6a589a8036adcea46fb0a5e68817f/src/fontra/views/editor/editor.js#L696-L699

An alternative is for Fontra to read the `projectManagerFeatures["export-as"]` as a list of supported export formats. Happy to PR this if it's a good idea.

simoncozens:
Fixes #1878. FontraPak PR incoming.

~Also makes "find-glyphs-that-use-glyph" conditional on the server supporting it.~

### DIFF
--- a/src/fontra/core/fonthandler.py
+++ b/src/fontra/core/fonthandler.py
@@ -174,7 +174,7 @@ class FontHandler:
             self.projectManager, "getSupportedExportFormats"
         ):
             projectManagerFeatures["export-as"] = (
-                self.projectManager.getSupportedExportFormats()
+                self.projectManager.getSupportedExportFormats()  # type: ignore[union-attr]
             )
         return dict(
             name=self.backend.__class__.__name__,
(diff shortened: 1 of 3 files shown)


justvanrossum:
I'm not sure I like the addition of the `supportedExportFormats` field to the _protocol_ while `exportAs()` is entirely optional. Perhaps it's better to have an additional optional partner method called `getSupportedExportFormats()`.

Ideally the changes should not affect https://github.com/googlefonts/fontra-rcjk/blob/main/src/fontra_rcjk/projectmanager.py

simoncozens:
Fair enough, yes. Will rethink.

justvanrossum:
The `EXPORT_FORMATS` constant should probably be deleted, and this snippet can also use the dynamic list of export types:

https://github.com/googlefonts/fontra/blob/0c9659cb893b40e10a6206632824a8d1f8e529aa/src/fontra/views/editor/editor.js#L737-L748

PR #1902, opened 2024-12-30
Adjust related glyphs panel css

giovanniTramonto:
This is the follow-up of https://github.com/googlefonts/fontra/pull/1877

The change adjusts the box behavior of related glyphs panel so `calc(100% - 2em)` could be removed.

@justvanrossum I would suggest to set border-box globally. (read here https://css-tricks.com/almanac/properties/b/box-sizing/) I could improve this in a follow-up if you like.


justvanrossum (approved):
Looks good to me, thanks!

justvanrossum:
A global border-box thing sounds good to me: that keeps being a gotcha when you least expect it (for me at least).

giovanniTramonto:
Global border-box thing is here: https://github.com/googlefonts/fontra/pull/1908

PR #1908, opened 2025-01-02
set universal css box-sizing to border-box

giovanniTramonto:
Changes universal default `box-sizing` from `content-box` to `border-box`. (https://css-tricks.com/almanac/properties/b/box-sizing/)

An inconvenient side of this is that shadow-dom does not inherit such settings, however, a large part of the application components is shadow-dom. I actually wonder why? I think for most cases this is not recommendable.
As a solution here this introduces a `css/shadow-dom.js` file which gets appended in `SimpleElement` constructor (all shadow-dom components extend from this).

To avoid adjusting visual appearance  in this MR I set some elements individually back to `content-box` and marked them with `FIXME`. So they can be improved in a follow-up.

### DIFF
--- a/src/fontra/client/core/html-utils.js
+++ b/src/fontra/client/core/html-utils.js
@@ -2,7 +2,6 @@
 // object notation to construct dom elements instead of HTML.
 
 import { consolidateCalls } from "./utils.js";
-import shadowDomStyles from "/css/shadow-dom.js";
 
 export class SimpleElement extends HTMLElement {
   constructor() {
@@ -16,7 +15,7 @@ export class SimpleElement extends HTMLElement {
   }
 
   _attachStyles() {
-    this._appendStyle(shadowDomStyles);
+    this._appendStyleSheetLink("/css/shared.css");
     if (this.constructor.styles) {
       this._appendStyle(this.constructor.styles);
     }
@@ -24,6 +23,10 @@ export class SimpleElement extends HTMLElement {
 
   _appendStyle(cssText) {
     addStyleSheet(cssText, this.shadowRoot);
+  }
+
+  _appendStyleSheetLink(href) {
+    addStyleSheetLink(href, this.shadowRoot);
   }
 
   appendStyle(cssText) {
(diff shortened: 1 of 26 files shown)


justvanrossum:
My use of the shadow dom may indeed be a bit much, but at the time it felt like a good way to avoid CSS name conflicts and generally to keep things isolated from one another. It is indeed often unpleasant how completely isolated is (apart from CSS variables).

giovanniTramonto:
Yes, I see your point 👍
In vue for example there is scoped styles or css modules to isolate classes, but staying out of frameworks BEM is still very safe way to avoid conflicts.
But certainly not so important at this point. Perhaps could be simplified later, occasionly.

src/fontra/client/core/html-utils.js
```diff
@@ -15,6 +16,7 @@ export class SimpleElement extends HTMLElement {
   }
 
   _attachStyles() {
+    this._appendStyle(shadowDomStyles);
```
justvanrossum, line 19: Shouldn't we be able to append a `<link>` element that references a "proper" CSS file instead? There is a "link" convenience function in this module.

giovanniTramonto:
Created `shared.css` so global rules like `box-sizing` that can easily get referenced in shadow-DOM.
Did not want to reference `core.css` in shadow since it has a lot of styles that only concern light-DOM.

src/fontra/client/core/html-utils.js
```diff
@@ -153,6 +156,16 @@ export function addStyleSheet(cssText, element = null) {
...
 }
 
+export function addStyleSheetLink(href, element = null) {
+  if (!element) {
+    element = document.head;
+  }
+  const stylesheetLink = link();
+  stylesheetLink.href = href;
+  stylesheetLink.rel = "stylesheet";
+  element.appendChild(stylesheetLink);
```
justvanrossum, line 166: You should be able to spell this as:

```js
  element.appendChild(link({ href, rel: "stylesheet" }));
```

giovanniTramonto: Nice 👌

justvanrossum (approved):

PR #1911, opened 2025-01-03
Add a conditional statement to ensure the multi-tool mousedown timer only affects the first child

### LINKED ISSUE #1762: Selecting subtool
When selecting a subtool, it sometimes works and sometimes fails. Then I realized why it was like that.

Selecting subtool with a fast-click: **Work**
Image

Selecting subtool with a little click-hold: **Fail**
I noticed, while mousedown, the subtool is hidden until the timeout reached. It makes the selection fail.
I think it should keep visible while mousedown.
Image

Selecting subtool with a click-hold until the subtool is back to visible (timeout reached): **Work**
Image

navv-1:
fix #1762

### DIFF
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -1217,6 +1217,7 @@ export class EditorController extends ViewController {
             window.addEventListener("keydown", globalListener, false);
           }, 650);
           if (toolButton !== editToolsElement.children[0]) {
+            // ensure the multi-tool mousedown timer only affects the first child
             event.stopImmediatePropagation();
           }
         };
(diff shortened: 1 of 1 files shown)


justvanrossum:
(FWIW, it wasn't the blank line that caused the pre-commit failure, but the fact that the blank line contained whitespace.)

src/fontra/views/editor/editor.js
```diff
@@ -1216,6 +1216,9 @@ export class EditorController extends ViewController {
             window.addEventListener("mousedown", globalListener, false);
             window.addEventListener("keydown", globalListener, false);
           }, 650);
+          if (toolButton !== editToolsElement.children[0]) {
```
justvanrossum, line 1219:
```suggestion
          if (toolButton !== editToolsElement.children[0]) {
            // ensure the multi-tool mousedown timer only affects the first child
```

PR #1920, opened 2025-01-06
Remove padding from Accorion, and let clients deal with that

justvanrossum:
Hey @giovanniTramonto, I was just running into a situation where it's undesirable to have Accordion do its own padding. I think this fixes it. Just in case you'd like to have a look? I immediately benefitted from the border-box improvement.

### DIFF
--- a/src/fontra/client/web-components/ui-accordion.js
+++ b/src/fontra/client/web-components/ui-accordion.js
@@ -5,7 +5,6 @@ import { enumerate } from "/core/utils.js";
 export class Accordion extends UnlitElement {
   static styles = `
   .ui-accordion-contents {
-    padding: 1em;
     display: grid;
     grid-template-rows: auto;
     align-content: start;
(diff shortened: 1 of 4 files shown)


justvanrossum:
(`GlyphCellView` inherited that behavior from `Accordion`, hence the changes where that is being used as well.)

giovanniTramonto:
@justvanrossum It’s good to remove the padding here from accordion. Inline style in the wrapper is not so great. Also I would solve this situation differently, by creating a `panel-section` that goes in `panel` to wrap contents. (Remember the MR I started and closed) I’m still up for implementing this. At this point I totally align with the change of wrapping the accordion. Just go for it. I can rework this in a follow-up with proper classes, etc.

justvanrossum:
There is a _lot_ of inline styles in that module. I'm inclined to just leave it for now. I'll merge.

PR #1924, opened 2025-01-07
Refactor top menu bar

### LINKED ISSUE #1845: [font overview] Menu bar for font overview

ollimeier:
Fixes #1845 

This is a draft PR for open questions and early feedback.

Below you will find open questions.


ollimeier:
Info:
I added `initTopBar` to view-controller. It included for each menu item a method. This way we can modify the method in the different views as we need it. And the default menus like 'Fontra' or 'Help' stay as say are without any code duplication.

ollimeier:
I added the possibility to 'disable' a menu item via `enabled: () => false,`. My thinking here: The menu bar does not change in the number of items, which helps being consistent in the look and feel when changing between windows. But we don't need all menu items for every view. For example 'Glyph' won't be needed for font-info or application settings, therefore it can be greyed-out.

Here a screenshot:
Screenshot 2025-01-07 at 21 05 01

This is just a proposal and up for discussion, of cause.

ollimeier:
If we add the top menu bar to font-info and application settings, the left menu is not needed anymore, as we have the different items already in the drop-down of our top menu bar. Therefore the left menu can be deleted. I think this is actually good, because we gain more space, which might be especially useful for sections like 'Axes':
Screenshot 2025-01-07 at 21 14 47
Screenshot 2025-01-07 at 21 14 43
Screenshot 2025-01-07 at 21 14 02
Screenshot 2025-01-07 at 21 10 53

ollimeier:
I am not 100% happy with the variable names like `makeMenuBarSubmenuGlyph`. `makeMenuBarGlyph` sounds like it makes the menu bar of `Glyph`, but it only is one submenu of the top menu bar, why I added `Submenu`. 
What about:
`makeMenuBar_fontra`
`makeMenuBar_edit`
`makeMenuBar_glyph`
etc.

Any other ideas/proposals are very welcome.

ollimeier:
I am wondering if the following features/buttons should then go into the submenu of 'Edit' (thinking out loud):
Screenshot 2025-01-07 at 21 24 03

justvanrossum:
Superseded by #1925

PR #1925, opened 2025-01-08
Refactoring top bar / menu bar

### LINKED ISSUE #1845: [font overview] Menu bar for font overview

ollimeier:
Fixes #1845

### DIFF
--- a/src/fontra/client/core/fontra-menus.js
+++ b/src/fontra/client/core/fontra-menus.js
@@ -98,32 +98,18 @@ function getFileMenuItems() {
   return {
     title: translate("menubar.file"),
     getItems: () => {
-      let exportFormats =
-        this.fontController.backendInfo.projectManagerFeatures["export-as"] || [];
-      if (exportFormats.length > 0) {
-        return [
-          {
-            title: translate("menubar.file.export-as"),
-            getItems: () =>
-              exportFormats.map((format) => ({
-                actionIdentifier: `action.export-as.${format}`,
-              })),
-          },
-        ];
-      } else {
-        return [
-          {
-            title: translate("menubar.file.new"),
-            enabled: () => false,
-            callback: () => {},
-          },
-          {
-            title: translate("menubar.file.open"),
-            enabled: () => false,
-            callback: () => {},
-          },
-        ];
-      }
+      return [
+        {
+          title: translate("menubar.file.new"),
+          enabled: () => false,
+          callback: () => {},
+        },
+        {
+          title: translate("menubar.file.open"),
+          enabled: () => false,
+          callback: () => {},
+        },
+      ];
     },
   };
 }
--- a/src/fontra/client/core/fontra-menus.js
+++ b/src/fontra/client/core/fontra-menus.js
@@ -3,7 +3,7 @@ import { translate } from "/core/localization.js";
 import { MenuBar } from "/web-components/menu-bar.js";
 
 const mapMenuItemKeyToFunction = {
-  // File: getFileMenuItems, // TODO: this does not work, becuase viewController.fontController?.backendInfo is undefined. And I don't know why.
+  File: getFileMenuItems,
   Font: getFontMenuItems,
   Edit: getEditMenuItems,
   View: getViewMenuItems,
@@ -14,24 +14,17 @@ export function makeFontraMenuBar(menuItemKeys, viewController) {
   const menuBarArray = [getFontraMenu()]; // Fontra-Menu at the beginning.
 
   for (const itemKey of menuItemKeys) {
-    const methodName = `get${itemKey}MenuItems`;
-    let menuItems = [];
-    if (typeof viewController[methodName] === "function") {
-      menuItems = viewController[methodName]();
-    } else if (mapMenuItemKeyToFunction[itemKey]) {
-      menuItems = mapMenuItemKeyToFunction[itemKey](viewController);
-    } else if (menuItems.length === 0) {
-      console.log("Menu has not items, skip: ", itemKey);
-      continue;
-    } else {
-      console.log("Method/Function does not exist, skip: ", itemKey, methodName);
+    const getMenuItemsFunc =
+      viewController[`get${itemKey}MenuItems`] || mapMenuItemKeyToFunction[itemKey];
+    if (!getMenuItemsFunc) {
+      console.log("Method/Function does not exist, skip: ", itemKey);
       continue;
     }
 
     const menu = {
       title: translate(`menubar.${itemKey.toLowerCase()}`),
       getItems: () => {
-        return menuItems;
+        return getMenuItemsFunc(viewController);
       },
     };
     menuBarArray.push(menu);
--- a/src/fontra/client/core/fontra-menus.js
+++ b/src/fontra/client/core/fontra-menus.js
@@ -1,3 +1,4 @@
+import * as html from "./html-utils.js";
 import { translate } from "/core/localization.js";
 import { MenuBar } from "/web-components/menu-bar.js";
 
@@ -18,7 +19,15 @@ export function makeFontraMenuBar(menuItemKeys, delegate) {
   }
 
   menuBarArray.push(getHelpMenuItems()); // Help-Menu at the end.
-  return new MenuBar(menuBarArray);
+  const menuBar = new MenuBar(menuBarArray);
+
+  // add project file name to the right side of the menu bar
+  const projectNameWrapper = html.div();
+  projectNameWrapper.textContent = delegate.projectFileName;
+  projectNameWrapper.style.marginLeft = "auto";
+  menuBar.contentElement.appendChild(projectNameWrapper);
+
+  return menuBar;
 }
 
 function getFontraMenuItems() {
--- a/src/fontra/client/core/fontra-menus.js
+++ b/src/fontra/client/core/fontra-menus.js
@@ -11,7 +11,7 @@ const mapMenuItemKeyToFunction = {
 };
 
 export function makeFontraMenuBar(menuItemKeys, viewController) {
-  const menuBarArray = [getFontraMenuItems()]; // Fontra-Menu at the beginning.
+  const menuBarArray = [getFontraMenu()]; // Fontra-Menu at the beginning.
 
   for (const itemKey of menuItemKeys) {
     const methodName = `get${itemKey}MenuItems`;
@@ -24,12 +24,12 @@ export function makeFontraMenuBar(menuItemKeys, viewController) {
     }
   }
 
-  menuBarArray.push(getHelpMenuItems()); // Help-Menu at the end.
+  menuBarArray.push(getHelpMenu()); // Help-Menu at the end.
   const menuBar = new MenuBar(menuBarArray);
   return menuBar;
 }
 
-function getFontraMenuItems() {
+function getFontraMenu() {
   return {
     title: "Fontra",
     bold: true,
@@ -54,7 +54,7 @@ function getFontraMenuItems() {
   };
 }
 
-function getHelpMenuItems() {
+function getHelpMenu() {
   return {
     title: translate("menubar.help"),
     getItems: () => {
--- a/src/fontra/client/core/fontra-menus.js
+++ b/src/fontra/client/core/fontra-menus.js
@@ -1,0 +1,214 @@
+import { translate } from "/core/localization.js";
+import { MenuBar } from "/web-components/menu-bar.js";
+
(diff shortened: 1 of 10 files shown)


ollimeier:
While working on it, I was wondering if `fontra-menus.js` is really needed, because `MenuBar` is only used as the top menu bar and nowhere else.

ollimeier:
Or the `fontra-menus.js` is required when we think about adding the project name and user-icon (future vision)?

justvanrossum:
> While working on it, I was wondering if fontra-menus.js is really needed, because MenuBar is only used as the top menu bar and nowhere else.

> Or the fontra-menus.js is required when we think about adding the project name and user-icon (future vision)?

Neither:
- menu-bar.js implements "a menu bar". A general purpose menu bar UI element.
- fontra-menus.js contains helper code to build _Fontra's_ menu bar.

The first is the "how" the second is the "what". Basic sensible separation of concerns.

justvanrossum:
Speaking of project name and user icon: this explains the terms "top bar" vs. "menu bar":
- The menu bar is the part with menus
- The top bar is the area at the top of the window in which the menu bar lives, and where we will _also_ add project name and user icon

src/fontra/client/core/view-controller.js
```diff
@@ -37,6 +37,7 @@ export class ViewController {
 
   constructor(font) {
     this.fontController = new FontController(font);
+    this.projectFileName = font.wsURL.split("/").slice(-1)[0];
```
justvanrossum, line 40: You can't do this, as this breaks the abstraction: the `wsURL` part is specific to the web socket implementation. This will break Simon's code.

src/fontra/client/css/core.css
```diff
@@ -18,6 +18,7 @@
 :root {
   /* https://css-tricks.com/the-css-custom-property-toggle-trick/ */
   --fontra-theme-marker: ; /* a space, which is valid */
+  --menu-top-bar-height: 35px;
```
justvanrossum, line 21: We have a "top bar" that contains a "menu bar". "menu top bar" makes no sense.

src/fontra/client/web-components/menu-bar.js
```diff
@@ -1,14 +1,24 @@
...
   static styles = `
+
+  ${themeColorCSS(colors)}
+
   .menu-bar {
     display: flex;
     align-items: center;
     font-size: 1rem;
-    height:100%;
+    height: var(--menu-top-bar-height);
```
justvanrossum, line 21: The menu bar should not depend on an external var. Please keep it "100%" as it was, and if that doesn't work anymore, a different solution is needed.

src/fontra/views/fontoverview/fontoverview.js
```diff
@@ -50,6 +63,77 @@ export class FontOverviewController extends ViewController {
     );
   }
 
+  getFileMenuItems() {
```
justvanrossum, line 66: Why isn't this code in fontra-menus.js? Seems like the editor and font info should have this in common.

src/fontra/views/fontoverview/fontoverview.js
```diff
@@ -235,6 +328,104 @@ export class FontOverviewController extends ViewController {
...
+
+  async zoomIn() {
+    console.log("font overview zoom in");
+  }
+
+  async zoomOut() {
+    console.log("font overview zoom out");
+  }
+
+  async duplicateGlyph() {
```
justvanrossum, line 420: Please do not invent new functionality. Do the basics first. Stubs for copy/paste/delete/undo etc. is fine. But this is just a random idea that isn't fully thought-through.

justvanrossum (commented):
Not a full review yet, but I added some comments.

Let's NOT add the project name (not project *file* name, you don't _know_ that it is a file), but do it later as a followup. As I commented, this is not going to be part of the menu bar, but of the _top bar_ (which _contains_ the menu bar), so this factorisation is wrong. It is also clearly a separate task that requires more thought and design.

src/fontra/client/web-components/menu-bar.js
justvanrossum: This is again a matter of composition: the *top bar* height will be specified with a var. The menu bar that it contains has no business with a hard coded height. This should (become) the hierarchy:

```

+-- top bar ---------------------------------------------------------------+
| +-------------------------------+        +--------------+  +-----------+ |
| | menu bar                      |        | project name |  | user icon | |
| +-------------------------------+        +--------------+  +-----------+ |
+--------------------------------------------------------------------------+
```

src/fontra/client/core/view-controller.js
ollimeier: Ok. I removed this line of code + the part wich will add fileName to the menu bar.

src/fontra/client/css/core.css
ollimeier: ok, before my changes, they were called `--editor-top-bar-XXX`. Should I rename it to `--top-bar-XXX`? 
For me it's not clearly separated from each other, for example `--editor-top-bar-link-hover` was used in the css classes `.menu-item.current {`, which would be more about "menu bar", then "top bar".
Let me think about it. Maybe there is more about the whole idea of separating between "top bar" and "menu bar".

src/fontra/client/web-components/menu-bar.js
ollimeier: Thanks. I changed it.

src/fontra/client/css/core.css
justvanrossum: Yes, there was indeed already some confustion between "top bar" and "menu bar". Moving some of these to the menu-bar module fixes most of this, no?

I would name this particular variable `--top-bar-height`.

src/fontra/views/fontoverview/fontoverview.js
ollimeier: Good point. I moved it to fontra-menus.js. I had applications settings in mind, where it might not be used. But in this case, we can skip the 'File' menu item.

src/fontra/client/core/fontra-menus.js
```diff
@@ -0,0 +1,222 @@
...
+    },
+  };
+}
+
+function getFileMenuItems(delegate) {
+  return {
+    title: translate("menubar.file"),
+    getItems: () => {
+      let exportFormats =
+        delegate.fontController?.backendInfo.projectManagerFeatures["export-as"] || [];
```
justvanrossum, line 100: The "delegate" is meant to be an abstraction that *only* provides optional methods to populate menus. This now assumes that it is an object with a "fontController" propery. Either we should let go of the "delegate" abstraction, and just call the thing "viewController", or we need to solve this differently. Let me think.

justvanrossum: Ok, I thought about it. Let's bite the bullet and rename "deletage" to "viewController" everywhere in this module.

src/fontra/views/fontoverview/fontoverview.js
ollimeier: I added all the stubs for discussion, because even the basics are not clear to me where they belong. Usually we have copy/paste/delete/undo under "Edit". But this is based on the the selected source in the editor view, am I right?
As we can switch between sources, would a "copy" in the menu bar "Edit" copy only the path of the source selected or the whole glyph with all sources? Where does Delete belongs "Edit" or "Glyph"? Would it be more like `action.delete-selection`, where we can delete multiple glyphs based on the 'Selection'? 
I removed `duplicateGlyph`, but did not add copy/paste or others. I think it would be good to discuss it first.

src/fontra/client/core/fontra-menus.js
```diff
@@ -0,0 +1,222 @@
...
+            callback: () => {},
+          },
+        ];
+      }
+    },
+  };
+}
+
+// TODO: Is a default possible, or do we skip it if not provided by the 'View'?
+// function getEditMenuItems() {
```
justvanrossum, line 130: Yes, it should be populated with action identifiers.

src/fontra/views/fontoverview/fontoverview.js
justvanrossum: I just commented "it should be populated with action identifiers" for the standard edit items, but I realize that doesn't give us menu titles if they're not defined. *Maybe* they should be: action idnetifier + default title.

There should be no functionality here at all. Views can register actions, and that will automatically enable the menu item appropriately.

src/fontra/client/core/fontra-menus.js
justvanrossum: ... and fallback default titles, as commented in the other comment thread.

src/fontra/client/core/fontra-menus.js
```diff
@@ -0,0 +1,222 @@
...
+          callback: () => {
+            window.open("https://github.com/googlefonts/fontra");
+          },
+        },
+      ];
+    },
+  };
+}
+
+function getFileMenuItems(delegate) {
```
justvanrossum: Oh, we had a misunderstanding here, that is shown in the function name. These functions return menu descriptors, not menu *items*. The delegate (viewController) should *not* be responsible for the whole menu, but only for the getItems() part. Hence the name.

With the Fontra and Help menu's it's differnt: it is useful to build the whole menu in a helper func like you do, but the function names are bad. `getFontraMenuItems()` doesn't return items, but a whole menu. So it should be called `getFontraMenu()`. Same with the Help menu.

justvanrossum: So the fallback functions (like `getFileMenuItems()`) should probably _also_ just do the "getItems()" part.

src/fontra/client/css/core.css
ollimeier: Done:
- Removed "top" from "--menu-top-bar-XXX" as it belongs to menu-bar only
- Removed unused "menu-top-bar-border-color"
- "--top-bar-height"

src/fontra/client/core/fontra-menus.js
ollimeier: So, I made the 'get' functions return only items, not descriptors. But `getFileMenuItems` does not work anymore, because `viewController.fontController.backendInfo` is `undefined` and I don't know why. It makes no sense to me. I do a little break now.

src/fontra/client/core/fontra-menus.js
```diff
@@ -0,0 +1,195 @@
...
+  Font: getFontMenuItems,
+  Edit: getEditMenuItems,
+  View: getViewMenuItems,
+  Glyph: getGlyphMenuItems,
+};
+
+export function makeFontraMenuBar(menuItemKeys, viewController) {
+  const menuBarArray = [getFontraMenu()]; // Fontra-Menu at the beginning.
+
+  for (const itemKey of menuItemKeys) {
```
justvanrossum, line 16: Independent of the problem with the fontController you are seeing: you should *not* precompute the menu items. The should be constructed when getItems() is *called*.

ollimeier: That was exactly the hint I needed. Thanks. It solves the other issue as well. I will push something in a bit.

ollimeier: Ok, it's fixed now.

src/fontra/client/core/fontra-menus.js
ollimeier: Is fixed now.

src/fontra/views/fontoverview/fontoverview.js
ollimeier: I added action identifiers for `getEditMenuItems` and `getViewMenuItems` as defaults. `getGlyphMenuItems` is empty. I don't know what this should contain as defaults. I keep this feedback still open, because it's still not full clear to me, where default items should belong to which menu.

src/fontra/client/core/fontra-menus.js
ollimeier: I added action identifiers for `getEditMenuItems` and `getViewMenuItems` as defaults. I am closing this.

ollimeier:
@justvanrossum Question: Do we want to add this top bar (including the menu bar) into font-info and applications settings as well or is this a follow up?

justvanrossum:
> Do we want to add this top bar (including the menu bar) into font-info and applications settings as well or is this a follow up?

Followup. Smaller PRs == Better PRs.

PR #1926, opened 2025-01-08
Panel component

giovanniTramonto:
Here is the proposal for panel-sections that I mentioned here https://github.com/googlefonts/fontra/pull/1920

I placed it in `/web-componets` to not conflict with `/views`, but could be moved later if preferred, of course.
Its a custom element instead of shadow. CSS inheritance persists.
I adjusted `addStyleSheet` in utils so using custom-elements multiple times loads the style only once. 
Actually wondering if the use of private properties in classes meets the project conventions, but found it useful to isolate local styles from extends.

I applied it only to DesignSpace-Navigation for now for this proposal. But using it all across the sidebars will improve consistency and make a lot of style repetitions redundant.

### DIFF
--- a/src/fontra/client/core/html-utils.js
+++ b/src/fontra/client/core/html-utils.js
@@ -148,11 +148,13 @@ export function htmlToElements(html) {
   return template.content.childNodes;
 }
 
-export function addStyleSheet(cssText, element = null) {
+export function addStyleSheet(cssText, element = null, id = null) {
   if (!element) {
     element = document.head;
   }
-  element.appendChild(style({}, [cssText]));
+  if (!document.querySelector(`style#css-${id}`)) {
+    element.appendChild(style(id ? { id: `css-${id}` } : {}, [cssText]));
+  }
 }
 
 export function addStyleSheetLink(href, element = null) {
--- a/src/fontra/client/web-components/panel.js
+++ b/src/fontra/client/web-components/panel.js
@@ -1,0 +1,56 @@
+import { addStyleSheet, div } from "/core/html-utils.js";
+
+export default class Panel extends HTMLElement {
+  #cssName = "panel";
+  #styles = `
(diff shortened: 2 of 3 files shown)


justvanrossum:
I will need a bit of time to digest this.

Private properties is indeed something we don't yet use (mostly because I didn't know about them at the start of the project), but I totally see how they are useful, even if the syntax is a bit jarring if you aren't used to it.

I'm not sure about that change in addStyleSheet: you check the document for presence of the style sheet, but add it to the element if it's not in the document. This feels a bit weird to me.

giovanniTramonto:
Yes, it’s a bit crude. Is there another approach in the project for appending multiple components? 
I mean actually, the .panel css module is just a few lines, it could simply go to `editor.css`. Of course, ideally styles and template are in the same place, but that would add a lot of complexity, I guess.
Perhaps better to bury the approach in this MR another time and to simply add a some editor css classes that can be used …

giovanniTramonto:
I tried this out https://github.com/giovanniTramonto/fontra/commit/2fc0fe3664ce1dc0571f2cbfbb24dedee8ad2de8
Styles got simpler, template bigger 😄
But more consistent.

justvanrossum:
But now the "template for everything contains" styles for something specific, that may not be needed for everything that includes "shared.css". So that feels a bit off to me.

giovanniTramonto:
Yes, would be a work around :/
Actually, utility classes could be really helpful in this scenario. I’ll give it a try later.

giovanniTramonto:
Ah, of course, it can simply go to `panel.js` https://github.com/giovanniTramonto/fontra/commit/865563e1a7aa59b2571a9ba01caa3e1938eee73b

giovanniTramonto:
One more try https://github.com/googlefonts/fontra/pull/1932

PR #1932, opened 2025-01-10
Add reusable css classes to panel view

giovanniTramonto:
Another attempt to make panel styles more consistent.
This time logic of `panel.js` did not change, but only reusable styles were added.
The node structure of all inheriting panels got harmonized to `.panel` wrapper containing `panel-section` and made a lot of styles redundant.

### DIFF
--- a/src/fontra/client/css/shared.css
+++ b/src/fontra/client/css/shared.css
@@ -36,3 +36,7 @@ button,
 input[type="button"]:focus {
   outline: none;
 }
+
+.text-wrap {
+  white-space: wrap;
+}
--- a/src/fontra/client/web-components/add-remove-buttons.js
+++ b/src/fontra/client/web-components/add-remove-buttons.js
@@ -34,6 +34,7 @@ class AddRemoveButtons extends html.UnlitElement {
       background-color: var(--button-color);
       fill: var(--text-color);
       border: none;
+      line-height: 0;
     }
 
     button > inline-svg {
--- a/src/fontra/client/web-components/icon-button.js
+++ b/src/fontra/client/web-components/icon-button.js
@@ -5,6 +5,10 @@ import { FocusKeeper } from "/core/utils.js";
 
 export class IconButton extends UnlitElement {
   static styles = `
+    :host {
+      line-height: 0;
+    }
+
     button {
       background-color: transparent;
       border: none;
@@ -36,7 +40,6 @@ export class IconButton extends UnlitElement {
     button:disabled svg {
       transform: none;
     }
-
   `;
 
   constructor(src) {
--- a/src/fontra/client/web-components/ui-accordion.js
+++ b/src/fontra/client/web-components/ui-accordion.js
@@ -50,7 +50,7 @@ export class Accordion extends UnlitElement {
 
   .ui-accordion-item-content {
     display: block;
-    overflow: hidden;
+    overflow: auto;
   }
   `;
 
--- a/src/fontra/views/editor/panel-designspace-navigation.js
+++ b/src/fontra/views/editor/panel-designspace-navigation.js
@@ -209,7 +209,9 @@ export default class DesignspaceNavigationPanel extends Panel {
       },
     ];
 
-    return html.div({ style: "height: 100%; padding: 1em;" }, [this.accordion]);
+    return html.div({ class: "panel" }, [
+      html.div({ class: "panel-section panel-section--flex" }, [this.accordion]),
+    ]);
   }
 
   get fontAxesElement() {
--- a/src/fontra/views/editor/panel-designspace-navigation.js
+++ b/src/fontra/views/editor/panel-designspace-navigation.js
@@ -210,7 +210,9 @@ export default class DesignspaceNavigationPanel extends Panel {
     ];
 
     return html.div({ class: "panel" }, [
-      html.div({ class: "panel-section panel-section--flex" }, [this.accordion]),
+      html.div({ class: "panel-section panel-section--flex panel-section--noscroll" }, [
+        this.accordion,
+      ]),
     ]);
   }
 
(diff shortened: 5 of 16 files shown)


giovanniTramonto:
Thanks for reviewing and no worries, I was busy with other stuff too : )
Branch is rebased, including fixes for the two issues above.

justvanrossum:
The scroll behavior in the Designspace Navigation panel has changed: the individual accordion sections should scroll, but in your branch it doesn't scroll at all:

image

Before:


https://github.com/user-attachments/assets/05641f34-8bf1-499a-8691-8ee05c2c7d3c

giovanniTramonto:
Should work now. I have not seen this, because my local font does not have a lot of values in there. This is tricky. Are there fonts for tests, so we can ensure to see the same?

justvanrossum:
Here's one I just made for this:

[ComplexVariableGlyph.fontra.zip](https://github.com/user-attachments/files/18535928/ComplexVariableGlyph.fontra.zip)

Maybe we should have a folder in the repo with test fonts like this, to (manually) test UI edge cases.

justvanrossum:
I can confirm that the scroll issue is gone now, though.

justvanrossum:
But I'm not sure the fix is correct or complete: now the font overview sidebar has conflicting scroll behavior:
- the whole panel scrolls (that isn't necessarily the desired behavior, but that's what it was. TBH, I'm undecided about this, but I couldn't figure out the "per accordion item" scroll behavior in this context, so if you could show how it could be done for each so we can experience the difference, that would be great)
- *and* individual sections scroll, but poorly.


https://github.com/user-attachments/assets/e648fb6c-26ee-4e2a-a9bc-dc4932ec27d7




To reproduce this just add many preset glyph set to both glyphset sections with the "+" buttons.

Note there's some fiddling with accordion styling in [glyph-cell-view.js](https://github.com/googlefonts/fontra/blob/08523e0d73d15e55752fc30dd9ae19c8de45ddea/src/fontra/client/web-components/glyph-cell-view.js#L93-L101). This view is used in the fontoverview as well as in the related glyphs panel. In the related glyphs panel it should scroll per item, in the font overview the whole thing should scroll. This is behaving correctly both in main as in your branch, but perhaps you have suggestions to clean this up a little.

giovanniTramonto:
Thanks for the test font. I guess a folder in the repo is a very good idea. Could also be used later for unit or e2e tests.
I will take some time to go over this asap, max middle of next week.

justvanrossum:
Thank you so much, no rush at all.

giovanniTramonto:
Scroll behavoir of font-overview sidebar is adjusted now to align with editors designspace-navigation.
It could be good (later on) to turn `panel.js` into a core component so it could be shared across pages.

giovanniTramonto:
Also added a `text-wrap` helper since I understood that the whole document is set to `nowrap`. Actually, I would prefer the opposite, removing `nowrap` from `html, body` and create a helper like `text-nowrap`. But that’s not part of this MR, of course.

giovanniTramonto:
Regarding [glyph-cell-view.js](https://github.com/googlefonts/fontra/blob/08523e0d73d15e55752fc30dd9ae19c8de45ddea/src/fontra/client/web-components/glyph-cell-view.js#L93-L101) I am happy to adjust in a different MR

giovanniTramonto:
> Regarding [glyph-cell-view.js](https://github.com/googlefonts/fontra/blob/08523e0d73d15e55752fc30dd9ae19c8de45ddea/src/fontra/client/web-components/glyph-cell-view.js#L93-L101) I am happy to adjust in a different MR

👉 (https://emojipedia.org/de/nach-rechts-weisender-zeigefinger)https://github.com/googlefonts/fontra/pull/1995

justvanrossum:
While the overall behavior in the font overview sidebar is now correct, there is still a scroll wobble when there's plenty space:


https://github.com/user-attachments/assets/1996f18b-d5a2-4250-9d2b-30b11b03bb90


Could this be a border-box thing?

giovanniTramonto:
The bug is caused by `icon-button` height. Once its removed, it’s fine. 
Bildschirmfoto 2025-01-31 um 14 06 04
Should I fix this within this MR? I would also suggest to refactor the layout of the accordion items. Most of the `grid` use is not necessary. Could be much simpler.

justvanrossum:
If you can fix that scrolling there here, that would be great. I'm open to your other proposed simplifications, but would prefer a new PR.

giovanniTramonto:
OK, fixed. I set the icon size a little bit smaller.

justvanrossum:
> I set the icon size a little bit smaller.

Hm, I don’t understand why the appearance should change to fix that scroll behavior.

giovanniTramonto:
The problem was caused by the height of the icon. Maybe too high for the line, which is a bit odd, indeed.

justvanrossum:
But that problem didn’t exist before your other changes, and I really don’t like changing the size to fix a problem that wasn’t there before.

giovanniTramonto:
OK Sure. I can investigate further

giovanniTramonto:
The icon buttons create the excess height even though their computed height is the same as the line. Odd. I test `box-sizing`, but did not have an effect. The `overflow: hidden` fixes it.

giovanniTramonto:
(I also spot the same issue in designspace-navigation)

giovanniTramonto:
I compared with `main` branch and found out that the issue with the icons would appear too, as soon as accordion’s content was set to `overflow: auto`. So I investigated further, and of course, it is a `line-height` issue (already thought it :-). The icons are set to appear in square dimension but this does not prevent the line-height’s effect on its surroundings, however, you can’t measure this in inspector. So `line-height: 0` does it.

justvanrossum:
(Btw if you activate pre-commit in for fontra folder, formatting fails like these can be prevented)

giovanniTramonto:
> (Btw if you activate pre-commit in for fontra folder, formatting fails like these can be prevented)

OK Will activate in the future

giovanniTramonto:
Scrolling in the font-reference panel is working now.

justvanrossum (approved):

justvanrossum:
Thank you!

Issue #1919, opened 2025-01-06
Make Javascript imports consistent


simoncozens:
There are three kinds of Javascript import: *relative* (`import { Foo } from "./foo.js"`, often used by people who bundle their imports into a single deliverable); *root-relative* (`import { Foo } from "/foo.js"`, often used by people who deploy everything on the web) and the ES2020 *absolute* (`import { foo } from "foo.js"`).

Currently Fontra uses an inconsistent mix of relative and root-relative:

<img width="1036" alt="Image" src="https://github.com/user-attachments/assets/0417fc2c-b26b-45a7-9752-6ad67cfe3819" />

Inconsistency is confusing in itself, because it makes developers question why one thing is one way and another thing is another, but as you can see from the screenshot, this also confuses code analysis tools.

Root-relative imports confuse them because `/core/var-glyph.js` is interpreted as an absolute filesystem path, which is manifestly not true; but this can be worked around potentially by tweaking an import map. And relative imports confuse them similarly because the import is just not correct from a filesystem point of view: in this example, `src/fontra/views/editor/edit-tools-pointer.js` includes `../core/vector.js`, which you would think means `src/fontra/views/core/vector.js` but is intended to mean `src/fontra/core/vector.js`! (And code analysis tools are helpful because, as you can see above, they're good at finding unused imports!)

I propose that `core` (and `web-components`) imports become ES2020 absolute imports:

```javascript
import { VarPackedPath } from "core/var-path.js";
import * as vector from "core/vector.js";
import { EditBehaviorFactory } from "./edit-behavior.js";
import { BaseTool, shouldInitiateDrag } from "./edit-tools-base.js";
import { getPinPoint } from "./panel-transformation.js";
import {
  registerVisualizationLayerDefinition,
  strokeRoundNode,
   strokeSquareNode,
 } from "./visualization-layer-definitions.js";
import { translate } from "core/localization.js";
import { copyBackgroundImage, copyComponent } from "core/var-glyph.js";
```

I have this working locally and can make a PR if the idea is agreed.

justvanrossum:
1. This may break the cache busting logic, which I think currently expects import strings to start with "/" or "./". This could be fixed by tweaking the assiciated regex.
2. The "views" are "plugins" registered in pyproject.toml, and can live *anywhere*. The import paths in them are only valid at runtime. I'm not surprised code analysis tools are getting confused. Perhaps an import map can help here?
[...]

PR #1935, opened 2025-01-10
Change import assertion to import attribute

simoncozens:
See https://github.com/googlefonts/fontra/issues/1919#issuecomment-2573346423

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,7 @@ jobs:
       matrix:
         os: [windows-latest, ubuntu-latest]
         python-version: ["3.10", "3.12"]
-        node-version: [19.x]
+        node-version: [23.x]
         exclude:
           - os: windows-latest
             python-version: "3.10"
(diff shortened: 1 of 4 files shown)


justvanrossum:
If this approach is cumbersome, alternatively we can add a variant of getTestData in test-support.js, and read this file explicitly.

simoncozens:
Just debugging - tests are passing for me locally although prettier is unhappy (which is odd because support for import attributes landed in [prettier 3.0.0](https://prettier.io/blog/2023/07/05/3.0.0.html#support-import-attributes-14861httpsgithubcomprettierprettierpull14861-14863httpsgithubcomprettierprettierpull14863-by-sosukesuzukihttpsgithubcomsosukesuzuki)). So I'm going to have to juggle a few versions.

justvanrossum:
What is the lowest Node version that supports the "import / with" syntax?

simoncozens:
20.10.0

justvanrossum:
I'm preferring this approach: https://github.com/googlefonts/fontra/pull/1940

I don't like to be too picky about which Node version to use.

simoncozens:
Fair enough, simpler is better.

PR #1947, opened 2025-01-15
Add placeholder if no glyphs are found

### LINKED ISSUE #1946: [font overview] Show placeholder text if no glyphs are visible

ollimeier:
Fixes #1946

Screenshot 2025-01-15 at 17 03 15

### DIFF
--- a/src/fontra/views/fontoverview/fontoverview.js
+++ b/src/fontra/views/fontoverview/fontoverview.js
@@ -167,7 +167,13 @@ export class FontOverviewController extends ViewController {
     await this.fontController.subscribeChanges(rootSubscriptionPattern, false);
 
     const sidebarContainer = document.querySelector("#sidebar-container");
-    this.glyphCellViewContainer = document.querySelector("#glyph-cell-view-container");
+    const glyphCellViewContainer = document.querySelector("#glyph-cell-view-container");
+
+    glyphCellViewContainer.appendChild(
+      html.div({ id: "font-overview-no-glyphs", class: "hidden" }, [
+        translate("(No glyphs found)"), // TODO: translation
+      ])
+    );
 
     this.navigation = new FontOverviewNavigation(this);
 
@@ -185,7 +191,7 @@ export class FontOverviewController extends ViewController {
     this.glyphCellView.onOpenSelectedGlyphs = (event) => this.openSelectedGlyphs();
 
     sidebarContainer.appendChild(this.navigation);
-    this.glyphCellViewContainer.appendChild(this.glyphCellView);
+    glyphCellViewContainer.appendChild(this.glyphCellView);
 
     this.fontController.addChangeListener({ glyphMap: null }, () => {
       this._updateGlyphItemList();
@@ -341,17 +347,14 @@ export class FontOverviewController extends ViewController {
     const glyphSections = this.glyphOrganizer.groupGlyphs(glyphItemList);
     this.glyphCellView.setGlyphSections(glyphSections);
 
-    // Add placeholder if no glyphs are found
+    // Show placeholder if no glyphs are found
+    const fontOverviewNoGlyphsContainer = document.querySelector(
+      "#font-overview-no-glyphs"
+    );
     if (glyphSections.length === 0) {
-      if (!this.glyphCellViewContainer.querySelector(".font-overview-no-glyphs")) {
-        this.glyphCellViewContainer.appendChild(
-          html.div({ class: "font-overview-no-glyphs" }, [
-            translate("(No glyphs found)"), // TODO: translation
-          ])
-        );
-      }
+      fontOverviewNoGlyphsContainer.classList = "";
     } else {
-      this.glyphCellViewContainer.querySelector(".font-overview-no-glyphs")?.remove();
+      fontOverviewNoGlyphsContainer.classList = "hidden";
     }
   }
 
--- a/src/fontra/views/fontoverview/fontoverview.js
+++ b/src/fontra/views/fontoverview/fontoverview.js
@@ -167,7 +167,7 @@ export class FontOverviewController extends ViewController {
     await this.fontController.subscribeChanges(rootSubscriptionPattern, false);
 
     const sidebarContainer = document.querySelector("#sidebar-container");
-    const glyphCellViewContainer = document.querySelector("#glyph-cell-view-container");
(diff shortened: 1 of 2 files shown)


justvanrossum:
What alternative phrasings could we consider?

Turning off all glyph sets would have the same effect. A font that has no glyphs also.

Perhaps it is good as it is, but…

src/fontra/views/fontoverview/fontoverview.js
```diff
@@ -340,6 +340,19 @@ export class FontOverviewController extends ViewController {
     const glyphItemList = this.glyphOrganizer.filterGlyphs(combinedGlyphItemList);
     const glyphSections = this.glyphOrganizer.groupGlyphs(glyphItemList);
     this.glyphCellView.setGlyphSections(glyphSections);
+
+    // Add placeholder if no glyphs are found
+    if (glyphSections.length === 0) {
```
justvanrossum, line 345: Please look at the related glyphs panel like I told you. There is a callback for this.

ollimeier: I assume it would look like this:
Screenshot 2025-01-15 at 21 40 20

But this does not work. 
Within `setGlyphSections` it does not get to `this.onNoGlyphsToDisplay?.();` because `this.accordion.items.length` is 0, this is not the case for 'related glyphs'. So, there must be a difference between font-overview and related glyphs which I don't understand, currently.

ollimeier: When the search field in the font-overview is empty -> `this.accordion.items` is also empty. Which makes sense to me. This is different for 'related glyphs' it still has 5 items in `this.accordion.items`, even though it displays '(No related glyphs or characters were found)'.

justvanrossum: Ahhh yes, so we know for sure there are glyphs once there are accordion items, and no glyphs if there are no accorion items. And we don't have the situation where sections can turn out to be empty after the fact, as in the related glyph panel. So this situation is simpler, and your check for "are there items" is always enough. Apologies.

ollimeier: Thanks. 
I am not sure what to do now. Do we keep the PR as it is or do we modify other parts of the code?

justvanrossum: The logic is correct, right? I have some feedback about the details of the placeholder element, but otherwise this is good.

src/fontra/views/fontoverview/fontoverview.js
```diff
@@ -340,6 +340,19 @@ export class FontOverviewController extends ViewController {
...
+      if (!this.glyphCellViewContainer.querySelector(".font-overview-no-glyphs")) {
```
justvanrossum, line 346: How about we add this element at the same time as the glyphCellView, and just leave it there with empty content?
The we don't need `glyphCellViewContainer` as a property, and we'd add `this.noGlyphsElement` with fixed content, and set `display: none` in the CSS, and add a class if there are indeed no glyphs, which will make the display set to normal.

ollimeier: Please see the latest commit.

src/fontra/views/fontoverview/fontoverview.js
```diff
@@ -340,6 +346,16 @@ export class FontOverviewController extends ViewController {
     const glyphItemList = this.glyphOrganizer.filterGlyphs(combinedGlyphItemList);
     const glyphSections = this.glyphOrganizer.groupGlyphs(glyphItemList);
     this.glyphCellView.setGlyphSections(glyphSections);
+
+    // Show placeholder if no glyphs are found
+    const fontOverviewNoGlyphsContainer = document.querySelector(
+      "#font-overview-no-glyphs"
+    );
+    if (glyphSections.length === 0) {
```
justvanrossum, line 354: Simpler:
```js
    fontOverviewNoGlyphsContainer.classList.toggle("hidden", !glyphSections.length);
```

ollimeier: Thanks. I changed it, but `!glyphSections.length` was wrong -> without `!` is correct.

src/fontra/views/fontoverview/fontoverview.js
```diff
@@ -340,6 +346,12 @@ export class FontOverviewController extends ViewController {
     const glyphItemList = this.glyphOrganizer.filterGlyphs(combinedGlyphItemList);
     const glyphSections = this.glyphOrganizer.groupGlyphs(glyphItemList);
     this.glyphCellView.setGlyphSections(glyphSections);
+
+    // Show placeholder if no glyphs are found
+    const fontOverviewNoGlyphsContainer = document.querySelector(
+      "#font-overview-no-glyphs"
+    );
+    fontOverviewNoGlyphsContainer.classList.toggle("hidden", glyphSections.length);
```
justvanrossum, line 354: Am I missing something? I axpected this to be `!glyphSections.length`

ollimeier: By default is is hidden, so if there is a glyphs, it will remove it. Ah, maybe here is the misunderstanding. Maybe better show it by default and only hide it if there are glyphs. Let me change it.

justvanrossum: The argument to `.toggle()` is "on/off", so `.toggle("xxx", true)` will _add_ the class.

ollimeier: If `glyphSections.length` equal 0 -> `false`
and if `glyphSections.length` greater 0 -> `true`
Then add `hidden` if `true` which is `glyphSections.length` greater 0. 
So, it's correct. Hide it if there are glyphs and don't hide it if there are no glyphs.
Anyway, I tested it and it works like expected.

ollimeier: Proof:
https://github.com/user-attachments/assets/872bf567-7f03-4c18-ae3c-0aa2e722d20b

justvanrossum: You're absolutely right. It's late here... Let's make the condition explicit by adding `> 0`.

ollimeier: OK. Done.

PR #1955, opened 2025-01-16
Show subtools with right or control click

### LINKED ISSUE #1953: [editor] Subtools right-click / control-click should open the subtools
This currently opens the browser contextual menu, which is confusing.

The main tool should listen to "oncentext", and open the submenu.

ollimeier:
Fixes #1953

I first had the following code
```
        window.addEventListener("mousedown", globalListener, false);
        window.addEventListener("keydown", globalListener, false);
```
part of `showSubTools` to reduce duplicate code as much as possible, but I removed it again, because I believe it's more clear this way.

### DIFF
--- a/src/fontra/views/editor/editor.js
+++ b/src/fontra/views/editor/editor.js
@@ -980,6 +980,14 @@ export class EditorController extends ViewController {
 
     const editToolsElement = document.querySelector("#" + wrapperID);
 
+    const globalListener = {
+      handleEvent: (event) => {
+        if (event.type != "keydown" || event.key == "Escape") {
+          collapseSubTools(editToolsElement);
+        }
+      },
+    };
+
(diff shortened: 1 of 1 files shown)


justvanrossum:
Hm, this doesn't look so good: I don't think we can guarantee that the "global listener" is removed. The original code is already quite dodgy, and this adds to it.

Do you remember why we need this global listener in the first place?

ollimeier:
> Hm, this doesn't look so good: I don't think we can guarantee that the "global listener" is removed. The original code is already quite dodgy, and this adds to it.
> 
> Do you remember why we need this global listener in the first place?

We hide the subtools with the `globalListener`: So, if we don't focus the tools anymore, either by clicking somewhere else or if we hit the escape key.

justvanrossum:
Superseded by https://github.com/googlefonts/fontra/pull/1956

Issue #1952, opened 2025-01-16
Separate the (Python) server from the frontend in the source tree


justvanrossum:
This is with @simoncozens: to clean up the Fontra source tree, and clearly separate the Python server source code from the front end code and assets, with the purpose of opening up more deployment options for Fontra, such as a serverless version.

Possible breakage that we must keep in mind
- Fontra Pak
- Fontra RCJK deployment

Steps:

- [x] #1960
- [x] Move script imports and inline JS in HTML files into view-specific `start.js` files (#1984)
- [ ] Add scaffolding which does not interact with existing development (new `webpack.config.js`, package subdirectory structure, etc.) (#1985)
- [x] Rename `bundle-rollup` to `build` in package.json, make build hook similarly generically named (#1983)
- [x] Update Node requirement to >=20, in mention in README.md, set GH Actions workflow Node to 20.x (or latest?)
  - Make test-js/test-classes.js use "with", now that we require Node >=20
- [x] Update README.md:
  - Build instructions for deployment
  - Build/watch instructions for development cycle
- [x] Update .gitignore to ignore src/fontra/client/
- [x] A check that we have Node >=20, rather than confusing syntax error with npm test
- [x] Implement solution for fontra-rcjk's project manager (does it need its own little big-bang?)
- [x] Commit correct package-lock.json
- [x] Schedule a big-bang migration of JS files into packages plus server modifications

_(To be edited...)_

simoncozens:
OK, I think I am getting close to this. Everything is working nicely locally; you can test my [js-split](https://github.com/simoncozens/fontra/tree/js-split) branch.

To answer the remaining question about JS tracebacks, I've added two build modes - when Fontra is installed with `pip install .`, it runs `npm run bundle` which runs webpack in production mode. This gives you minified JS and tracebacks like `start.1a2b3c.js:17839`. When you run in development mode (`npm run bundle-watch`), then you get sourcemaps, sensible backtraces, and original JS code in the developer tools console. There are apparently [many options](https://webpack.js.org/configuration/devtool/) between those two extremes if we want to tweak things.

FontraPak works with a small change to the fontoverview URL.

I think I need to check that other projectmanager implementations also work.

simoncozens:
Oh... and the renaming of JavaScript imports is now scripted, so we can quickly recreate the "big-bang" commit whenever we decide to do it.

justvanrossum:
> the renaming of JavaScript imports is now scripted

[...]

PR #1959, opened 2025-01-16
Use ?project= instead of /-/

### LINKED ISSUE #1960: Rework URL scheme so the "project identifier" becomes a query key/value
The server should recognize the old pattern and do a redirect.

simoncozens:
This fixes #1960, part of #1952. Supports redirecting existing URLs. I've chosen `?project=` as the query string parameter to hold the project ID. Assumes that URLs requesting static assets do not have a query string attached.

### DIFF
--- a/src/fontra/core/server.py
+++ b/src/fontra/core/server.py
@@ -78,10 +78,7 @@ class FontraServer:
             )
         for viewName, viewPackage in self.viewEntryPoints.items():
             routes.append(
-                web.get(
-                    f"/{viewName}/-/{{path:.*}}",
-                    partial(self.viewPathHandler, viewName),
-                )
+                web.get(f"/{viewName}/-/{{path:.*}}", self.viewRedirectHandler)
             )
             routes.append(
                 web.get(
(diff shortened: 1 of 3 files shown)


src/fontra/core/server.py
```diff
@@ -280,8 +277,15 @@ async def viewPathHandler(
...
 
-        path = request.match_info["path"]
-        if not await self.projectManager.projectAvailable(path, authToken):
+        if not request.query:
+            return await self.staticContentHandler(
+                self.viewEntryPoints[viewName], request
+            )
+
+        project = request.query.get("project")
+        if project and not await self.projectManager.projectAvailable(
```
justvanrossum, line 286: Shouldn't this be
```suggestion
        if project is None or not await self.projectManager.projectAvailable(
```
?

justvanrossum (commented):
Looks good, thanks! I left one comment inline.

This would also need a change in Fontra Pak, but that conflicts with [this open PR](https://github.com/googlefonts/fontra-pak/pull/152/files). I'll take care of that.

Thanks to the redirect nothing breaks: excellent!

src/fontra/core/server.py
justvanrossum: Yes: I can confirm this changes the case of "we do have a query, but it doesn't have a `project` key" to a 404.

PR #1984, opened 2025-01-23
Move JS code from HTML to start.js

simoncozens:
This is another PR aimed at reducing the size of the #1952 change when it comes. It's good to have all the JS things in JS, not in HTML!


justvanrossum (approved):
Great, thanks!

Ideally, I think only the theme-settings imports should stay in the start.js files. I'll see what I can clean up after this.

PR #1985, opened 2025-01-24
JS package scaffolding

simoncozens:
This is another non-invasive change as part of #1952. It adds a bunch of files which don't do anything right now, but will magically become relevant once we move the JS sources out of the Python tree.

Draft for now, because I'm going to make another branch on top of this locally to confirm I can do the migration and it all works.

### DIFF
--- a/src-js/fontra-core/package.json
+++ b/src-js/fontra-core/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "@fontra/core",
+  "dependencies": {
+    "@fontra/core": "file:",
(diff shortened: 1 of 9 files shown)


justvanrossum:
After the "big changeover" view names no longer need to be valid Python identifiers, and I'd like to rename the views to:

- `application-settings`
- `font-info`
- `font-overview`

This will require some changes that aren't strictly related to the changeover, so I can easily do this afterwards. This is just a FYI, just in case you see'd some benefit in doing this along the way.

PR #1988, opened 2025-01-24
Turn off edit mode if the selected glyph does not exist

### LINKED ISSUE #1987: [related glyphs] Double-clicking a non-existing related glyph gets inserted in canvas in edit mode
Non-existing glyphs should are not allowed to be in edit mode, but this isn't properly enforced.

justvanrossum:
Fix the insertion code in EditorController to turn off edit mode when inserting a non-existing glyph.

This fixes #1987.


justvanrossum:
Hm, this didn't properly work. Fixed by https://github.com/googlefonts/fontra/commit/51f9ff31f4279750c0fd18f8959621987094c72c

PR #1993, opened 2025-01-28
Fix sorting of Koeberlin's glyph sets

ollimeier:


justvanrossum:
As discussed: this file is generated by `scripts/rebuild_glyphset_presets.py`, and the sorting should be improved there.

PR #1994, opened 2025-01-28
Fix sorting of koeberlins glyphsets

ollimeier:
Fix sorting in `rebuild_glyphset_presets.py` and write to file (remove print) + update data

### DIFF
--- a/scripts/rebuild_glyphset_presets.py
+++ b/scripts/rebuild_glyphset_presets.py
@@ -161,10 +161,14 @@ def getKoeberlinLatinGlyphSets():
             }
         )
 
-    order = ["S", "M", "L", "XL"]
+    order = ["XS", "S", "M", "L", "XL", "XXL"]
     glyphSets.sort(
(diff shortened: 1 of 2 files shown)


scripts/rebuild_glyphset_presets.py
```diff
@@ -160,6 +161,18 @@ def getKoeberlinLatinGlyphSets():
             }
         )
 
+    order = ["XS", "S", "M", "L", "XL", "XXL"]
```
justvanrossum, line 164:
```python
order = {k: i for i, k in enumerate(["XS", "S", "M", "L", "XL", "XXL"])}
```
Then use `order.get(glyphSet["name"].split()[-1], len(order))` (not `len(order) - 1`)

This avoids the ugly and unnecessary duplication of `glyphSet["name"].split()[-1]`.

ollimeier: Thanks. Done.

PR #1995, opened 2025-01-28
Peel off glyph-cell-view displayMode option

giovanniTramonto:
CSS only solution to peel off `displayMode` option from `glyph-cell-view`


justvanrossum:
Thanks! This looks good and works well.

PR #2007, opened 2025-02-04
Rework font sources UI

### LINKED ISSUE #1997: Rework font sources UI
Remove cards and move to columns.

Image

ollimeier:
Fixes #1997

- I use "add-remove-buttons" (similar to other areas of Fontra) to add or remove a font source, this makes the repeating trash-icon redundant
- When adding a new font source, this will be 'selected'.
- With arrow up and down you can go to previous or next source (this will be ignored if the focus is for example on an input)

### DIFF
--- a/src/fontra/views/fontinfo/fontinfo.js
+++ b/src/fontra/views/fontinfo/fontinfo.js
@@ -112,16 +112,12 @@ export class FontInfoController extends ViewController {
 
   getUndoRedoLabel(isRedo) {
     const panel = this.panels[this.selectedPanel];
-    const info = panel?.undoStack?.getTopUndoRedoRecord(isRedo)?.info;
-    return (
-      (isRedo ? translate("action.redo") : translate("action.undo")) +
-      (info ? " " + info.label : "")
-    );
+    return panel.getUndoRedoLabel(isRedo);
   }
 
   canUndoRedo(isRedo) {
     const panel = this.panels[this.selectedPanel];
-    return panel?.undoStack?.getTopUndoRedoRecord(isRedo)?.info;
+    return panel.canUndoRedo(isRedo);
   }
 
   doUndoRedo(isRedo) {
--- a/src/fontra/views/fontinfo/fontinfo.js
+++ b/src/fontra/views/fontinfo/fontinfo.js
@@ -1,3 +1,4 @@
+import { registerActionCallbacks } from "../core/actions.js";
 import * as html from "../core/html-utils.js";
 import { AxesPanel } from "./panel-axes.js";
 import { CrossAxisMappingPanel } from "./panel-cross-axis-mapping.js";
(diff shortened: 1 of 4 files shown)


ollimeier:
Short demo:

https://github.com/user-attachments/assets/f3ff8c63-ca63-4f59-b7a6-aee898c379cc

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -21,7 +27,30 @@ import "/web-components/add-remove-buttons.js";
...
 import { dialogSetup, message } from "/web-components/modal-dialog.js";
 
-const cardsInfos = {};
+let selectedSourceIdentifier = undefined;
+
+addStyleSheet(`
+.font-sources-container {
+  display: grid;
+  grid-template-columns: auto 1fr;
+  height: calc(100vh - var(--top-bar-height));
```
justvanrossum, line 36: Why is this needed? The total page already has a calculation for the top bar. A panel like this should not have to worry.

justvanrossum (commented):
It mostly works well. If I use the arrow keys, and the list of sources does not fit, the selection can get out of view. You should use scrollIntoView.

I have not done a thorough code review yet, but left one comment.

justvanrossum (commented):
Visual feedback: I would like the elements of the source panel to be more left aligned, instead of floating in the middle.

src/fontra/views/fontinfo/panel-sources.js
ollimeier: This is required to make the two elements (source names, source content) scrollable separately. Without it the the whole container scrolls, but not each element separately.

justvanrossum: There must be something wrong elsewhere then, because the container for this container should already have the correct size.

ollimeier: We fixed this together. I am closing this as resolved.

src/fontra/views/fontinfo/fontinfo.js
```diff
@@ -86,6 +93,41 @@ export class FontInfoController extends ViewController {
...
+
+    registerActionCallbacks(
+      "action.redo",
+      () => this.doUndoRedo(true),
+      () => this.canUndoRedo(true),
+      () => this.getUndoRedoLabel(true)
+    );
+  }
+
+  getUndoRedoLabel(isRedo) {
```
justvanrossum, line 113: It would be nicer to delegate getUndoLabel _and_ canUndoRedo to the panel itself. The panel base class can implement them, just like doUndoRedo.

ollimeier: Thanks for the feedback. Done.

justvanrossum (approved):

Issue #2005, opened 2025-02-01
Allow glyph sets to define a custom string for placeholders


khaledhosny:
Some glyph sets contain many glyphs without code points (for example, in scripts like Arabic, most of the glyph sets has no code points), and currently the placeholders for such glyphs is an empty cell which makes it hard to navigate.

It is sometimes possible to have a usable placeholder using a string without necessarily wanting to set a code point for the glyph.

For example, in GF Arabic Plus, `jeh-ar.fina` does not have a code point, but for placeholder one can use the (legacy) Arabic Presentation Form character U+FB8B ARABIC LETTER JEH FINAL FORM. On the other hand, `rehFourdots-ar.fina` has no code point or even a Presentation Forms code point, but final form can be forced by using ZWJ + the base glyph code point i.e. the string U+200D U+0699.

I think allowing for an optional column for the placeholder string in the CSV format should allow having usable placeholders for many glyphs without code points.

khaledhosny:
I tried working on this, but it seems to require too many changes, so instead I tried something simpler: when getting the placeholder character, if the glyph has no code point and the name has ".init", ".medi", etc. extension, use get the code point for the base glyph name and ZWNJ to force the desired positional form. This is less flexible but the required changes are isolated and seems to work well enough for my use case.

I can open a PR if a change like this can be considered.

khaledhosny:
Current:
<img width="1624" alt="Image" src="https://github.com/user-attachments/assets/a06b233c-19e8-40b4-a528-95080938bf7c" />

After proposed change:
<img width="1624" alt="Image" src="https://github.com/user-attachments/assets/cc6404f2-bf2c-4035-a5ff-a682c1a13070" />

justvanrossum:
(There also is the concept of "associatedCodePoints", which is used to guess the script of unencoded glyphs. Not sure how useful that is for this use case.)

I like the idea of having an optional third column to explicitly specify a placeholder character (or char sequence?).

I am also not opposed to do some guessing based on glyph name extension, for well established common cases like these, so I'm curious about your changes.

justvanrossum:
Did https://github.com/googlefonts/fontra/pull/2010 resolve this for you?

khaledhosny:
Yes, that covers my needs.

PR #2010, opened 2025-02-07
Guess glyph string for placeholder glyphs without code points

khaledhosny:
See the discussion in #2005

This works in the font overview, but I didn’t manage to find the relevant code in editor, so editing such glyphs still does not shows blank placeholder.

### DIFF
--- a/src-js/fontra-core/src/glyph-data.js
+++ b/src-js/fontra-core/src/glyph-data.js
@@ -119,6 +119,8 @@ export function getGlyphInfoFromGlyphName(glyphName) {
 
 export function guessGlyphPlaceholderString(codePoints, glyphName) {
   let glyphString = "";
+  let direction = undefined;
+
   if (codePoints?.[0]) {
     glyphString = getCharFromCodePoint(codePoints[0]);
   }
@@ -145,12 +147,15 @@ export function guessGlyphPlaceholderString(codePoints, glyphName) {
             break;
           case ".init":
             glyphString = glyphString + ZWJ;
+            direction = "rtl";
             break;
           case ".medi":
             glyphString = ZWJ + glyphString + ZWJ;
+            direction = "rtl";
             break;
           case ".fina":
             glyphString = ZWJ + glyphString;
+            direction = "rtl";
             break;
           default:
             break;
@@ -159,5 +164,5 @@ export function guessGlyphPlaceholderString(codePoints, glyphName) {
     }
   }
 
-  return glyphString;
+  return [glyphString, direction];
 }
--- a/src-js/fontra-core/src/glyph-data.js
+++ b/src-js/fontra-core/src/glyph-data.js
@@ -134,7 +134,9 @@ export function guessGlyphPlaceholderString(codePoints, glyphName) {
       baseGlyphNames = base.split("_").map((name) => name + suffix);
     }
 
-    const codePoints = baseGlyphNames.map((name) => getCodePointFromGlyphName(name));
+    const codePoints = baseGlyphNames
+      .map((name) => getCodePointFromGlyphName(name))
+      .filter((item) => item);
     if (codePoints.length == baseGlyphNames.length) {
       glyphString = codePoints
         .map((codePoint) => getCharFromCodePoint(codePoint))
--- a/src-js/fontra-core/src/glyph-data.js
+++ b/src-js/fontra-core/src/glyph-data.js
@@ -125,23 +125,32 @@ export function guessGlyphPlaceholderString(codePoints, glyphName) {
 
   if (!glyphString && glyphName) {
     const [baseGlyphName, extension] = splitGlyphNameExtension(glyphName);
-    if (extension) {
-      const codePoint = getCodePointFromGlyphName(baseGlyphName);
-      if (codePoint) {
-        const baseChar = getCharFromCodePoint(codePoint);
+
+    let baseGlyphNames = [baseGlyphName];
+    if (baseGlyphName.indexOf("_") != -1) {
+      const [base, suffix] = splitGlyphNameExtension(baseGlyphName, "-");
+      baseGlyphNames = base.split("_").map((name) => name + suffix);
+    }
+
+    const codePoints = baseGlyphNames.map((name) => getCodePointFromGlyphName(name));
+    if (codePoints.length == baseGlyphNames.length) {
+      glyphString = codePoints
+        .map((codePoint) => getCharFromCodePoint(codePoint))
+        .join("");
+
+      if (extension) {
         const ZWJ = "\u200D";
         switch (extension) {
           case ".isol":
-            glyphString = baseChar;
             break;
           case ".init":
-            glyphString = baseChar + ZWJ;
+            glyphString = glyphString + ZWJ;
             break;
           case ".medi":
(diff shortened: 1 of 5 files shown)


justvanrossum:
Thank you very much!

I wonder if the same can be achieved with the "associatedCodePoints" feature. This does roughly the same (including the "_" split), except it also takes the font's cmap/glyphMap into account, and only falls back to glyphs-data if the font doesn't define the base glyph name.

I think we can add `associatedCodePoints` as an (optional?) argument to the GlyphCell constructor. 

Alternatively, we could consider decoupling the making of the placeholder string from the GlyphCell completely, and pass the placeholder string as an additional argument to the GlyphCell constructor.

As for the placeholder string in the canvas: let's indeed focus on the GlyphCell first, and I'll try to figure out what we need to do to make it work in the canvas, too.

khaledhosny:
> I wonder if the same can be achieved with the "associatedCodePoints" feature.

The main difference between `associatedCodePoints` and this placeholder string is that I insert ZWJ before and/or after the code point to force an Arabic positional form.

Take for example the four glyphs `beh-ar`, `beh-ar.init`, `beh-ar.medi`, and `beh-ar.fina`. The `associatedCodePoints` for the last three would be the code point for `beh-ar`, so using this would show the same fallback for all the four glyphs. Instead, I insert ZWJ after the code point in `.init` to force initial form, before the code point in `.fina` to force final form, and before and after the code point in `.medi` to force the medial form.

Is there is a way to achieve this with `associatedCodePoints`?

justvanrossum:
(Thanks for keeping this up to date, I'll try to continue reviewing this soon. I am eager to land this feature.)

src-js/fontra-core/src/utils.js
```diff
@@ -486,8 +486,8 @@ export function* iter(iterable) {
   }
 }
 
-export function splitGlyphNameExtension(glyphName) {
-  const periodIndex = glyphName.indexOf(".");
+export function splitGlyphNameExtension(glyphName, separator = ".") {
+  const periodIndex = glyphName.indexOf(separator);
```
justvanrossum, line 490: Maybe call this variable `separatorIndex`.

src-js/fontra-core/src/glyph-data.js
```diff
@@ -111,3 +116,48 @@ export function getGlyphInfoFromGlyphName(glyphName) {
...
+    const [baseGlyphName, extension] = splitGlyphNameExtension(glyphName);
+
+    let baseGlyphNames = [baseGlyphName];
+    if (baseGlyphName.indexOf("_") != -1) {
+      const [base, suffix] = splitGlyphNameExtension(baseGlyphName, "-");
+      baseGlyphNames = base.split("_").map((name) => name + suffix);
+    }
+
+    const codePoints = baseGlyphNames.map((name) => getCodePointFromGlyphName(name));
+    if (codePoints.length == baseGlyphNames.length) {
```
justvanrossum, line 136: Isn't this condition always true? Or do you mean to filter out null/undefined from the `codePoints` array?

src-js/views-editor/src/visualization-layer-definitions.js
```diff
@@ -145,9 +146,16 @@ registerVisualizationLayerDefinition({
...
         -lineDistance * glyphNameFontSize
       );
+    }
+    const placeholderString = guessGlyphPlaceholderString(
+      positionedGlyph.character?.codePointAt(0),
+      positionedGlyph.glyphName
+    );
+    if (placeholderString) {
       context.font = `${placeholderFontSize}px fontra-ui-regular, sans-serif`;
+      context.direction = "rtl"; // FIXME "auto" does not work here
```
justvanrossum, line 156: `guessGlyphPlaceholderString()` could return a guessed direction, using the (optional) `direction` field of the `getGlyphInfoFromCodePoint(codePoint)` result.

```js
    const { placeholderString, direction } guessGlyphPlaceholderString(...);
```

justvanrossum (commented):
This is looking very good, thank you. I left some minor notes, and a suggestion how to maybe fix the `context.direction = "rtl";` line.

src-js/views-editor/src/visualization-layer-definitions.js
khaledhosny: Done

justvanrossum: I meant to really use the "direction" property of the glyph info, as returned by `getGlyphInfoFromCodePoint()` to guess the writing direction of the placeholder string. I'm not quite comfortable with the assumption that `.init`, `.medi` and `.fina` are only used for RTL fonts.

Additionally, I'd really like the return value to be an object with named field, not a list with two items. This is more idiomatic for JS, and allows us to add more items to the return value. (I'd like to take this idea further later, and return OpenType feature tags as well — thinking of tags like `scmp`, `subs`, `onum`, `dnum`.)

src-js/fontra-core/src/glyph-data.js
```diff
@@ -111,3 +116,55 @@ export function getGlyphInfoFromGlyphName(glyphName) {
...
+            direction = "rtl";
+            break;
+          default:
+            break;
+        }
+      }
+    }
+  }
+
+  return [glyphString, direction];
```
justvanrossum, line 169:
```suggestion
  return { glyphString, direction };
```
As mentioned in https://github.com/googlefonts/fontra/pull/2010#discussion_r2038957884

src-js/fontra-webcomponents/src/glyph-cell.js
```diff
@@ -223,13 +220,16 @@ export class GlyphCell extends UnlitElement {
             : html.div(
                 {
                   class: "glyph-shape-placeholder",
+                  ...(this._placeholderDirection
```
justvanrossum, line 223: Suggest to simplify to 
```js
    dir: this._placeholderDirect || "auto",
```
Or, make `this._placeholderDirection` default to `auto` and assign to `dir` unconditionally.

src-js/views-editor/src/visualization-layer-definitions.js
justvanrossum: Your PR also gives placeholders for glyphs like this:

image

I think that's really nice. But would be even better if the "right" OT feature(s) were applied. Something for later. Also: localized variants?

khaledhosny: I thought I’d get away with this as I didn’t want to think about the complication of having glyphs with different directions, but that probably is not something that would normally happen (it wouldn’t make sense as we can’t have ligatures across scripts).

The features idea is really nice. It might depend on #2075 to be fully functional (currently there is no guarantee that the font used for fallback strings has any of these features), but at least for mandatory features like `init`, `fina`, etc. it should actually work and I wouldn’t need to use ZWJ and the whole direction business. So I think I’m going to try it. Except that canvas context does not have a way to apply font features 😢

justvanrossum: Arg, canvas is indeed missing a bunch of features. For the reference font I actually set the language on the html canvas element. Perhaps CSS on the canvas element also trickles down to drawing, but we need it to be set per string we draw, hmmhmmm.

I think I'd prefer this PR to just set the direction explicitly, and applying features should be a followup project.

PR #2016, opened 2025-02-10
Remove redundant css from panel-transforamtion

giovanniTramonto:
styles are not in use anymore since https://github.com/googlefonts/fontra/commit/804117b4af67c8bfbf2c2f57c7c7dbeec9487f80


ollimeier (approved):
I can confirm, that the removed classes are not anymore in use. I did also a quick test and the panel works as expected.

PR #2017, opened 2025-02-10
Add reference-font remove button

giovanniTramonto:
Before reference-font items could only be deleted by selecting (clicking) an item and pressing delete-key after. Not self-explanatory. 
Now there is a remove (-) button at the end of each line.

Bildschirmfoto 2025-02-10 um 16 01 28

### DIFF
--- a/src/fontra/views/editor/panel-reference-font.js
+++ b/src/fontra/views/editor/panel-reference-font.js
@@ -15,7 +15,7 @@ import {
 import { ObservableController } from "/core/observable-object.js";
 import { getOPFS } from "/core/opfs.js";
 import { fetchJSON, fileNameExtension, modulo, withTimeout } from "/core/utils.js";
-import { dialog, dialogSetup, message } from "/web-components/modal-dialog.js";
+import { dialog, message } from "/web-components/modal-dialog.js";
 import "/web-components/range-slider.js";
 import { UIList } from "/web-components/ui-list.js";
 
@@ -431,6 +431,8 @@ export default class ReferenceFontPanel extends Panel {
   }
 
   async _filesHandler(files) {
+    if (files.length === 0) return;
+
     const fontItemsInvalid = [];
     const fontItems = [...files]
       .filter((file) => {
@@ -660,41 +662,20 @@ export default class ReferenceFontPanel extends Panel {
       []
     );
 
+    const fileInput = input(
+      {
+        type: "file",
+        accept: ".ttf,.otf,.woff,.woff2",
+        multiple: true,
+      },
+      []
+    );
+    fileInput.addEventListener("change", async () => {
+      await this._filesHandler(fileInput.files);
+      fileInput.value = null;
+    });
     const addRemoveButtons = createDomElement("add-remove-buttons");
-    addRemoveButtons.addButtonCallback = async () => {
-      // TODO: translation
-      const dialog = await dialogSetup("Add Reference Font(s)", null, [
-        {
-          title: translate("dialog.cancel"),
-          resultValue: "cancel",
-          isCancelButton: true,
-        },
-        {
-          title: translate("dialog.add"),
-          resultValue: "ok",
-          isDefaultButton: true,
-        },
-      ]);
-      const fileInput = input(
-        {
-          type: "file",
-          accept: ".ttf,.otf,.woff,.woff2",
-          multiple: true,
-        },
-        []
-      );
-      dialog.setContent(
-        form({ enctype: "multipart/form-data", class: "content" }, [
-          label({}, [fileInput]),
-        ])
-      );
-      const result = await dialog.run();
-      const { files } = fileInput;
-      if (result === "ok" && files.length > 0) {
-        await this._filesHandler(files);
-      }
-    };
-
+    addRemoveButtons.addButtonCallback = () => fileInput.click();
     addRemoveButtons.removeButtonCallback = async () => {
       const index = this.filesUIList.getSelectedItemIndex();
       const maxIndex = () => this.filesUIList.items.length - 1;
@@ -725,6 +706,7 @@ export default class ReferenceFontPanel extends Panel {
             div({}, [translate("sidebar.reference-font.info")]),
             div({ class: "reference-font-items" }, [
               this.filesUIList,
+              form({ enctype: "multipart/form-data", hidden: true }, [fileInput]),
               addRemoveButtons,
             ]),
             div(
--- a/src/fontra/views/editor/panel-reference-font.js
+++ b/src/fontra/views/editor/panel-reference-font.js
@@ -431,7 +431,9 @@ export default class ReferenceFontPanel extends Panel {
   }
 
   async _filesHandler(files) {
-    if (files.length === 0) return;
+    if (files.length === 0) {
+      return;
+    }
 
     const fontItemsInvalid = [];
     const fontItems = [...files]
@@ -544,10 +546,7 @@ export default class ReferenceFontPanel extends Panel {
   }
 
   async _deleteSelectedItemHandler() {
-    const index = this.filesUIList.getSelectedItemIndex();
-    if (index > -1 && index <= this.filesUIList.items.length - 1) {
-      await this._deleteItemOrAll(index);
-    }
+    await this._deleteItemOrAll(this.filesUIList.getSelectedItemIndex());
   }
 
   async _deleteAllHandler() {
@@ -670,7 +669,7 @@ export default class ReferenceFontPanel extends Panel {
     const fileInput = input(
       {
         type: "file",
-        accept: ".ttf,.otf,.woff,.woff2",
+        accept: [...fontFileExtensions].map((extension) => `.${extension}`).join(","),
         multiple: true,
       },
       []
--- a/src/fontra/views/editor/panel-reference-font.js
+++ b/src/fontra/views/editor/panel-reference-font.js
@@ -562,7 +562,7 @@ export default class ReferenceFontPanel extends Panel {
     model.selectedFontIndex = -1;
     model.referenceFontName = "";
 
-    filesUIList.setItems(listItems);
+    filesUIList.setItems(listItems, false, true);
     if (!filesUIList.getSelectedItemIndex() && listItems.length > 0) {
       filesUIList.setSelectedItemIndex(index === 0 ? 0 : index - 1, true);
     }
--- a/src/fontra/views/editor/panel-reference-font.js
+++ b/src/fontra/views/editor/panel-reference-font.js
@@ -463,10 +463,9 @@ export default class ReferenceFontPanel extends Panel {
       dialog(dialogTitle, dialogMessage, [{ title: translate("dialog.okay") }], 5000);
     }
 
-    const newSelectedItemIndex = this.filesUIList.items.length;
     const newItems = [...this.filesUIList.items, ...fontItems];
     this.filesUIList.setItems(newItems);
-    this.filesUIList.setSelectedItemIndex(newSelectedItemIndex, true);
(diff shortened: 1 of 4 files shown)


justvanrossum:
We use at least two different things for this in Fontra:
- (+) (-) buttons _below_ the list (for example in glyph source list)
- a "trash" icon at the right end side of a panel/row/thing

In the context of lists, so far we only use the (+) (-) approach. So that would be more consistent with other parts of Fontra. (The "+" button could trigger a file upload dialog)

On the other hand, to have a delete icon ("trash" would be preferred then), is a bit more user friendly, as it doesn't require an item to be selected first.

I don't think this icon needs to be so deeply integrated into the ui-list class. Look at for exampkle the glyph source list: it contains clickable icons.

I'm leaning towards to (+) (-) solution.

giovanniTramonto:
Ah yes, I see, that’s much easier, actually. Thanks for the hint

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -649,6 +661,44 @@ export default class ReferenceFontPanel extends Panel {
       []
     );
 
+    const addRemoveButtons = createDomElement("add-remove-buttons");
+    addRemoveButtons.addButtonCallback = async () => {
+      // FIXME Add dialog title translation
```
ollimeier, line 666: When ever we don't have a translation, we add `// TODO: translation` to the code. This way, we can find it easily one day.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -649,6 +661,44 @@ export default class ReferenceFontPanel extends Panel {
...
+          title: translate("dialog.add"),
+          resultValue: "ok",
+          isDefaultButton: true,
+        },
+      ]);
+      const fileInput = input({ type: "file", id: "reference-font-file" }, []);
+      dialog.setContent(
+        form({ enctype: "multipart/form-data", class: "content" }, [
+          label({ for: "reference-font-file", class: "reference-font-label" }, [
+            "Choose a file",
```
ollimeier, line 683: Please add `// TODO: translation` + I am not sure about the wording. Maybe better 'Choose a font', because we don't want any file type to be uploaded.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -649,6 +661,44 @@ export default class ReferenceFontPanel extends Panel {
...
+          resultValue: "cancel",
+          isCancelButton: true,
+        },
+        {
+          title: translate("dialog.add"),
+          resultValue: "ok",
+          isDefaultButton: true,
+        },
+      ]);
+      const fileInput = input({ type: "file", id: "reference-font-file" }, []);
```
ollimeier, line 679:
```suggestion
      const fileInput = input({ type: "file", id: "reference-font-file", accept: "font/*"}, []);
```

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -649,6 +661,44 @@ export default class ReferenceFontPanel extends Panel {
...
+      dialog.setContent(
+        form({ enctype: "multipart/form-data", class: "content" }, [
+          label({ for: "reference-font-file", class: "reference-font-label" }, [
+            "Choose a file",
+          ]),
+          fileInput,
+        ])
+      );
+      const result = await dialog.run();
+      const { files } = fileInput;
```
ollimeier, line 689: Dropping multiple files do work, but selecting multiple files don't work with the dialog, currently. You may want to add `multiple: true` to the input, but please double check if it works or not. I am not 100% sure about this.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -649,6 +661,44 @@ export default class ReferenceFontPanel extends Panel {
...
+      }
+    };
+
+    addRemoveButtons.removeButtonCallback = () => {
+      const index = this.filesUIList.getSelectedItemIndex();
+      if (!isNaN(index) && index <= this.filesUIList.items.length - 1) {
+        this._deleteItemOrAll(index);
+      }
+    };
+
```
ollimeier, line 701: If there is no reference font, please disable the remove button via `disableRemoveButton`.

ollimeier (changes requested):
Hi @giovanniTramonto (Johannes), thanks a lot that you help us with Fontra. I much appreciate that and it's great to be in contact with you this way, again. 
I tested it and it works well. Please see my comments.

src/fontra/views/editor/panel-reference-font.js
justvanrossum: Is this enough to make it accept ttf, otf, woff and woff2 and nothing else?

giovanniTramonto:
Hi @ollimeier! 
Thanks for the review. Yes, indeed, great to work with you guys on this. Appreciate it very much, too.
I have implemented all suggestions as well as some improvement on the selection: Now, it automatically selects last row after removal, so it’s easier to remove multiple items as well.
I spot that the supported languages do not update when the list becomes empty. If it’s OK with you I would adjust this in a follow-up.

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -649,6 +660,60 @@ export default class ReferenceFontPanel extends Panel {
...
+          id: "reference-font-file",
+          accept: ".ttf,.otf,.woff,.woff2",
+          multiple: true,
+        },
+        []
+      );
+      dialog.setContent(
+        form({ enctype: "multipart/form-data", class: "content" }, [
+          label({ for: "reference-font-file", class: "reference-font-label" }, [
+            "Choose a font", // TODO: translation
```
ollimeier, line 690: I am wondering if we really need this "Choose a font" or if we can get rid of it. If I take a look at the dialog, It says "Choose a font" then "Choose file" and then "No file chosen". So, three time the same information, but slightly different. So, maybe just get rid of "Choose a font".

Screenshot 2025-02-13 at 09 00 54

ollimeier (dismissed):
Thanks a lot for your update. Seems to work well. I have only a few minor change requests related to the text strings in the dialog.
I noticed that the reference font is not displayed if we add only one font. Here a quick demo:

https://github.com/user-attachments/assets/21b98885-7748-40b1-9be0-fe4d94fbc548

This was not introduced by you. It was there already before. We may want to have a look into that issue as a follow up.

src/fontra/views/editor/panel-reference-font.js
giovanniTramonto: Yes, thats better!

giovanniTramonto:
> Thanks a lot for your update. Seems to work well. I have only a few minor change requests related to the text strings in the dialog. I noticed that the reference font is not displayed if we add only one font. Here a quick demo:
> 
>  Screen.Recording.2025-02-13.at.09.06.43.mp4 
> This was not introduced by you. It was there already before. We may want to have a look into that issue as a follow up.

Oh, interesting! Just tried to reproduce. For me it seems to work:

https://github.com/user-attachments/assets/105621f5-f57a-4014-af3a-3bcbe7540383

justvanrossum:
Is it possible for the "+" button to immediately do to the choose files dialog, without the in-between dialog?

giovanniTramonto:
Oh, yes. This is much more practical

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -452,10 +465,9 @@ export default class ReferenceFontPanel extends Panel {
       dialog(dialogTitle, dialogMessage, [{ title: translate("dialog.okay") }], 5000);
     }
 
-    const newSelectedItemIndex = this.filesUIList.items.length;
     const newItems = [...this.filesUIList.items, ...fontItems];
     this.filesUIList.setItems(newItems);
-    this.filesUIList.setSelectedItemIndex(newSelectedItemIndex, true);
+    this.filesUIList.setSelectedItemIndex(this.filesUIList.items.length - 1, true);
```
justvanrossum, line 470: Any particular reason for this change?

I think the idea was to select the _first_ item of a multiple-font drop. This would change it to the _last_ item.

justvanrossum (commented):
I left one comment, otherwise looks good to me.

src/fontra/views/editor/panel-reference-font.js
giovanniTramonto: Yes, true. I changed it that way, because the removal of items starts from the end. So let’s say you add a stack of fonts and want to remove some of them after, it seemed easier this way. Let me know what you think

justvanrossum: But this code is for adding fonts, no? I don't immediately see how this code is involved when deleting. I prefer to keep the old behavior when adding fonts.

Hm, the behavior between "hit delete key" and "click minus button" is now different. It should be factored to be the same.

I don't like the "select last item after I removed an arbitrary font". I think it should be the item after the deleted item, or the the one before, if deleted item was the last item. So the selected item index stays the same, or becomes length - 1.

justvanrossum (commented):
Thanks.

Should the remove button perhaps also respond to the alt key, to mean "delete all"?

So currently no item gets selected after removing the selection, right? This matches the original behavior, but not wat we recently discussed. Is that your intention indeed?

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -419,7 +430,9 @@ export default class ReferenceFontPanel extends Panel {
     return this.controller.model;
   }
 
-  async _filesDropHandler(files) {
+  async _filesHandler(files) {
+    if (files.length === 0) return;
```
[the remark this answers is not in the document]
giovanniTramonto: Alright

src/fontra/views/editor/panel-reference-font.js
```diff
@@ -531,7 +544,10 @@ export default class ReferenceFontPanel extends Panel {
   }
 
   async _deleteSelectedItemHandler() {
-    await this._deleteItemOrAll(this.filesUIList.getSelectedItemIndex());
+    const index = this.filesUIList.getSelectedItemIndex();
```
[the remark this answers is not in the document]
giovanniTramonto: Ah, I see, so the first item of the added fonts is selected, right after the existing items. Makes sense

giovanniTramonto:
> Thanks.
> 
> Should the remove button perhaps also respond to the alt key, to mean "delete all"?
> 
> So currently no item gets selected after removing the selection, right? This matches the original behavior, but not wat we recently discussed. Is that your intention indeed?

Now, it selects the item before in the list, or after in case its the first list item. Feels good

giovanniTramonto:
For the deleting multiple files, I guess, really nice would be to select a range with shift key.

But I have to solve this render glitch before:


https://github.com/user-attachments/assets/9f94f8f5-4dcf-49cb-9aaa-13b2c4326be5

justvanrossum:
> For the deleting multiple files, I guess, really nice would be to select a range with shift key.

Yes, but the UIList class currently does not support that. While we need that eventually, it is out of scope for this PR.

> But I have to solve this render glitch before:

Is it related to the file names being wrapped?

justvanrossum:
I'm willing to merge this as it is: it works well and the code looks fine to me. What do you say?

giovanniTramonto:
Glitch is fixed. It was related to the render update of `ui-list` which caused update of parent grid. Moving the list inside a wrapper helped. I also set `keepScrollPosition` to `true`, so when having many items after removal user can still see the selected item without scrolling down again. If that’s all fine with you, I’m willing to merge, too.

justvanrossum:
But now the layout behavior is different: the preview section at the bottom takes up more space than it needs when the size is small, leaving a lot of space unused:

image

PR #2021, opened 2025-02-12
Add support for font guidelines

### LINKED ISSUE #1963: Please add support for Global guidelines
I can add glyphs specific but not a global guidelines which is essential for designing Indic fonts which must align along a headline. 
Please add support for global guidelines

ollimeier:
Fixes #1963 #909 #2026

This allows adding/editing of font level guidelines via the sources panel. 
Currently it's not possible to edit font level guidelines within the editor view (we may want to create a new issue for this and see it as a follow up).

Quick demo:

https://github.com/user-attachments/assets/ad122b3b-ce95-43d1-98de-809658f33e2d

### DIFF
--- a/src/fontra/client/core/font-controller.js
+++ b/src/fontra/client/core/font-controller.js
@@ -20,6 +20,7 @@ import {
   colorizeImage,
   getCharFromCodePoint,
   mapObjectValues,
+  normalizeGuidelines,
   sleepAsync,
   throttleCalls,
   uniqueID,
@@ -1140,6 +1141,7 @@ function ensureDenseSources(sources) {
           return { value: metric.value, zone: metric.zone || 0 };
         }
       ),
+      guidelines: normalizeGuidelines(source.guidelines || []),
       customData: source.customData || {},
     };
   });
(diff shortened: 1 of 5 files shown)


justvanrossum (commented):
I noticed that the guideline label alignment is not consistent. This is technically unrelated to this PR, and has to do with the (use of) getTextHeight(): The text height is the total distance between (box) descender and (box) ascender, but I think the drawing is relative to the baseline. So labels containing _descender_ letters are poorly aligned:
image

justvanrossum:
> So labels containing descender letters are poorly aligned

Instead of getTextHeight(), we should consider a function getVerticalCenter(), that returns `(metrics.actualBoundingBoxAscent - metrics.actualBoundingBoxDescent) / 2` (given that `actualBoundingBoxDescent` is a positive value for a distance _below_ the baseline)

ollimeier:
> > So labels containing descender letters are poorly aligned
> 
> Instead of getTextHeight(), we should consider a function getVerticalCenter(), that returns `(metrics.actualBoundingBoxAscent - metrics.actualBoundingBoxDescent) / 2` (given that `actualBoundingBoxDescent` is a positive value for a distance _below_ the baseline)

OK, I have reworked it in this PR. Done.

ollimeier:
@justvanrossum I am done with this and I hope we can be merged soon. 
I have not changed the icon to a lock, because it does to feel right. It is technically wrong and also just a temporary thing as we want to be able to edit the font guidelines in the editor view. But if you have a different opinion on that, I am happy to change it. Just let me know. It's easy implementing it.

PR #2024, opened 2025-02-12
Don't show 'Transform selection' if only one point is selected

### LINKED ISSUE #2022: [Transform selection] Should not be visible, if only one off-curve handle is selected
If there is only one off-curve handle selected, the Transform selection box should not be visible.

ollimeier:
Fixes #2022

### DIFF
--- a/src/fontra/views/editor/edit-tools-pointer.js
+++ b/src/fontra/views/editor/edit-tools-pointer.js
@@ -817,6 +817,13 @@ function getTransformHandles(transformBounds, margin) {
 }
 
 function getTransformSelectionBounds(glyph, selection, getBackgroundImageBoundsFunc) {
+  const { point: pointIndices } = parseSelection(selection);
+  if (selection.size == 1 && pointIndices?.length == 1) {
+    // Return if only one point is selected.
+    // NOTE: We need to check selection.size == 1 because otherwise we don't know
+    // if someone selected a single point + a different object like a component, anchor, etc.
+    return undefined;
+  }
   const selectionBounds = glyph.getSelectionBounds(
     selection,
     getBackgroundImageBoundsFunc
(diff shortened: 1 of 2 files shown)


src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -817,6 +817,13 @@ function getTransformHandles(transformBounds, margin) {
 }
 
 function getTransformSelectionBounds(glyph, selection, getBackgroundImageBoundsFunc) {
+  const { point: pointIndices } = parseSelection(selection);
```
justvanrossum, line 820: Please write this so that parseSelection is only called when selection.size equals 1. Something like this maybe:

```js
   if (selection.size == 1 && parseSelection(selection).point?.length == 1) {
```

ollimeier: Good point. I changed it as recommended and also tested it. Works well. Done.

src/fontra/views/editor/edit-tools-pointer.js
```diff
@@ -817,6 +817,12 @@ function getTransformHandles(transformBounds, margin) {
 }
 
 function getTransformSelectionBounds(glyph, selection, getBackgroundImageBoundsFunc) {
+  if (selection.size == 1 && parseSelection(selection).point?.length == 1) {
+    // Return if only one point is selected.
+    // NOTE: We need to check selection.size == 1 because otherwise we don't know
```
justvanrossum, line 822: This comment no longer fully makes sense with the current code.

ollimeier: Mmhhh, for me it still makes sense. But I am happy to change it if you have any other suggestions.

justvanrossum: The comment suggests the `selection.size` check is "extra", but in the code it is the primary check. It is explaining things that should be obvious if you know how the selection set works, and how parseSelection() works. This is not the place to explain those concepts.

I think the whole comment should read:
```js
    // Return if only a single point is selected, as in that case the "selection bounds"
    // is not really useful for the user, and is distracting instead.
```

justvanrossum (approved):

PR #2028, opened 2025-02-13
Refactor/UI list style settings

giovanniTramonto:
First commit:
The `_updateVisibility` function never set any element to `display: none`. All `ui-list` (except glyph-search) were using `min-height` and `this.items` is defined as array in constructor (should not be `undefined`). Also I guess, if a list should become hidden, its better practice when the parent decides it (where data is set).

Second commit:
I guess setting `min-height` for container with custom properties is cleaner.

### DIFF
--- a/src/fontra/client/web-components/plugin-manager.js
+++ b/src/fontra/client/web-components/plugin-manager.js
@@ -34,7 +34,7 @@ export class PluginManager extends SimpleElement {
     observable.synchronizeWithLocalStorage("fontra.plugins");
     this.observable = observable;
     this.pluginList = new UIList();
-    this.pluginList.minHeight = "5em";
+    this.pluginList.style.setProperty("--container-min-height", "5em");
     this.pluginList.setItems(observable.model.plugins);
     this.pluginList.columnDescriptions = [
       {
--- a/src/fontra/client/web-components/ui-list.js
+++ b/src/fontra/client/web-components/ui-list.js
@@ -27,6 +27,7 @@ export class UIList extends UnlitElement {
 
     .container {
       overflow: auto;
+      min-height: var(--container-min-height, auto);
       height: 100%;
       width: 100%;
       border: solid 1px var(--border-color);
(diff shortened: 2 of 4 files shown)


justvanrossum:
> The _updateVisibility function never set any element to display: none 

I don't think that's true: it is set to none if no minHeight is given, and the items list is empty.

If you don't like the inline style thing that sets min-height, we can still change it to use a variable, but I'd prefer then to have a setter for `minHeight` that sets this variable, so the other code doesn't need to be changed. The external setting of the CSS variable to me isn't an api improvement.

giovanniTramonto:
Ah, no, you are right, `this.items?.length` casts to `false` if `0` : D 
But is in fact only used for glyph search. The other implementations always set min-height.
Actually, I don’t care so much for the min-height. It got me because of this display grid style. It looks odd in the stylesheet that grid gets overridden with grid. And I think it’s better to control this in parent where the items are produced.

src/fontra/client/web-components/glyph-search-list.js
```diff
@@ -113,6 +113,7 @@ export class GlyphSearchList extends SimpleElement {
   _setFilteredGlyphNamesListContent() {
     const filteredGlyphItems = this.glyphOrganizer.filterGlyphs(this.glyphsListItems);
     this.glyphNamesList.setItems(filteredGlyphItems);
+    this.glyphNamesList.style.display = filteredGlyphItems.length === 0 ? "none" : null;
```
justvanrossum, line 116: Would it perhaps be better to use `this.glyphNamesList.hidden` instead?

giovanniTramonto: OK Sure. That’s elegant. A pitty that `:host[hidden]` does not work, but it’s possible with a class.

giovanniTramonto: Actually, we can just set `[hidden]` in `shared.css`

giovanniTramonto: Or perhaps it’s better to set the shared `[hidden]` in a different MR, since it might cause regressions as well as it could be used for other cases too, i.e. `.ui-accordion-item[hidden]` in [`ui-accordion.js`.  ](https://github.com/googlefonts/fontra/blob/main/src/fontra/client/web-components/ui-accordion.js)

justvanrossum: I thought "hidden" was a standard html attribute? I must be missing something.

PR #2030, opened 2025-02-14
[reference font] Improve reference font var location

justvanrossum:


justvanrossum:
Oh ouch, this branch was accidentally based on #2017. Let me try this again.

PR #2039, opened 2025-02-18
Edit font infos metadata

### LINKED ISSUE #2023: [Font Info] Edit and author OpenType font info
As discussed in todays meeting, the goal is an area where we can edit the customData for each font source (keep in mind also Font Family level customData).

It should be similar to the new font guidelines list (at the end of the font source card). There we should be able to add and remove a new customData entry via ( + ) and ( - ) buttons. Enter a key and a value. So a list with two columns.

Have a look at `ufoInfoAttributesToRoundTrip` and also research how GlyphsApp customData is stored/managed and figure out the used keys.

### LINKED ISSUE #2088: [ui-list] Maybe we can allow items to carry their own formatters
Maybe we can allow items to carry their own formatters, then the item would be able override the column-level formatter:

```diff
diff --git a/src/fontra/client/web-components/ui-list.js b/src/fontra/client/web-components/ui-list.js
index 614540b94..abc61bdac 100644
--- a/src/fontra/client/web-components/ui-list.js
+++ b/src/fontra/client/web-components/ui-list.js
@@ -273,7 +273,8 @@ export class UIList extends UnlitElement {
             [colDesc.cellFactory(item, colDesc)]
           );
         } else {
-          const formatter = colDesc.formatter || DefaultFormatter;
+          const formatter =
+            item.formatters?.[colDesc.key] || colDesc.formatter || DefaultFormatter;
 
           cell = html.div(
             { class: `text-cell ${colDesc.key} ${colDesc.align || "left"}` },
```

(Maybe the key shouldn't be "formatter", but we can decide later.)

_Originally posted by @justvanrossum in https://github.com/googlefonts/fontra/pull/2039#discussion_r1961777821_

ollimeier:
Fixes #2023, fixes #2088

This is an early draft for feedback.

My main question: Is it the right direction to add and use a Fontra to UFO mapping for customData parameters (it's similar to `lineMetricsVerMapping`)?

I am also wondering if `fontInfoNameMapping` should be a dict instead of a list? But this is not related to this issue. I am just noticing it and wondering if we should make it consistent?


ollimeier:
Here a short demo:

https://github.com/user-attachments/assets/9072deba-c59a-42ba-8448-8e8cdc3311c7

src/fontra/views/fontinfo/panel-sources.js
```diff
@@ -676,6 +675,21 @@ class SourceBox extends HTMLElement {
       }, `edit guidelines`); // TODO: translation
     });
 
+    this.controllers.customData.addListener((event) => {
+      this.editSource((source) => {
+        source.customData = {};
+        for (const item of event.newValue) {
+          const value = parseFloat(item["value"]);
+          // TODO: How do we handle different types of values?
```
justvanrossum, line 683: Maybe we can allow items to carry their own formatters, then the item would be able override the column-level formatter:

```diff
diff --git a/src/fontra/client/web-components/ui-list.js b/src/fontra/client/web-components/ui-list.js
index 614540b94..abc61bdac 100644
--- a/src/fontra/client/web-components/ui-list.js
+++ b/src/fontra/client/web-components/ui-list.js
@@ -273,7 +273,8 @@ export class UIList extends UnlitElement {
             [colDesc.cellFactory(item, colDesc)]
           );
         } else {
-          const formatter = colDesc.formatter || DefaultFormatter;
+          const formatter =
+            item.formatters?.[colDesc.key] || colDesc.formatter || DefaultFormatter;
 
           cell = html.div(
             { class: `text-cell ${colDesc.key} ${colDesc.align || "left"}` },
```

(Maybe the key shouldn't be "formatter", but we can decide later.)

src/fontra/backends/designspace.py
```diff
@@ -62,6 +62,9 @@
 from .filewatcher import Change, FileWatcher
 from .ufo_utils import extractGlyphNameAndCodePoints
 
+# from glyphsLib.builder.custom_params import KNOWN_PARAM_HANDLERS
```
justvanrossum, line 65: We don't have a hard dependency on glyphsLib anymore, and also this would break the abstraction boundary, so this is not an option.

src/fontra/backends/designspace.py
```diff
@@ -123,77 +126,85 @@
     ("vendorID", "openTypeOS2VendorID"),
 ]
 
-
-ufoInfoPrefix = "ufo.info."
```
justvanrossum, line 127: There is a problem with the changing of the keys: it breaks existing data we have as part of the GS project.

Additionally, I'm not really comfortable with removing the "openType" and "postscript" prefixes.

I also don't understand why some names keep their OTF table prefix ("gaspRangeRecords", "headFlags", etc.) and others don't ("created", "codePageRanges", etc.).

src/fontra/client/core/ui-utils.js
```diff
@@ -243,6 +243,7 @@ export const DefaultFormatter = {
   },
 };
 
+// TODO: Maybe move formatters to utils.js, so we can write unittests for it.
```
justvanrossum, line 246: We can consider creating formatters.js, but utils.js is not the right place.

src/fontra/backends/designspace.py
ollimeier: Thanks for the info. I removed it. Closing this as resolved.

src/fontra/backends/designspace.py
ollimeier: I went back to the original list of `ufoInfoAttributesToRoundTrip`.

Background infos:
Why did I do a new mapping attributes mapping: I tried to make a similar to `fontInfoNameMapping`. 
There we have mappings like:` ("vendorID", "openTypeOS2VendorID"),`, but I was not able to do a proper name for all the attributes, for example removing `head` from `headFlags` would be too generic, plus: the whole dictionary was not final, it was up for discussion (where do we want to go). And I got some valid answers from you.

I mark this as resolved as well.

src/fontra/client/core/ui-utils.js
ollimeier: I created `formatters.js`, but did not move existing formatters, only the newly added + I added a `TODO` comment. I mark this as resolved.

ollimeier:
Note: We will add UI for `lineMetricsVerticalLayout` to this PR as we noticed that there is an overlap with `openTypeVheaVertTypoLineGap`.

ollimeier:
> Note: We will add UI for `lineMetricsVerticalLayout` to this PR as we noticed that there is an overlap with `openTypeVheaVertTypoLineGap`.

@justvanrossum I think this requires more and thorough consideration of how we really want to deal with it. There are too many open questions: How would a UI look like? Does it look like "Line metrics" or like "Custom Data". Is it always there, like "Guidelines", even though we don't really need it always? In general I would think it should look like "Line metrics", but we don't set these values always (which differs from "Line Metrics"), and we don't have an add or remove concept for that UI, yet.

ollimeier:
Notes from todays call together with @justvanrossum:
- maybe get rid of `ufoInfoPrefix = "ufo.info." `. Just wants to think about it.
- add-button should open a dialog where the user can select an attribute from a list (similar to "Add Source"-dialog select layer)

justvanrossum:
I just realized there are several items that you currently propose to edit as part of font sources, that are actually font-level bits of data, that we should incorporate in the "Font info" panel.

Examples:
- openTypeHeadCreated
- openTypeHeadLowestRecPPEM
- openTypeHeadFlags
- openTypeOS2Selection
- openTypeOS2Type

We should very carefully select which items are potentially interpolatable (see also: the MVAR table), and which are not, and therefore do not belong in the font sources panel.

justvanrossum:
I've merged the "Big Bang" changes from main, and fixed a couple of merge conflicts.

After you update your local clone, before you can continue you'll have to do `pip install -e .`. 
You should use the `--dev` option in the fontra invocation to make the bundler do an instantaneous rebuild when you make front end changes:

    fontra --dev filesystem path/to/folder/with/fonts

Alternatively, you can run `npm run bundle-watch` in a separate terminal window.

justvanrossum:
Note the new locations of .js source files: `src-js/` etc.

ollimeier:
> I just realized there are several items that you currently propose to edit as part of font sources, that are actually font-level bits of data, that we should incorporate in the "Font info" panel.
> 
> Examples:
> 
> * openTypeHeadCreated
> * openTypeHeadLowestRecPPEM
> * openTypeHeadFlags
> * openTypeOS2Selection
> * openTypeOS2Type
> 
> We should very carefully select which items are potentially interpolatable (see also: the MVAR table), and which are not, and therefore do not belong in the font sources panel.

The mentioned attributes have been in `ufoInfoAttributesToRoundTrip` before my changes. So, this is nothing I proposed.
But I agree, we should speak about it.

Currently we distinguish between:
- Font Family Level
- Font Sources Level

But what do we do with attributes which belong to single instances/exports, like  `openTypeOS2Panose` ? Panose should be set for each static font individually, not on Font Source Level and it's definitely not interpolatable.
So, I am wondering if we need something like:
- Font exports/instances

If we see "Font info" panel as a collection of font family infos, I would disagree with setting `openTypeOS2Selection` or `openTypeHeadLowestRecPPEM` in the "Font info" panel, because for me it is more a font instance info than a font family info.
`openTypeOS2Selection`: The 'Bold' might be set for each static font individually (depending if it's a style linked bold font)
`openTypeHeadLowestRecPPEM`: In a font family with optical size, the smallest readable size might different between fonts (eg. Text vs. Display)

Does it make sense to you? Or is there a fundamental misunderstanding on my side?

ollimeier:
@justvanrossum  About the label of customData, currently 'Advanced Information': What do you think about 'Low-level settings'?

src-js/fontra-webcomponents/src/custom-data-list.js
```diff
@@ -0,0 +1,347 @@
...
+      }
+
+      if (customDataKey == undefined) {
+        setWarning("⚠️ Key is empty, please enter."); // TODO: translation
+        return;
+      }
+
+      if (!customDataNames.includes(customDataKey)) {
+        // We know we have a key, but it is not supported, yet.
+        setWarning(`⚠️ ${translate("Key not supported, yet.")}`); // TODO: translation
```
justvanrossum, line 211: The "yet" makes no sense for garbage input, better to remove it.

src-js/fontra-webcomponents/src/custom-data-list.js
```diff
@@ -0,0 +1,347 @@
...
+
+    const labelList = new UIList();
+    labelList.classList.add("fontra-ui-font-info-sources-panel-list-element");
+    labelList.style = `min-width: 12em;`;
+    labelList.columnDescriptions = [
+      {
+        key: "key",
+        title: "Key", // TODO: translation
+        width: "14em",
+        editable: true,
```
justvanrossum, line 99: I don't think `editing` is useful, given we restrict ourselves to known keys.

justvanrossum: Also, if I dismiss the dialog, the key goes into edit mode, which isn't desirable, either.

justvanrossum (commented):
When adding a key, the list isn't reordered, but the key is added at the end. When reloading, the order differs. I think the order should be maintained at all times.

I left a couple more comments.

justvanrossum:
What determines the width of the "add key" dialog when selecting a key from the list? The dialog gets very wide.

justvanrossum:
Some keys (for example `openTypeOS2SubscriptYOffset`) do not allow floating points (writing to UFO will complain). We should probably use an `IntegerFormatter` for these. We may also need a `UnsignedIntegerFormatter` and/or `UnsignedNumberFormatter`, if there are float values that must be non-negative (not sure that situation exists, off the top of my head).

justvanrossum:
`openTypeOS2WinAscent`, `openTypeOS2WinDescent` are examples of "unsigned integer".

src-js/fontra-webcomponents/src/custom-data-list.js
ollimeier: Thanks. Done.

src-js/fontra-webcomponents/src/custom-data-list.js
ollimeier: Thanks. Done.

ollimeier:
> What determines the width of the "add key" dialog when selecting a key from the list? The dialog gets very wide.

It depends on the info text content, it expands the width of the dialog to `max-width: 44em;`.

justvanrossum:
> It depends on the info text content

Can we make the content wrap so the dialog doesn't change its width so radically?

src/fontra/views/fontinfo/panel-sources.js
ollimeier: I am closing this as resolved, because I created an issue out of your comment, so we can take care of it later: https://github.com/googlefonts/fontra/issues/2088

ollimeier:
> When adding a key, the list isn't reordered, but the key is added at the end. When reloading, the order differs. I think the order should be maintained at all times.
> 
> I left a couple more comments.

I fixed the reordering with the last commit: https://github.com/googlefonts/fontra/pull/2039/commits/9482f2c6f455ccfc795ee1c8e00658e5899efe74

justvanrossum (commented):
The text input for the key should have focus when the dialog opens. Unfortunately this requires a bit of a hack. See for example this:
https://github.com/googlefonts/fontra/blob/12fe8ab08f53d397f1b9ceef6dd70fa046ba932d/src-js/views-editor/src/panel-designspace-navigation.js#L1532-L1535

ollimeier:
> The text input for the key should have focus when the dialog opens. Unfortunately this requires a bit of a hack. See for example this:
> 
> https://github.com/googlefonts/fontra/blob/12fe8ab08f53d397f1b9ceef6dd70fa046ba932d/src-js/views-editor/src/panel-designspace-navigation.js#L1532-L1535

I missed that. Thanks for the hint.

src-js/fontra-webcomponents/src/custom-data-list.js
```diff
@@ -0,0 +1,342 @@
+import * as html from "@fontra/core/html-utils.js";
+import { SimpleElement } from "@fontra/core/html-utils.js";
+import { translate } from "@fontra/core/localization.js";
+import { ObservableController } from "@fontra/core/observable-object.js";
+import { DefaultFormatter, labeledTextInput } from "@fontra/core/ui-utils.js";
+import { zip } from "@fontra/core/utils.js";
+import { dialogSetup, message } from "@fontra/web-components/modal-dialog.js";
```
justvanrossum, line 7: `message` is not used.

src-js/views-fontinfo/src/panel-axes.js
```diff
@@ -917,7 +917,7 @@ export function updateRemoveButton(list, buttons) {
   });
 }
 
-function arraysEqual(arrayA, arrayB) {
+export function arraysEqual(arrayA, arrayB) {
```
justvanrossum, line 920: I don't think you actually import this function, so this change isn't needed.

src-js/fontra-webcomponents/src/custom-data-list.js
ollimeier: Thanks. Done.

src-js/views-fontinfo/src/panel-axes.js
ollimeier: Thanks. Done.

justvanrossum (approved):

PR #2041, opened 2025-02-20
Small bug fixes for text entry panel

qwerasd205:
These two small UI bugs were really bothering me so I fixed them.
1. `/slash` -> `/` -> *nothing* (or invalid character) (triggered by modification to the glyph lines such as by `cmd+left/right`)
2. Text area too tall for browser window, bad overflow behavior

Video demonstrating the bugs:

https://github.com/user-attachments/assets/7bf27d8e-d988-4aee-ba80-2c7ae572c29c

These are very trivial fixes, since I wanted to start off by contributing something small- but there are various improvements that would be desirable for the text panel, such as the ability to toggle whether to represent certain classes of glyphs as their character or `/name`, for example this is kind of annoying when working with combining glyphs since you may type `A/acutecomb` and have it turned in to `Á`, making it not possible to hand edit e.g. which base character it's applied on top of via the text panel (you can right click -> replace glyph on canvas, but that's not at all ergonomic).

### DIFF
--- a/src/fontra/client/core/glyph-lines.js
+++ b/src/fontra/client/core/glyph-lines.js
@@ -109,7 +109,7 @@ export function textFromGlyphLines(glyphLines) {
       if (glyphInfo.character === "/") {
         // special-case slash, since it is the glyph name indicator character,
         // and needs to be escaped
-       textLine += "//";
+        textLine += "//";
       } else if (glyphInfo.character) {
         textLine += glyphInfo.character;
       } else {
--- a/src/fontra/client/core/glyph-lines.js
+++ b/src/fontra/client/core/glyph-lines.js
@@ -106,7 +106,7 @@ export function textFromGlyphLines(glyphLines) {
     let textLine = "";
     for (let i = 0; i < glyphLine.length; i++) {
       const glyphInfo = glyphLine[i];
-      if (glyphInfo.character) {
+      if (glyphInfo.character && glyphInfo.character !== "/") {
         textLine += glyphInfo.character;
       } else {
         textLine += "/" + glyphInfo.glyphName;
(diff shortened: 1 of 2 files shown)


src/fontra/client/core/glyph-lines.js
```diff
@@ -106,7 +106,7 @@ export function textFromGlyphLines(glyphLines) {
     let textLine = "";
     for (let i = 0; i < glyphLine.length; i++) {
       const glyphInfo = glyphLine[i];
-      if (glyphInfo.character) {
+      if (glyphInfo.character && glyphInfo.character !== "/") {
```
justvanrossum, line 109: I prefer "/" to be inserted as "//" which is the character escape for "/":

```suggestion
      if (glyphInfo.character) {
       textLine += "//";
      } else if (glyphInfo.character) {
```

justvanrossum (commented):
Thanks! The text entry fix is good to go: thanks for the catch, this must have been a regression.

I left a suggestion for the slash issue.

src/fontra/client/core/glyph-lines.js
justvanrossum: (The original suggestion was messed up, I think I fixed it now.)

src/fontra/client/core/glyph-lines.js
```diff
@@ -106,7 +106,11 @@ export function textFromGlyphLines(glyphLines) {
     let textLine = "";
     for (let i = 0; i < glyphLine.length; i++) {
       const glyphInfo = glyphLine[i];
-      if (glyphInfo.character) {
+      if (glyphInfo.character === "/") {
+        // special-case slash, since it is the glyph name indicator character,
+        // and needs to be escaped
+       textLine += "//";
```
justvanrossum, line 112: Formatting oops.
```suggestion
        textLine += "//";
```

PR #2057, opened 2025-02-26
ui-list paste plaintext-only

### LINKED ISSUE #2043: [ui-list] Prevent pasting rich text
From Mark:
On a different note, currently in the “Edit glyph axes” dialogue box, it seems that it’s still possible to paste whatever that is in the clipboard into the field, like a whole shape. Although it disappears after I click somewhere else.
Also if I copy/paste a text into the “Name” field, it seems to keep the formatting like background, font size, etc.
I think maybe it’s better to limit these field to text/numeral only?

https://github.com/user-attachments/assets/42b66c80-9717-4f26-a450-1d173eb6f9c6

https://github.com/user-attachments/assets/b7c11061-8966-465e-9fc0-465f31d0f442

From @justvanrossum: 
We should somehow prevent pasting rich text. I wonder if the shape was SVG?

ollimeier:
Fixes #2043

### DIFF
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -311,10 +311,7 @@ export async function fetchJSON(url, options) {
 export function isActiveElementTypeable() {
   const element = findNestedActiveElement(document.activeElement);
 
-  if (
-    element.contentEditable === "plaintext-only" ||
-    element.contentEditable === "true"
-  ) {
+  if (element.isContentEditable) {
     return true;
   }
   if (element.tagName.toLowerCase() === "textarea") {
--- a/src/fontra/client/core/utils.js
+++ b/src/fontra/client/core/utils.js
@@ -311,7 +311,10 @@ export async function fetchJSON(url, options) {
 export function isActiveElementTypeable() {
   const element = findNestedActiveElement(document.activeElement);
 
-  if (element.contentEditable === "true") {
+  if (
+    element.contentEditable === "plaintext-only" ||
+    element.contentEditable === "true"
+  ) {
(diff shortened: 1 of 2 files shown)


justvanrossum (approved):
Awesome! I didn't know about `"plaintext-only"`.

justvanrossum:
https://github.com/googlefonts/fontra/blob/715766525773908242750e99b5eec268ef290981/src/fontra/client/core/utils.js#L314

This may need to be changed as well. Maybe it is as simple as changing it to `element.contentEditable !== "false"`.

ollimeier:
> https://github.com/googlefonts/fontra/blob/715766525773908242750e99b5eec268ef290981/src/fontra/client/core/utils.js#L314
> 
> This may need to be changed as well. Maybe it is as simple as changing it to `element.contentEditable !== "false"`.

`element.contentEditable` can also be `inherit` (which would also be  !== "false") why I think it's probably better to use `element.contentEditable === "plaintext-only" ||`.
I would like to write a unittest for that, but I don't know how, because `isActiveElementTypeable` uses `document.activeElement `and I don't know how to mock up such an element for a unittest.

justvanrossum:
Ah, actually, we should use isContentEditable: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/isContentEditable

ollimeier:
> Ah, actually, we should use isContentEditable: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/isContentEditable

Thanks for the hint. That's much better. Please see the latest commit.

PR #2053, opened 2025-02-26
Split Javascript and Python

simoncozens:
Implements #1952.

Issue #2026, opened 2025-02-13
Guideline label alignment is not consistent


ollimeier:
> I noticed that the guideline label alignment is not consistent. This is technically unrelated to this PR, and has to do with the (use of) getTextHeight(): The text height is the total distance between (box) descender and (box) ascender, but I think the drawing is relative to the baseline. So labels containing _descender_ letters are poorly aligned:
<img width="110" alt="image" src="https://github.com/user-attachments/assets/a5c7f7db-d17d-4d4a-aaa2-49f059c9fb46" />

_Originally posted by @justvanrossum in https://github.com/googlefonts/fontra/pull/2021#pullrequestreview-2612014193_

> Instead of getTextHeight(), we should consider a function getVerticalCenter(), that returns (metrics.actualBoundingBoxAscent - metrics.actualBoundingBoxDescent) / 2 (given that actualBoundingBoxDescent is a positive value for a distance below the baseline)

_Originally posted by @justvanrossum in https://github.com/googlefonts/fontra/pull/2021#issuecomment-2653863144_

justvanrossum:
Fixed by https://github.com/googlefonts/fontra/pull/2021

PR #2062, opened 2025-02-28
Fix minor menu key navigation problems

### LINKED ISSUE #2061: [menu, menu bar, context menu] Three minor key navigation problems with submenus
1. When using arrow keys to navigate a menu, submenu's do not unfold. For example, in the editor, click the View, arrow down until "Select glyph/source/layer", the submenu does not show
2. When we are in a submenu, typing Escape only dismisses the submenu, but I think it should dismiss all open menus
3. When control-clicking in the editor and press Arrow down, the "Edit" submenu immediately opens, and we don't get a chance to arrow-navigate the top level context menu

In general, I think arrow left/right should go in and out of the submenu. macOS has good behavior that I think we should aim to copy.

I don't think we have deeper nested submenus, but it would be good if those would work the same as well.

@giovanniTramonto, this might interest you?

giovanniTramonto:
Fixes #2061

Menu behavior is now pretty close to mac OS navigation.

### DIFF
--- a/src-js/fontra-core/src/utils.js
+++ b/src-js/fontra-core/src/utils.js
@@ -665,10 +665,6 @@ export function sleepAsync(ms) {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
-export async function nextFrame() {
-  return new Promise((resolve) => requestAnimationFrame(() => resolve()));
-}
-
 export function readFileOrBlobAsDataURL(fileOrBlob) {
   return new Promise((resolve, reject) => {
     const reader = new FileReader();
--- a/src-js/fontra-core/src/utils.js
+++ b/src-js/fontra-core/src/utils.js
@@ -665,6 +665,10 @@ export function sleepAsync(ms) {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
+export async function nextFrame() {
+  return new Promise((resolve) => requestAnimationFrame(() => resolve()));
+}
+
 export function readFileOrBlobAsDataURL(fileOrBlob) {
   return new Promise((resolve, reject) => {
     const reader = new FileReader();
--- a/src-js/fontra-webcomponents/src/menu-panel.js
+++ b/src-js/fontra-webcomponents/src/menu-panel.js
@@ -7,7 +7,7 @@ import {
 import { dispatchCustomEvent } from "@fontra/core/event-utils.js";
 import * as html from "@fontra/core/html-utils.js";
 import { SimpleElement } from "@fontra/core/html-utils.js";
-import { enumerate, nextFrame, reversed } from "@fontra/core/utils.js";
+import { enumerate, reversed, sleepAsync } from "@fontra/core/utils.js";
 import { InlineSVG } from "@fontra/web-components/inline-svg.js";
 import { themeColorCSS } from "./theme-support.js";
 
@@ -144,7 +144,7 @@ export class MenuPanel extends SimpleElement {
       this.hide();
     } else if (this.immediatelyActive) {
       // Wait next cycle to ensure `tabindex` is available
-      await nextFrame();
+      await sleepAsync(0);
       this.setActive(true);
     }
   }
--- a/src-js/fontra-webcomponents/src/menu-panel.js
+++ b/src-js/fontra-webcomponents/src/menu-panel.js
@@ -165,9 +165,19 @@ export class MenuPanel extends SimpleElement {
     this.style.display = "none";
   }
 
-  show() {
+  async show(options = { animated: false }) {
     this.hidden = false;
     this.style.display = null;
+    if (options.animated) {
+      // hide immediately since animation is delayed
+      this.style.opacity = 0;
+      const { finished } = this.animate(
+        { opacity: 1 },
+        { delay: 250, duration: 50, iterations: 1, easing: "ease-in-out" }
+      );
+      await finished;
+      this.style.opacity = null;
+    }
   }
 
   place() {
@@ -290,7 +300,7 @@ export class MenuPanel extends SimpleElement {
     );
   }
 
-  selectItem(item, submenuHidden = false) {
+  selectItem(item, fromChild = false) {
     for (const panel of MenuPanel.openMenuPanels) {
       panel.setActive(panel === this);
       if (panel.childOf === this) {
@@ -320,7 +330,7 @@ export class MenuPanel extends SimpleElement {
         const itemRect = item.getBoundingClientRect();
 
         this.submenu = new MenuPanel(this.menuItems[item.dataset.index].getItems(), {
-          hidden: submenuHidden,
+          hidden: true,
           immediatelyActive: false,
           position: {
             x: itemRect.width,
@@ -333,6 +343,9 @@ export class MenuPanel extends SimpleElement {
           },
         });
         this.menuElement.appendChild(this.submenu);
+        if (!fromChild) {
+          this.submenu.show({ animated: true });
+        }
         item.classList.add("has-open-submenu");
       }
     }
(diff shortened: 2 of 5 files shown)


justvanrossum:
Thanks!

As a FYI: we temporarily stop merging front end PRs because there is a huge change incoming that will move many files: https://github.com/googlefonts/fontra/pull/2053 

This big PR should get merged very soon, but your PR may have to be manually rebased for the new file location. We will help with that if needed.

Thanks for your understanding.

justvanrossum:
(Feel free to convert to a draft PR to signal WIP)

justvanrossum:
The Big Bang js-split PR has been merged. Let me know if you need help reworking this PR for the new file locations.

Pro tip: we now use a bundler, and you'll have to do `pip install -e .` before you can continue, and you should use the `--dev` option in the `fontra` invocation to make the bundler do an instantaneous rebuild when you make front end changes:

    fontra --dev filesystem path/to/folder/with/fonts

Alternatively, you can run `npm run bundle-watch` in a separate terminal window.

justvanrossum:
So, the file that you've edited now lives here:

- https://github.com/googlefonts/fontra/blob/main/src-js/fontra-webcomponents/src/menu-panel.js

Apart from the imports, nothing should have changed.

giovanniTramonto:
OK Great. Thanks for the update. Everything works on my side.

giovanniTramonto:
Successfully rebased and updated : )

giovanniTramonto:
There is a lot of change in 5d59009bcf588c7d0319640233c184dd081ffd33 at once but a few things were a confusing, so it felt easier to rework parts extensively. I hope that’s OK.

src-js/fontra-core/src/utils.js
```diff
@@ -665,6 +665,10 @@ export function sleepAsync(ms) {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
+export async function nextFrame() {
```
justvanrossum, line 668: The idiom for this in Fontra is to either use setTimeout with 0ms delay, or to do `await sleepAsync(0)`. No need for a new function.

giovanniTramonto: Alright 👍

giovanniTramonto:
> Maybe you're not ready for a review, but I noticed this: if you arrow down to a menu item with a submenu, the submenu should open (macOS does this with a slight delay, perhaps to avoid to much motion when typing fast), and arrow right should navigate into the submenu. If this is work in progress then sorry for the noise!

Thanks for pointing out. It’s delayed now : )
Actually, since there are a lot of changes already (and #2026 is fixed) I was wondering if it’s better to complete this PR and to see again more details in a follow-up. I.e. pressing `ArrowRight` when in submenu should dismiss the menu-panel and navigate to next menu item. Or when `with-submenu` is selected pressing `ArrowLeft` should navigate to previous menu-item. And I did not even try with children of submenu : | At the same time I could also go on and add to this PR. Not sure if anyone else is planning changes to the menu soon? (just to avoid conflicts)

justvanrossum:
I'm afraid this isn't ready to merge yet.

- clicking on a menu opens it, clicking again should close it again (previous behavior, and macOS)
- key-navigating into a submenu works (yay!), but the parent item stays "selected" (red). When doing it with the pointer, this does not happen.

With pointer (good):
image

With keys (not good):
image

- When key-navigating to a submenu and pressing enter, the menu action is not triggered
- Possibly related, in an editor:
  - click the View menu
  - key-navigate to Select glyph/source/layer -> Select previous glyph
  - type Enter
  - (action is not performed, as noted above)
  - click the View menu again
  - try to key-navigate: it gets stuck after the first item, and Escape doesn't work

I don't expect other changes here, so I'm not worried about conflicts.

giovanniTramonto:
Thanks for investigating. Bugs are fixed. I will continue improving towards OS behavior in this PR.

justvanrossum:
Excellent, it works quite well now!

Tiny detail: normally, when you're at a menu item, arrow left will go to the previous menu. This only works if the menu item does not have a submenu, but I think it should behave the same.

giovanniTramonto:
Ah, that’s probably what meant above :
> Or when `with-submenu` is selected pressing `ArrowLeft` should navigate to previous menu-item.

giovanniTramonto:
Following issues are fixed 
> I.e. pressing `ArrowRight` when in submenu should dismiss the menu-panel and navigate to next menu item. Or when `with-submenu` is selected pressing `ArrowLeft` should navigate to previous menu-item.

Still only works one submenu level. Maybe deeper levels could be treated in follow-up?

justvanrossum:
Ok, some things got better, others got worse.

1. Click on a menu, type Escape: the menu title gets a focus ring
2. Click on a menu, type ArrowLeft/Right: the menu title gets a focus ring
3. Click on a menu, click again to dismiss. Now type ArrowDown: menu opens again

giovanniTramonto:
OK. Issues fixed.
The focus ring was actually a result of `:focus-visible` allowing to tab-navigate directly into menu when arriving on the site.

giovanniTramonto:
Branch is rebased to latest main

src-js/views-fontoverview/src/fontoverview.js
```diff
@@ -552,6 +552,9 @@ export class FontOverviewController extends ViewController {
   }
 
   handleKeyDown(event) {
+    if (document.activeElement === document.querySelector("menu-bar")) {
```
justvanrossum, line 555: This doesn't seem to work. If I click a menu, then type arrow keys, the cell view navigates, but not the menu.

justvanrossum: Should the (font overview's) key event handler perhaps not be attached to the document, but to the element that contains everything except the menu bar?

giovanniTramonto: Yes, could cause less conflicts when its attached to `main-container`. I just pushed a try.

justvanrossum: This works, but is not the same as before: arrow-keying through the glyph cells now only works if the glyph cell view has focus. So arrow-keying does not work if you come fresh to the page, or once you've clicked in the sidebar.

giovanniTramonto: OK I reverted the change. The navigation issue was fixed anyways in [b707c42](https://github.com/googlefonts/fontra/pull/2062/commits/b707c4228e64aa11dfc6d412e878a94eb85d6c57)

justvanrossum: But now this is back:
- click on menu
- (optionally arrow navigate anywhere)
- (optionally move the pointer anywhere in any menu, but don't click)
- click again anywhere in the menubar, to dismiss
- type arrow down
- menu appears

giovanniTramonto: Should work now

src-js/fontra-webcomponents/src/menu-panel.js
```diff
@@ -415,3 +477,23 @@ function getMenuContainer() {
...
 }
+
+function isEnabledItem(item) {
+  return item.classList.contains("enabled");
+}
+
+// Get activeElement, even if in ShadowRoot.
+// @param root: Document | ShadowRoot
+// @return Element | null
+function getActiveElement(root = document) {
```
justvanrossum, line 488: Looks like this is equivalent to `findNestedActiveElement()` in "utils.js".

giovanniTramonto: Oh, great! Thanks

justvanrossum (approved):
This looks and works great, thank you so much!

PR #2060, opened 2025-02-27
Drag to open menu

giovanniTramonto:
Resolves #2049

I realized that `navigateMenuBar` did not work, because of `stopImmediatePropagation` here  https://github.com/googlefonts/fontra/compare/main...giovanniTramonto:fontra:drag-to-open-menu?expand=1#diff-b35c3a0aab4b807296ca1634d437d26deffcdb268652aabf8495b802c1ceabcaR304 
So solved this too in b7fed8c19add114b0f9acac1fa386bc89ef9a91c by custom event

PR #2069, opened 2025-03-03
Fix context-menu close #2068

### LINKED ISSUE #2068: Context menu doesn't go away
Scenario, in the editor:
- control- or right-click anywhere to get the context menu
- click anywhere outside the context menu should make it go away, but it doesn't

This is a regression caused by d86f79db04f1783fb7804ecfe8ea143b63c5110d. @giovanniTramonto can you help?

giovanniTramonto:
Fixes #2068

### DIFF
--- a/src-js/fontra-webcomponents/src/menu-bar.js
+++ b/src-js/fontra-webcomponents/src/menu-bar.js
@@ -145,6 +145,7 @@ export class MenuBar extends SimpleElement {
     };
     const menuPanel = new MenuPanel(items, {
       position,
+      context: "menu-bar",
       onSelect: () => this.closeMenu(),
       onClose: () => this.closeMenu(),
     });
--- a/src-js/fontra-webcomponents/src/menu-bar.js
+++ b/src-js/fontra-webcomponents/src/menu-bar.js
@@ -45,7 +45,7 @@ export class MenuBar extends SimpleElement {
     this.contentElement.classList.add("menu-bar");
     this.render();
     window.addEventListener("blur", this.closeMenu.bind(this));
-    window.addEventListener("click", this.onClick.bind(this));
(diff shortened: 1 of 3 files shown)


src-js/fontra-webcomponents/src/menu-bar.js
```diff
@@ -145,6 +145,7 @@ export class MenuBar extends SimpleElement {
     };
     const menuPanel = new MenuPanel(items, {
       position,
+      context: "menu-bar",
```
justvanrossum, line 148: I'm just wondering, we didn't need this before, why do we need it now?

justvanrossum:
This doesn't fix everything: repeated *control*-click still give multiple context menus.

justvanrossum:
Would it be better to revert #2060 first, and do a fresh attempt?

src-js/fontra-webcomponents/src/menu-bar.js
giovanniTramonto: It’s easier to figure out if target is in `menu-bar`, when testing if the panel should close. The `mousedown` event in panel fires at last and would close the panel immediately after opening, when hitting menu-item. Before this was done on `click`. Now all happens in `mousedown`.
I tried to avoid by using click for the closing. But it did not change much, actually. And in OS menu all is done with `mousedown`. So now works with `mousedown` too.

giovanniTramonto:
> This doesn't fix everything: repeated _control_-click still give multiple context menus.

Should work now

giovanniTramonto:
> Would it be better to revert #2060 first, and do a fresh attempt?

Sure. Feel free to revert

justvanrossum:
Thanks, I can confirm it works. I'll merge.

(Would be nice to use dispatchCustomEvent() in other places, too.)

PR #2071, opened 2025-03-03
Fix redirects

### LINKED ISSUE #2070: Redirect "old" URLs does not work anymore
With #2053 merged, the URL redirect from the "old" format does not work anymore.

This is in server.py.

@simoncozens

simoncozens:
Fixes #2070, apparently those things are only regular expressions when they're captured.


justvanrossum:
Thanks!

PR #2072, opened 2025-03-03
Improve webpack config for HtmlBundlerPlugin

webdiscus:
Hello,

I'm the author of the [html-bundler-webpack-plugin](https://github.com/webdiscus/html-bundler-webpack-plugin).
I want to improve the Webpack configuration to save processed JS and CSS in separate folders.

Defaults Webpack places generated HTML, JS and CSS files into directory specified in the `output.path` Webpack option.
After running `npm run bundle`, the `src/fontra/client/` directory contains a mix of generated HTML, CSS, and JS files.

image

This PR allows configuring the Bundler Plugin to save JS and CSS, resolved in HTML, into separate directories:
- CSS -> `src/fontra/client/css/`
- JS -> `src/fontra/client/js/`

See  [js](https://github.com/webdiscus/html-bundler-webpack-plugin?tab=readme-ov-file#option-js) and [css](https://github.com/webdiscus/html-bundler-webpack-plugin?tab=readme-ov-file#option-css) plugin options.

Hashed file names ensure that the output file name is unique, avoiding conflicts with copied files.
Now, the root web directory `src/fontra/client/` contains only generated HTML files.

Pasted Graphic 2

All assets will be loaded in a browser:

Pasted Graphic 1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -248,6 +248,9 @@ dist
 # Stores VSCode versions used for testing VSCode extensions
 .vscode-test
 
+# JetBrains IDE
+.idea
+
 # yarn v2
 .yarn/cache
 .yarn/unplugged
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
       "devDependencies": {
         "copy-webpack-plugin": "^12.0.2",
         "css-loader": "^7.1.2",
-        "html-bundler-webpack-plugin": "^4.12.2",
+        "html-bundler-webpack-plugin": "^4.18.0",
         "npm": "^11.0.0",
         "prettier-plugin-organize-imports": "^4.1.0",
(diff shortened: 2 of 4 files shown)


justvanrossum:
@webdiscus can you fix the linting problem? We use pre-commit.

@simoncozens can you please review this?

webdiscus:
> @webdiscus can you fix the linting problem? We use pre-commit.

The linting is fixed.

simoncozens:
Tested here, LGTM.

simoncozens (approved):

justvanrossum:
@simoncozens, do you gather we'd need similar changes to https://github.com/googlefonts/fontra-rcjk, or is this local enough that it can't cause breakage if we don't?

PR #2091, opened 2025-03-14
[designspace] Fix editing sources

### LINKED ISSUE #2040: Designspace backend has problem modifying sources when variable components are present
### To reproduce the issue:
- Create a designspace-based project
- Create a variable component in it (a glyph with a local axis and at least one non-default source defined for that axis)
- Go to the Sources panel
- Try to modify any data there, such as the line metrics or global guidelines
- Return to the editor view and try to do anything

### Result:
Errors, caused by the fact that the designspace backend replaces its stored sources and layers (`self.dsSources`, `self.ufoLayers`) with just the font ones (in `putSources`), neglecting the glyph-local ones, so when it then tries to get that glyph it's missing the layer for it. Fontra has to be restarted to recover from this.

I considered implementing a fix but I'm not familiar enough with the codebase to be confident that my fix is right, since there's a lot of stuff interacting here and I couldn't find a detailed explanation of how it's architected- so I figured I'd just make this issue and hope it's a trivial problem that a regular contributor can solve easily. 🤞

FWIW, this is a partial fix I wrote before deciding I wasn't confident enough in my understanding of the codebase:
```diff
--- a/src/fontra/backends/designspace.py
+++ b/src/fontra/backends/designspace.py
@@ -1030,13 +1030,22 @@ class DesignspaceBackend:
             {s.identifier: s for s in self.dsSources if s.identifier not in sources}
         )
 
-        self.dsSources = newDSSources
+        # Update stored layers
+        #
+        # Not all layers are associated with a source, since layers exist for
+        # variable components as well, so we can't just replace the list of
+        # ufo layers with one derived from the current dsSources, instead we
+        # selective throw out any layers from removed sources and add layers
+        # from newly added sources.
+
+        addedLayers = [s.layer for s in newDSSources if s not in self.dsSources]
+        removedLayers = [s.layer for s in self.dsSources if s not in newDSSources]
+
+        self.ufoLayers.items = addedLayers + [
+            layer for layer in self.ufoLayers if layer not in removedLayers
+        ]
 
-        # Prune layers
-        newLayers = ItemList()
-        for dsSource in newDSSources:
-            newLayers.append(dsSource.layer)
-        self.ufoLayers = newLayers
+        self.dsSources = newDSSources
 
         axisOrder = [axis.name for axis in self.dsDoc.axes]
         newSourceDescriptors = [
```
It's definitely incomplete since it doesn't account for `self.dsSources`, only `self.ufoLayers`.

Huge props to all the contributors who have worked on this project! Really glad to see it becoming genuinely useful and would love to become more familiar with the codebase so I can help push it even further ^-^

justvanrossum:
This fixes #2040.

`putSources()` threw away ufo layer definitions for non-source layers.

This additionally fixes two bugs I found while fixing the original bug:
- fontSource.name was not written to the ds source styleName attribute
- ds source styleName attribute was not _read_ for sparse masters (non-default layerName)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog for Fontra
 
+## 2025-03-16
+
+- Fixed several bugs in the designspace backend related to editing font sources. [Issue 2040](https://github.com/googlefonts/fontra/issues/2040) [PR 2091](https://github.com/googlefonts/fontra/pull/2091)
+
 ## 2025-03-14
 
 New features:
--- a/src/fontra/backends/designspace.py
+++ b/src/fontra/backends/designspace.py
@@ -351,15 +351,16 @@ class DesignspaceBackend:
         for source in self.dsDoc.sources:
             if self._familyName is None and source.familyName:
                 self._familyName = source.familyName
-            reader = manager.getReader(source.path)
+            ufoPath = os.path.normpath(source.path)
+            reader = manager.getReader(ufoPath)
             defaultLayerName = reader.getDefaultLayerName()
             ufoLayerName = source.layerName or defaultLayerName
 
-            sourceLayer = self.ufoLayers.findItem(path=source.path, name=ufoLayerName)
+            sourceLayer = self.ufoLayers.findItem(path=ufoPath, name=ufoLayerName)
             if sourceLayer is None:
                 sourceLayer = UFOLayer(
                     manager=manager,
-                    path=source.path,
+                    path=ufoPath,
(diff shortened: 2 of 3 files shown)


justvanrossum:
Hm, those Windows test failure seems unrelated to these changes, yet they pass in a different branch. Will need to look into this later.

justvanrossum:
Hm, the Windows failure seems a genuine bug, that magically only manifests itself on Windows only. Need to investigate deeper.

PR #2115, opened 2025-03-25
Fix menu-panel keydown propagation

giovanniTramonto:
Fixes #2113


justvanrossum (approved):
Excellent, thank you!

PR #2128, opened 2025-03-27
Fix undefined in axis value

### LINKED ISSUE #2127: [axes info panel] undefined in axis value
When you create a new Axis value the is an `undefined` in the dummy filed. Please see the following screenshot:

Image

ollimeier:
Fixes #2127

### DIFF
--- a/src-js/fontra-webcomponents/src/ui-list.js
+++ b/src-js/fontra-webcomponents/src/ui-list.js
@@ -305,10 +305,13 @@ export class UIList extends UnlitElement {
             [colDesc.cellFactory(item, colDesc)]
           );
         } else {
-          const formatter =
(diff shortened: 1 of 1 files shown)


justvanrossum:
Hm this can’t be the correct solution, as formatted is allowed to be undefined there by design, as eve fenced by the condition below. 
I would suspect this change breaks the glyph source list bold/gray formatting.

We may have to turn undefined into “” explicitly here.

ollimeier:
> Hm this can’t be the correct solution, as formatter is allowed to be undefined there by design, as evidenced by the condition below. I would suspect this change breaks the glyph source list bold/gray formatting.
> 
> We may have to turn undefined into “” explicitly here.

You're right, it breaks the "glyph source list":

Screenshot 2025-03-27 at 12 26 08

justvanrossum (approved):

PR #2159, opened 2025-05-06
Fix reference-font drop-area highlight border #2157

### LINKED ISSUE #2157: Reference Font: highlight border on drop area doesn't look right anymore
When hovering a font onto the drop area of the Reference Font panel, it used to show a thick light gray border all around the area. Currently, this border is only visible at the bottom of the area (font list), but not at the top, left or right.

@giovanniTramonto, I wonder if this may be related to your earlier work.

giovanniTramonto:
Fixes #2157

@justvanrossum I was wondering if the layout becomes easier if the upload-list and the preview get joined. What do you think?

### DIFF
--- a/src-js/views-editor/src/panel-reference-font.js
+++ b/src-js/views-editor/src/panel-reference-font.js
@@ -195,15 +195,14 @@ export default class ReferenceFontPanel extends Panel {
       align-content: start;
     }
 
-    .reference-font-items {
-      display: flex;
-      flex-direction: column;
-      gap: 0.5em;
-      overflow-y: auto;
-    }
-
     .reference-font-list {
       gap: 0;
+    }
+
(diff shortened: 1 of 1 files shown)


justvanrossum:
Thank you, I can confirm this PR fixes the issue.

> I was wondering if the layout becomes easier if the upload-list and the preview get joined. What do you think?

I'm not sure what you mean. I'd prefer to touch this code as little as possible, though.

justvanrossum (approved):

giovanniTramonto:
> > I was wondering if the layout becomes easier if the upload-list and the preview get joined. What do you think?
> 
> I'm not sure what you mean. I'd prefer to touch this code as little as possible, though.

At the moment there is the name in the list and a preview for each cut. I was wondering if those could be combined to one section, including both: preview and name in one item. Just an idea, for later perhaps.

PR #2165, opened 2025-05-16
Attempt to fix Safari web worker

### LINKED ISSUE #2156: Reference Font no longer works in Safari, since js-split
One more js-split regression: the Reference Font panel no longer works in Safari, due to its need for a Web Worker to write data to the Origin Private File System.

According to https://webpack.js.org/guides/web-workers/ I should be able to fix it like this:

```diff
diff --git a/src-js/fontra-core/src/opfs.js b/src-js/fontra-core/src/opfs.js
index 9b7c11847..6e4cff8c0 100644
--- a/src-js/fontra-core/src/opfs.js
+++ b/src-js/fontra-core/src/opfs.js
@@ -147,8 +147,12 @@ let worker;
 
 function getWriteWorker() {
   if (!worker) {
-    const path = "/core/opfs-write-worker.js"; // + `?${Math.random()}`;
-    worker = new Worker(path);
+    worker = new Worker(
+      /* webpackChunkName: "opfs-write-worker" */ new URL(
+        "./opfs-write-worker.js",
+        import.meta.url
+      )
+    );
   }
   return worker;
 }
```

But I can't get it to work (with or without the `webpackChunkName` magic comment).

To reproduce: go to the Reference Font panel, and drop a ttf on the list field.

Cc @simoncozens

justvanrossum:
Attempting to fix #2156.

- Changed creation of web worker according to https://webpack.js.org/guides/web-workers/
- Added onerror/onmessageerror handlers to worker

It is still not working:
- in Safari, go to a fontra glyph editor
- go to the reference font panel
- drop any ttf onto the font list

I get this in the console:

```
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (fontra-core.e894e05f.js, line 0)
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (fontra-core.e894e05f.js, line 0)
[Log] error from web worker: – ErrorEvent {isTrusted: true, message: "NetworkError: Load failed", filename: "http://localhost:8000/js/opfs-write-worker.chunk.js", …}
ErrorEvent {isTrusted: true, message: "NetworkError: Load failed", filename: "http://localhost:8000/js/opfs-write-worker.chunk.js", lineno: 263, colno: 28, …}ErrorEvent
[Error] NetworkError: Load failed (opfs-write-worker.chunk.js, line 263)
[Error] Unhandled Promise Rejection: NetworkError: Load failed
```
As a screen grab:
image

The file `http://localhost:8000/js/opfs-write-worker.chunk.js` aka. `src/fontra/client/js/opfs-write-worker.chunk.js` exists.
However, it tries to load `http://localhost:8000/js/js/fontra-core.e894e05f.js` which does not: it has one `js/` too many in the URL. It almost sounds like a webpack bug.

The code in `src/fontra/client/js/opfs-write-worker.chunk.js` is quite incomprehensible (it's internal webpack helper code), and I can't see why it uses the wrong URL to load the chunk.

@simoncozens, any help would be greatly appreciated.

### DIFF
--- a/src-js/fontra-core/src/opfs.js
+++ b/src-js/fontra-core/src/opfs.js
@@ -147,8 +147,12 @@ let worker;
 
 function getWriteWorker() {
   if (!worker) {
-    const path = "/core/opfs-write-worker.js"; // + `?${Math.random()}`;
-    worker = new Worker(path);
+    worker = new Worker(
+      /* webpackChunkName: "opfs-write-worker" */ new URL(
+        "./opfs-write-worker.js",
+        import.meta.url
+      )
+    );
   }
   return worker;
 }
--- a/src-js/fontra-core/src/opfs.js
+++ b/src-js/fontra-core/src/opfs.js
@@ -168,6 +168,14 @@ async function writeFileToOPFSInWorker(path, file) {
           resolve(event.data.returnValue);
         }
       };
+      worker.onmessageerror = (event) => {
+        console.log("message error from web worker:", event);
+        reject(event.message);
+      };
+      worker.onerror = (event) => {
+        console.log("error from web worker:", event);
+        reject(event.message);
+      };
       worker.postMessage({ path, file });
     }),
     5000


justvanrossum:
I found a fix that seems to work: the addional "js/" was added in `webpack.config.cjs`. Fingers crossed it doesn't have negative side effects.

PR #2174, opened 2025-06-04
Don’t set context.direction to undefined

### LINKED ISSUE #2171: New warning in Chrome: The provided value 'undefined' is not a valid enum value of type CanvasDirection.
```
visualization-layer-definitions.js:163 The provided value 'undefined' is not a valid enum value of type CanvasDirection.
```

We should probably fall back to `"inherit"` instead of `undefined`.

I'm assuming this is related to https://github.com/googlefonts/fontra/pull/2010, but I didn't propely investigate yet. Cc @khaledhosny.

khaledhosny:
If `guessGlyphPlaceholderString()` returns `undefined` for direction, keep `context.direction` unchanged.

Fixes https://github.com/googlefonts/fontra/issues/2171


justvanrossum (approved):
Fantastic, thank you!

PR #2188, opened 2025-06-27
Align the icon to the center

navv-1:
image


justvanrossum:
Oh, that's sloppy! Thanks for the fix. Ideally we should contribute this upstream, too: https://github.com/tabler/tabler-icons

PR #2205, opened 2025-07-03
Allows kerning jumps from 5 to 5 units

GaetanBaehr:
A pull request to allows jumping from 5 to 5 units when hold alt key and dragging the selector, or press arrows keys

### DIFF
--- a/src-js/views-editor/src/edit-tools-kerning.js
+++ b/src-js/views-editor/src/edit-tools-kerning.js
@@ -167,10 +167,9 @@ export class KerningTool extends BaseTool {
         this.sceneController.scrollAdjustBehavior = this.getScrollAdjustBehavior();
 
         const currentX = event.x / magnification;
-        var deltaX = Math.round(currentX - initialX);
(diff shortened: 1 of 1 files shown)


GaetanBaehr:
@justvanrossum I reformated with an external `_factor` function to handle the 3 situations: alt / shift / alt+shift
and now the shift works when dragging the selector.

src-js/views-editor/src/edit-tools-kerning.js
```diff
@@ -532,6 +527,11 @@ class KerningHandle extends HTMLElement {
 
 customElements.define("kerning-handle", KerningHandle);
 
+function getFactor(event) {
+    const factor = event.altKey ? (event.shiftKey ? 50 : 5) : event.shiftKey ? 10 : 1;
```
justvanrossum: No need for a variable, just do `return event.altKey ? (...etc);`

justvanrossum:
I made a couple of minor tweaks, otherwise this is done. Thank you!

PR #2206, opened 2025-07-03
Placeholder for selected glyph

### LINKED ISSUE #2198: Placeholder in the canvas string to display the currently edited glyph in different places (/?)
It would be awesome to be able to type something like /? in the canvas string as a placeholder for the current edited glyph. 

This will helps a lot for the spacing.

GaetanBaehr:
This fixes #2198: replacing in the canvas all the "/?" from the text entry by the selected glyph.


GaetanBaehr:
@justvanrossum I just pushed an update that remove the `this.sceneSettings.selectedGlyph = null;` in the activation of the kerningTool to keep the selected glyph in the /? and allows kern it with context

justvanrossum:
> I just pushed an update that remove the this.sceneSettings.selectedGlyph = null; in the activation of the kerningTool to keep the selected glyph in the /? and allows kern it with context

Hm, I'm not sure we can just do that. There was a reason there can't be a selected glyph during editing kerning. This definitely needs more thinking.

There is another problem: if you use the "next/previous glyph" context menu, or use the related chars/glyph panel to insert a glyph, or use the glyph search panel to switch the selected glyph, the `/?` is replaced by the actual glyph. This can be solved, but will need some special handling in `textFromGlyphLines`.

GaetanBaehr:
@justvanrossum I pushed a fix for the "next/previous glyph", by adding a placeholder key in the glyphNames list

GaetanBaehr:
> Hm, I'm not sure we can just do that. There was a reason there can't be a selected glyph during editing kerning. This definitely needs more thinking.

what was the reason ?

justvanrossum:
The kern tool currently does not allow selecting a glyph. It has to do with undo: normally, a selected glyph can be edited (via the info panel), but kerning has its own undo stack. So there's a lot of potential for confusion. So maybe the kern tool indeed needs to be able to select a glyph (in addition to selecting kern pairs), but then perhaps we shouldn't allow it to be edited. Or that (glyph) undo does not work while the kern tool is active.

Like I said, it needs more thinking. I don't know the best solution right now, just that as it stands it cannot work.

GaetanBaehr:
Can't we add a key like `previousSelectedGlyph` and when no selected glyph the placeholder use the `previousSelectedGlyph` ?

justvanrossum:
> Can't we add a key like previousSelectedGlyph and when no selected glyph the placeholder use the previousSelectedGlyph ?

That's worth trying!

GaetanBaehr:
> > Can't we add a key like previousSelectedGlyph and when no selected glyph the placeholder use the previousSelectedGlyph ?
> 
> That's worth trying!

I just pushed it, it looks to work well

justvanrossum:
> I just pushed it, it looks to work well

It's something, but it is not quite pretty: it disappears when reloading, ie. it doesn't get stored in the URL.

Like I said, it really needs more thinking, and I really need some time for that.

justvanrossum:
Like, the whole `/?` concept works well if the editor view is separate from the spacing/kerning view (like in RoboFont), but that's simply not what we have here.

GaetanBaehr:
> Like, the whole `/?` concept works well if the editor view is separate from the spacing/kerning view (like in RoboFont), but that's simply not what we have here.

Why is it not working well in our context? I'm using this pull request for spacing now, and it helps a lot

justvanrossum:
> Why is it not working well in our context? I'm using this pull request for spacing now, and it helps a lot

I'm not saying it doesn't help you or doesn't work as you intended, but it does not work for the edge cases I mentioned above. It has messy behavior and the concept deserves a better solution. And like I said many times now, I need more time to think through the implications and possible better solutions.

GaetanBaehr:
> > Why is it not working well in our context? I'm using this pull request for spacing now, and it helps a lot
> 
> I'm not saying it doesn't help you or doesn't work as you intended, but it does not work for the edge cases I mentioned above. It has messy behavior and the concept deserves a better solution. And like I said many times now, I need more time to think through the implications and possible better solutions.

Okay, I was not trying to push you, just trying to get a better understanding of all the implications behind this feature

justvanrossum:
I think I know how I want to finish this. I'll work on it after #2213 is done.

Issue #2106, opened 2025-03-20
Shortcut for comment/uncomment


justvanrossum:

justvanrossum:
Fixed by #2212.

Issue #2101, opened 2025-03-20
[OpenType Feature Editor] Color coding for feature code syntaxes


ollimeier:
As a follow up to the OpenType Feature Code Editor, we may want to add color coding for the code syntaxes.

ollimeier:
Here is a first draft of the code, which may help:
```
import { autocompletion } from "@codemirror/autocomplete";
import { StreamLanguage } from "@codemirror/language";
import { recordChanges } from "@fontra/core/change-recorder.js";
import * as html from "@fontra/core/html-utils.js";
import { addStyleSheet } from "@fontra/core/html-utils.js";
import { basicSetup, EditorView } from "codemirror";
import { BaseInfoPanel } from "./panel-base.js";

addStyleSheet(`
.font-info-opentype-feature-code-container {
  background-color: var(--ui-element-background-color);
  border-radius: 0.5em;
  padding: 1em;
}

.font-info-opentype-feature-code-header {
  font-weight: bold;
  padding-bottom: 1em;
}

#font-info-opentype-feature-code-text-entry-textarea {
  font-size: 1.1em;
}

`);

export class OpenTypeFeatureCodePanel extends BaseInfoPanel {
  static title = "opentype-feature-code.title";
  static id = "opentype-feature-code-panel";
  static fontAttributes = ["features"];

  async setupUI() {
    const features = await this.fontController.getFeatures();
    this.panelElement.innerHTML = "";
    const container = html.div(
      { class: "font-info-opentype-feature-code-container" },
      []
    );
    container.appendChild(
      html.div({ class: "font-info-opentype-feature-code-header" }, [
        "OpenType Feature Code",
      ])
    );
    const editorContainer = html.div(
      { id: "font-info-opentype-feature-code-text-entry-textarea" },
      []
    );

    const view = new EditorView({
      doc: features.text,
      extensions: [
        basicSetup,
        OpenTypeFeatureLanguage,
        autocompletion({ override: [myCompletions] }),
        EditorView.updateListener.of((update) => {
          if (update.docChanged) {
            const featureCodeText = update.state.doc.toString();
            this.editOpenTypeFeatureCode((root) => {
              root.features.text = featureCodeText;
            }, `edit OpenType feature code`); // TODO: translation
          }
        }),
      ],
      parent: editorContainer,
    });

    container.appendChild(editorContainer);
    this.panelElement.appendChild(container);
  }

  async editOpenTypeFeatureCode(editFunc, undoLabel) {
    const root = {
[...]

PR #2208, opened 2025-07-04
Enhance OpenType feature code editor

### LINKED ISSUE #2186: Excessive scroll bars in opentype feature editor
[Chrome on Windows]
Image

navv-1:
Fixes #2186 #2106 #2101

Also:

- Fix undo/redo shortcuts

- Add indentWithTab support

- Improve styling

For color coding/syntax highlighting is using CodeMirror 5-style [stream parser](https://codemirror.net/docs/ref/#language.StreamParser) which may need to be upgraded for more advanced needs. But I think it works for now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,9 +18,9 @@
         "@babel/preset-env": "^7.27.2",
         "@babel/preset-typescript": "^7.27.1",
         "@codemirror/autocomplete": "^6.18.6",
-        "@codemirror/highlight": "^0.19.8",
-        "@codemirror/history": "^0.19.2",
         "@codemirror/language": "^6.11.2",
+        "@codemirror/legacy-modes": "^6.5.1",
+        "@lezer/highlight": "^1.2.1",
         "babel-loader": "^10.0.0",
         "codemirror": "^6.0.2"
       },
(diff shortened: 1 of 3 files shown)


justvanrossum:
Oh wow, thank you!

With respect to the undo/redo shortcut, I *thought* this used to work (but I see that it currently doesn't). I tried to connect it to Fontra's own user-customizable shortcut system. Perhaps a recent update to codemirror changed the behavior, and broke my attempt?

Ideally, the toggle comment shortcut should also be handled via Fontra's shortcut system.

justvanrossum:
Ouch, the mypy failure is obviously unrelated to your work. I'm looking into it.

justvanrossum:
I worked around the mypy problem on main, please rebase.

navv-1:
> With respect to the undo/redo shortcut, I _thought_ this used to work (but I see that it currently doesn't). I tried to connect it to Fontra's own user-customizable shortcut system. Perhaps a recent update to codemirror changed the behavior, and broke my attempt?
> 
> Ideally, the toggle comment shortcut should also be handled via Fontra's shortcut system.

Previously, the undo/redo shortcut in the OpenType feature editor was not working in my environment, so I tried to fix that. Unfortunately, it doesn't connect to Fontra's shortcut system. Similarly with the comment shortcut toggle. :(

justvanrossum:
> Previously, the undo/redo shortcut in the OpenType feature editor was not working in my environment

What I'm trying to say is, that it's also currently not working correctly for me, but that I'm fairly convinced that it did work in the past. It would be much nicer if we could first figure out why it broke (if it did), and then if possible fix it so it keeps the integration with Fontra short cuts. We can also decide to look into that later, separate from this PR.

justvanrossum:
Ok, I played around, and here's what I found:

- the `@codemirror/history` extension seems unmaintained and/or obsolete, we should probably remove it from package.json
- your change to use redo/undo/etc from `@codemirror/commands` is correct
- this for me immediately fixes the undo behavior AND shortcuts: Fontra's programmatic invocation of undo/redo now seems to work correctly.
- So, no need to specify undo/redo shortcuts in the `keymap.of` section as far as I can tell

So please try to remove that section and see if it works for you with the Fontra shortcut mechanism.

Let's leave the comment toggle, after this PR is merged I will try to change that to also use the Fontra shortcut mechanism.

navv-1:
When the toggle comment's and  redo's part removed from `keymap.of`,

- The toggle comment shortcut with Ctrl+/ still works

- Undo with Ctrl+Z works

- Redo with Ctrl+Shift+Z doesn't work, but with Ctrl+Y (codemirror default for windows)

And I try to change the shortcuts in Fontra > Shortcuts, but the new shortcuts are not working and still use the previous shortcuts (in edit glyph, it works as expected).

justvanrossum:
Hm, when I checkout this branch, the features panel hangs and shows nothing in Chrome and Firefox, but does work in Safari.

navv-1:
> Hm, when I checkout this branch, the features panel hangs and shows nothing in Chrome and Firefox, but does work in Safari.

I also get blank in the shortcuts panel. It seems I made mistakes when resolving conflict. I will make a new branch for this. My apologies.

Issue #2219, opened 2025-07-14
Align / snap guidelines


752986:
Often, I want to make one stroke match the angle of another stroke, so I create a guideline with the same angle. However, this is quite tedious, as I have to do some `atan2` math on the two endpoints to get the angle, and then line up the guideline with the stroke I want to change. I'd like some way of automatically making a guideline match the angle of a stroke, and a way to snap a guideline onto a stroke.

I'm willing to work on this change myself, although I haven't looked at the codebase yet so I'm not sure how challenging it would be. Are there any opinions on what the UX for this should look like?

justvanrossum:
Best to approach this indeed as two features:
- create a guideline from a selection of two points
- snap to guidelines while dragging points

The codebase is quite challenging, but I'd definitely welcome PRs.

justvanrossum:
Add guideline between two points has been implemented in #2226, snap to guideline functionality is tracked in #1369.

PR #2226, opened 2025-07-15
Add guideline between 2 points

752986:
When 2 points are selected, a new context menu item is enabled to create a guideline between them. This resolves the more important half of #2219.

Hopefully the way I'm getting the currently selected points is reasonable; I had some trouble navigating the codebase so maybe there's a better method hiding somewhere.

Besides translations, is anything else needed in order to consider this change complete?

### DIFF
--- a/src-js/views-editor/src/editor.js
+++ b/src-js/views-editor/src/editor.js
@@ -370,6 +370,13 @@ export class EditorController extends ViewController {
       );
 
       registerAction(
+        "action.add-guideline-between-points",
+        { topic },
+        () => this.doAddGuidelineBetweenPoints(),
+        () => this.canEditGlyph() && this.sceneController.selection.size == 2
+      );
+
+      registerAction(
         "action.lock-guideline",
         { topic },
         () => this.doLockGuideline(!this.selectionHasLockedGuidelines()),
@@ -1388,6 +1395,9 @@ export class EditorController extends ViewController {
     this.glyphEditContextMenuItems.push({ actionIdentifier: "action.add-component" });
     this.glyphEditContextMenuItems.push({ actionIdentifier: "action.add-anchor" });
     this.glyphEditContextMenuItems.push({ actionIdentifier: "action.add-guideline" });
+    this.glyphEditContextMenuItems.push({
+      actionIdentifier: "action.add-guideline-between-points",
+    });
 
     this.glyphEditContextMenuItems.push({ actionIdentifier: "action.lock-guideline" });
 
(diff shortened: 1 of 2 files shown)


src-js/views-editor/src/editor.js
```diff
@@ -2743,6 +2753,81 @@ export class EditorController extends ViewController {
...
+    this.visualizationLayersSettings.model["fontra.guidelines"] = true;
+    const {
+      point: pointSelection,
+      anchor: anchorSelection,
+      guideline: guidelineSelection,
+    } = parseSelection(this.sceneModel.selection);
+
+    const glyph = this.sceneModel.getSelectedPositionedGlyph().glyph;
+
+    let pointA;
```
justvanrossum, line 2768: I would prefer a more compact code pattern. Something along these lines:

```js
    const points = [];

    points.push(... pointSelection?.map((index) => glyph.path.getPoint(index)) || []);
    // similar for anchors and guides

    const [pointA, pointB] = points;
    if (!pointA || !pointB) {
      return;
    }
    // calc guideline
```

justvanrossum (commented):
Thank you! I left some comments.

752986:
Alright, I've added code to place the origin between the points when possible, but testing on some real glyphs, that doesn't seem to happen extremely often. I could explore giving it some tolerance, which would pretty much guarantee a centered point, but I imagine that would add a lot of complexity.

I'll have to use this version for a bit to see, but I wonder if the unpredictability in placement will be more annoying than it not being centered.

justvanrossum:
You're probably avoiding placing it on point A because that makes the guideline hard to select? Ideally guidelines should be selectable all along the guideline, but that's a change we should probably tackle separately.

justvanrossum:
I implemented "click anywhere on guideline, not just the anchor point" in https://github.com/googlefonts/fontra/pull/2234. So if you can rebase on main, you can play with it in the context of your work. I think it's a good improvement.

752986:
Awesome, that's a much better way to solve the problem. The one thing I notice is that even when selected, the guideline is still behind the path point, so it's hard to tell that it is selected. Perhaps the entire guideline should become darker when selected, so you can tell even when its origin is off screen.

justvanrossum:
> The one thing I notice is that even when selected, the guideline is still behind the path point, so it's hard to tell that it is selected. Perhaps the entire guideline should become darker when selected, so you can tell even when its origin is off screen.

Yeah, that's true. I briefly looked into that and decided that's "for later", because that code is more complex than it should be, but if you want to give it a go, feel free.

752986:
If I do work on that, it'll be in a separate pull request.

justvanrossum (approved):
Thank you!

PR #2227, opened 2025-07-20
Add the ability to zoom to a specific font size

### LINKED ISSUE #105: Give a custom font size value to the text in the canvas

752986:
This adds a slider in the text entry panel to pick a font size, and a button that zooms to match that size. There is also a checkbox to automatically apply that font size in clean view.

image

I currently have some issues and questions about the UI:
- Even though there is a tooltip set for the button, it doesn't appear when hovering.
- The slider can't be adjusted by keyboard input until after interacting with its text input.
- The unit of the input (pt) needs to be communicated to the user somehow.
- The apply button feels too far right, even with a modified `--label-column-width`.
- I used a ui-form, since I'm trying to match the "set then apply" pattern found in the transform panel, and that uses ui-form. Is this the right approach here, or should it use a bespoke setup instead?
- Thoughts on the icon? For now I'm using `resize` as a placeholder, but other possibilities include `text-size`, `zoom-scan`, `ruler-measure-2`, and `eye`. I'm also happy to draw a new icon if needed.

Right now the UI strings are hardcoded with `// TODO: translate` comments, similar to surrounding code. Would it be better to use the translations framework, even though I can't provide translated text for other languages?

Closes #105, contributes towards #2215.

### DIFF
--- a/src-js/fontra-webcomponents/src/ui-form.js
+++ b/src-js/fontra-webcomponents/src/ui-form.js
@@ -416,6 +416,9 @@ export class Form extends SimpleElement {
       };
     }
 
+    this._fieldGetters[fieldItem.key] = () => rangeElement.value;
+    this._fieldSetters[fieldItem.key] = (value) => (rangeElement.value = value);
+
     valueElement.appendChild(rangeElement);
   }
 
--- a/src-js/fontra-webcomponents/src/ui-form.js
+++ b/src-js/fontra-webcomponents/src/ui-form.js
@@ -389,6 +389,7 @@ export class Form extends SimpleElement {
     rangeElement.minValue = fieldItem.minValue;
     rangeElement.defaultValue = fieldItem.defaultValue;
     rangeElement.maxValue = fieldItem.maxValue;
+    rangeElement.step = fieldItem.step;
 
     {
       // Slider change closure
--- a/src-js/views-editor/src/editor.js
+++ b/src-js/views-editor/src/editor.js
@@ -21,6 +21,7 @@ import {
 } from "@fontra/core/path-functions.js";
 import {
   centeredRect,
+  pointInRect,
   rectAddMargin,
   rectCenter,
   rectFromArray,
@@ -3369,7 +3370,11 @@ export class EditorController extends ViewController {
     const desiredHeight = canvasHeight / pixel_size;
 
     const selBox = this.sceneController.getSelectionBounds();
-    const center = rectCenter(selBox || viewBox);
+    let center = rectCenter(selBox || viewBox);
+    // avoid going completely off-screen when zoomed in very close
+    if (currentHeight < 1.0 || !pointInRect(center.x, center.y, viewBox)) {
+      center = rectCenter(viewBox);
+    }
     viewBox = rectScaleAroundCenter(viewBox, desiredHeight / currentHeight, center);
     if (animate) {
       this.animateToViewBox(viewBox);
--- a/src-js/views-editor/src/editor.js
+++ b/src-js/views-editor/src/editor.js
@@ -3351,6 +3351,7 @@ export class EditorController extends ViewController {
   }
 
   zoomToFontSize(size, animate = true) {
+    // TODO: does this function need modifications to work with vertical layout?
     let viewBox = this.sceneSettings.viewBox;
     const upm = this.fontController.unitsPerEm;
     const canvasHeight = this.canvasController.canvasHeight;
@@ -3359,7 +3360,8 @@ export class EditorController extends ViewController {
     const currentHeight = (viewBox.yMax - viewBox.yMin) / upm;
     const desiredHeight = canvasHeight / pixel_size;
 
-    const center = rectCenter(viewBox);
+    const selBox = this.sceneController.getSelectionBounds();
+    const center = rectCenter(selBox || viewBox);
     viewBox = rectScaleAroundCenter(viewBox, desiredHeight / currentHeight, center);
     if (animate) {
       this.animateToViewBox(viewBox);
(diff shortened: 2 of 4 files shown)


justvanrossum:
Couple of quick comments:
- I would not use ui-form for this, just a range-slider. Probably set to non-continuous
- I'm not sure about the "use this size in clean view" feature. That seems to overcomplicate the code + UI. Can you say a bit more about why you want that?

752986:
- Sounds good, would you also like me to undo my changes to `ui-form.js` then? And when you say non-continuous, do you mean it should only have certain preset sizes?
- Often when I'm designing for small sizes, I want to make an edit, quickly see how it looks at the target size, and then go back to editing. Right now it's somewhat tedious to zoom out, check my work, and then zoom back in, so I don't check it as often as I'd like to. I suppose a keyboard shortcut for "apply font size" would take care of the first half, but zooming back in to where I was working would still be a bit annoying.

justvanrossum:
> Sounds good, would you also like me to undo my changes to ui-form.js then? 

Yes please.

> And when you say non-continuous, do you mean it should only have certain preset sizes?

Sorry, no, it means that I think the change should only occur after the user lets go of the slider thumb, and not continuously while sliding.

> Often when I'm designing for small sizes, I want to make an edit, quickly see how it looks at the target size, and then go back to editing. Right now it's somewhat tedious to zoom out, check my work, and then zoom back in, so I don't check it as often as I'd like to. I suppose a keyboard shortcut for "apply font size" would take care of the first half, but zooming back in to where I was working would still be a bit annoying.

Understood. However, I feel this is a separate feature, and there would be different ways to accomplish that. For example, Glyphs App has a "preview" area at the bottom that shows the current glyph, either in context or in different interpolations. In Fontra, you can have a separate tab of window open, and you could use the browser shortcuts to switch tabs (or windows). In the future Fontra may want to offer a more integrated solution (perhaps a la Glyphs), but right now I feel it's a bit too arbitrary too couple it to the font size UI.

752986:
After using this branch for a few days, I ended up not really using the clean view toggle anyway, and I agree that there are better ways to get that functionality. I do think a keyboard shortcut for zooming to the size would be useful though, so I'll implement that and try it out for a bit.

Another question: I'm using [Google Fonts' documentation](https://fonts.google.com/knowledge/glossary/optical_size_axis) for the min and max values; it also says that 14pt should be the default for `opsz`, but most websites and browsers seem to use 12pt as the default font size. Which one do you think is more relevant to preview here?

davelab6:
14 is because of the mix up with 12 & 16, pt and px, so 14 is in the middle ;)

justvanrossum:
Hi @752986, is there a chance you will continue this PR, or should we consider closing it?

752986:
Apologies, I was busy with school and forgot this was still in draft status. I am happy with the functionality so I will open it for review.

justvanrossum:
Can you also rebase to main?

justvanrossum:
Hm, you somehow applied all later commits to your branch instead of rebasing. Please try to undo that because all those commits do not belong here.

justvanrossum:
Thanks! I think I would like it if moving the slider or entering a value would trigger a zoom, so the user doesn't have to click the resize button:
```diff
diff --git a/src-js/views-editor/src/panel-text-entry.js b/src-js/views-editor/src/panel-text-entry.js
index b03b26c85..c05fed8db 100644
--- a/src-js/views-editor/src/panel-text-entry.js
+++ b/src-js/views-editor/src/panel-text-entry.js
@@ -191,6 +191,7 @@ export default class TextEntryPanel extends Panel {
       onChangeCallback: (event) => {
         if (!event.isDragging) {
           this.textSettings.textSize = event.value;
+          this.editorController.zoomToFontSize(this.textSettings.textSize);
         }
       },
     });
```

I can even imagine omitting the `isDragging` condition, that would make the slider live.

What would you think about that?

Then, I think it would be nice to see in the UI whether the current zoom matches the explicit size or not. Perhaps the button could be disables/grayed-out if the zoom already matches.

752986:
I do like the zoom updating on drag; I personally think only doing it on drag end is better but I don't have a strong opinion.

justvanrossum:
On drag end is fine with me, too.

justvanrossum:
Just in case you want to continue with this, a couple of points of feedback:

- I think the [above suggestion](https://github.com/fontra/fontra/pull/2227#issuecomment-3562189536) should be applied
- There is a bit of inconsistency in naming: sometimes it's "text size" other times it is "font size". I prefer "text size"
- I would prefer to use the "text-size" icon from the [Tabler icon set](https://tabler.io/icons). (I have pasted the svg below)
- Should there be a menu item for "zoom to text size" under View? Probably.
- We need to update the [strings sheet](https://docs.google.com/spreadsheets/d/1woTU8dZCHJh7yvdk-N1kgQBUj4Sn3SdRsbKgn6ltJQs/edit?gid=1383907145#gid=1383907145) with an entry for `menubar.view.zoom-to-font-size` and then regenerate the language strings files. I can do that. Btw. the key is misspelled as `meunbar.view.zoom-to-font-size`.

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="icon icon-tabler icons-tabler-outline icon-tabler-text-size"><path stroke="none" d="M0 0h24v24H0z" fill="none"/><path d="M3 7v-2h13v2" /><path d="M10 5v14" /><path d="M12 19h-4" /><path d="M15 13v-1h6v1" /><path d="M18 12v7" /><path d="M17 19h2" /></svg>
```

justvanrossum:
Closing due to lack of activity and staleness.

PR #2280, opened 2025-09-26
Fix a minor alignment issue in ui-form-value

navv-1:
Make ui-form-value have the same height as ui-form-label


justvanrossum:
Where/how can I see the effect of this change?

navv-1:
In Chrome, if the glyph name is a lowercase letter, it will look lower than expected.
image

justvanrossum:
Thanks, I see it now!

PR #2320, opened 2025-11-26
Keep clean view active by removing keyup for meta key

GaetanBaehr:
This is a feature I used a lot, and I really miss it now. It allowed me to keep a clean view by pressing the Cmd modifier before releasing my clean-view shortcut key. This gave me a global clean overview without having to keep a key pressed, allowed me to navigate the canvas without holding the clean-view key, and also let me continue editing glyph shapes while staying in clean-view mode. I’d love to revert this change and get back the ability to lock the clean view.

### DIFF
--- a/src-js/views-editor/src/editor.js
+++ b/src-js/views-editor/src/editor.js
@@ -3148,7 +3148,7 @@ export class EditorController extends ViewController {
       // receive a keyup event for the space key. So let's also respond to any
       // keyup for the Meta key.
       // Oddly, event.metaKey is false at keyup, so we check event.key instead.
-      (this._matchingKeyUpHandler.code == event.code)
+      this._matchingKeyUpHandler.code == event.code
     ) {
       this._matchingKeyUpHandler.callback(event);
       delete this._matchingKeyUpHandler;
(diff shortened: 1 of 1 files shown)


justvanrossum:
You are pretty much undoing https://github.com/fontra/fontra/pull/2270 (without removing the matching comment, which is now stale), but this was added because the old behavior was accidental and confusing, and I consider it a bug, not a feature. It happened to me a lot, and it felt like "clean view" was randomly stuck.

So, I'm not very inclined to put it back I'm afraid.

GaetanBaehr:
Or could we add it in another way, maybe with another shortcut, so that we can keep the clean view locked? During the design process, it’s very helpful to be able to lock the clean view, and I feel frustrated when I’m designing and end up losing it.

justvanrossum:
I am open to aadding a different shortcut to go in and out of clean view. Do make sure that the two ways work peacefully together. It could even be just menu item under "View", using a new registered action, with no default shortcut.

GaetanBaehr:
But if it’s a menu item, it won’t let us keep using the pointer tool or the pen tool while Clean View is locked.

justvanrossum:
Ok, without a shortcut it is less useful. But do please make it a menu item anyway, so users can discover the shortcut.

GaetanBaehr:
okay I'll try

justvanrossum:
Please open a new PR if you come around to implementing this.

Issue #1864, opened 2024-12-16
Tiny outline for handles + off-curve nodes


ollimeier:
From @GaetanBaehr:

with a grey background image, I don't really see my offcurve on the left, and I actually don't see at all the one on the right, looks like it's behind the image or something

![Image](https://github.com/user-attachments/assets/8edd151c-31a3-4e9d-be79-8cce3111cf47)

Note from @justvanrossum:

Yeah, the on-curve regular nodes have a tiny outline, we should probably do the same with handles + off-curve nodes.

justvanrossum:
I think the "underlay" outline is only done for selected on-curve nodes. Perhaps this should be done for everything.

PR #2400, opened 2026-01-27
Glyph editor: improve node visibility

jdhoek:
When tracing glyphs by setting the background to, for example, scans of old specimen books, the colourise feature is already extremely helpful to help distinguish the sample from the glyph and its outline. However, I often have trouble spotting the gray nodes and handles, making editing a more vexing experience than necessary.

In this branch I am experimenting with making the nodes more visible. I am taking inspiration from Inkscape, but the approach used seems conventional for vector editing tools in general.

I don't expect this PR to be ready for merging without more experimenting and feedback, but I figured it would provide a good starting point for the issues this addresses. Please have a look.

## Changes introduced

A double border of alternating black and white is added to each node. With this it no longer matters what colour the background has, because if it matches one border (rendering it nearly invisible) it will be contrasted by the other.

This is the current rendering:

Screenshot from 2026-01-27 10-48-03

This is the rendering on this branch:

Screenshot from 2026-01-27 11-12-58

The second commit changes the shape of the nodes to further distinguish handles from nodes. The handles remain round, but the smooth nodes are now square and the angled nodes diamond shaped. This follows the convention used by Inkscape, which seems natural to me, but another approach could work. My primary goal here is to remove the ambiguity between smooth nodes and handles.

## Questions

* The nodes render below the glyph outline. This is more visible in this branch than on `main`, but the same thing happens there. Is this intentional?

## Related issues

* Possibly related to #1864.

### DIFF
--- a/src-js/views-editor/src/visualization-layer-definitions.js
+++ b/src-js/views-editor/src/visualization-layer-definitions.js
@@ -1679,12 +1679,23 @@ export const allGlyphsCleanVisualizationLayerDefinition = {
 
 function fillNode(context, pt, cornerNodeSize, smoothNodeSize, handleNodeSize) {
   if (!pt.type && !pt.smooth) {
-    fillSquareNode(context, pt, cornerNodeSize);
+    fillDiamondNode(context, pt, cornerNodeSize);
   } else if (!pt.type) {
-    fillRoundNode(context, pt, smoothNodeSize);
+    fillSquareNode(context, pt, smoothNodeSize);
   } else {
     fillRoundNode(context, pt, handleNodeSize);
   }
+}
+
+function fillDiamondNode(context, pt, nodeSize) {
+  const offset = nodeSize / 1.4;
+  context.beginPath();
+  context.moveTo(pt.x, pt.y - offset);
+  context.lineTo(pt.x + offset, pt.y);
+  context.lineTo(pt.x, pt.y + offset);
+  context.lineTo(pt.x - offset, pt.y);
+  context.closePath();
+  context.fill();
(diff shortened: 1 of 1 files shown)


jdhoek:
For reference; this is how Inkscape looks (in dark mode):

Screenshot from 2026-01-27 12-13-59

justvanrossum:
While I appreciate the feedback regarding visibility, I'm currently not accepting big visual changes in this area. Also, I have no desire to make Fontra's nodes look more similar to Inkscape, as Inkscape is just one tool out of many, and they all look and behave differently.

So, the diamonds are right out, and I need to think about the additional outlines around nodes, as it does add more noise in general. Can you suggest other (perhaps less intrusive) approaches to improve the visibility in the context of background images?

jdhoek:
> Can you suggest other (perhaps less intrusive) approaches to improve the visibility in the context of background images?

Nothing comes to mind. Editors used to XOR elements which needed to contrast with the background, but that approach has been largely abandoned for either the double border one, or a wider single border in combination with more saturated colours.

> […] as it does add more noise in general.

While noise should be avoided, a lack of signal comes with its own drawbacks.

mitradranirban:
FontForge uses various colors and shapes to indicate various types of node, which is definitely helpful for novice people working with bezier curves 
Screenshot from 2026-02-15 14-49-10

scutuatua-crypto (dismissed):

davelab6:
@justvanrossum i wonder if this is something that should be theme'able, afaik RoboFont is super theme'able with this stuff :)

PR #2540, opened 2026-04-04
Security: Untrusted SVG content inserted into DOM without sanitization

tuanaiseo:
## Problem

`InlineSVG` fetches SVG text and converts/inserts it into the DOM. If `src` is attacker-controlled or can be redirected to untrusted content, malicious SVG payloads (event handlers, scriptable elements, external references) may lead to XSS or data exfiltration.

**Severity**: `high`
**File**: `src-js/fontra-webcomponents/src/inline-svg.js`

## Solution

Restrict SVG sources to trusted same-origin paths, enforce a strict CSP, and sanitize SVG before DOM insertion (remove scripts, event attributes, `foreignObject`, external hrefs). Prefer rendering via safe `` when interactivity is not needed.

## Changes

- `src-js/fontra-webcomponents/src/inline-svg.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/src-js/fontra-webcomponents/src/inline-svg.js
+++ b/src-js/fontra-webcomponents/src/inline-svg.js
@@ -33,22 +33,66 @@ export class InlineSVG extends HTMLElement {
   }
 
   async fetchSVG(svgSRC) {
-    const svgElement = htmlToElement(await cachedSVGData(svgSRC));
-    svgElement.removeAttribute("width");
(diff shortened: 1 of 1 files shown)


justvanrossum:
Any sanitation should be done before caching.

PR #2555, opened 2026-04-27
Migrate to harfbuzzjs v1.x API

khaledhosny:


src-js/fontra-core/src/webpack-base.js
```diff
@@ -12,6 +12,7 @@ const fallback = {
   stream: false,
   path: false,
   url: false,
+  module: false,
```
justvanrossum, line 15: Do you remember why this is necessary? (Not contending, just curious.)

justvanrossum (approved):
Fantastic, thank you! This looks good.

Shall I merge it as it is, or would be be better to wait for a non-beta harfbuzz-js release? I'm okay with merging now.

src-js/fontra-core/src/webpack-base.js
khaledhosny: The emscripten generated `harfbuzz.js` has an `import("node:module")` (inside an `if (ENVIRONMENT_IS_NODE)` block), and without this line the webpack will complain it can’t find this module and fail.

khaledhosny:
> Fantastic, thank you! This looks good.
> 
> Shall I merge it as it is, or would be be better to wait for a non-beta harfbuzz-js release? I'm okay with merging now.

I think merging now is fine, the API is unlikely to change significantly and we avoid potential merge conflicts if we keep this open for long.

PR #2565, opened 2026-04-30
Center selection when jumping to feature code link

khaledhosny:
The `scrollIntoView: true` is equivalent EditorView.scrollIntoView() with default options which is `y: "nearest"`, but this would often scroll to the selection just enough to make the line appear at the end of the view, which is not very helpful.

Using EditorView.scrollIntoView() so we can set `y: "center"`, which is nicer as the start of the span becomes in the middle of the page.


justvanrossum (approved):
Nice, thanks!

PR #2566, opened 2026-04-30
Improve message formatting in feature debugger a bit

khaledhosny:
* Show lookup name in the Message column after the lookup number, not in the Feature source
* Re-order the message to have feature first, so they align better since lookup name vary in length
* Drop `<feature>:` part for Feature source for shaper fonts, it is a dummy file name and it occupies precious space.
* Make the Message column a bit wider since the Feature source column does not need much space now.

Before:
image

After:
image

### DIFF
--- a/src-js/views-editor/src/panel-characters-glyphs.js
+++ b/src-js/views-editor/src/panel-characters-glyphs.js
@@ -253,7 +253,7 @@ export default class CharactersGlyphsPanel extends Panel {
                 href: opentypeFeaturesURL,
                 target: `fontra.fontinfo.${this.editorController.projectIdentifier}`,
               },
-              [item.sourceLocation]
+              [`${lineno}:${column}`]
             );
           } else {
             return item.sourceLocation?.split(/\\|\//).at(-1);
--- a/src-js/views-editor/src/panel-characters-glyphs.js
+++ b/src-js/views-editor/src/panel-characters-glyphs.js
@@ -233,7 +233,7 @@ export default class CharactersGlyphsPanel extends Panel {
       {
         key: "formattedMessage",
         title: "Message",
-        width: 200,
+        width: 250,
         minWidth: 100,
       },
       {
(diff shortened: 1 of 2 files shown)


khaledhosny:
May be “Feature source” can better be “Source location”?

justvanrossum:
> May be “Feature source” can better be “Source location”?

Agreed.

justvanrossum (approved):
Nice, thanks!

src-js/views-editor/src/panel-characters-glyphs.js
```diff
@@ -253,7 +253,7 @@ export default class CharactersGlyphsPanel extends Panel {
                 href: opentypeFeaturesURL,
                 target: `fontra.fontinfo.${this.editorController.projectIdentifier}`,
               },
-              [item.sourceLocation]
+              [`${lineno}:${column}`]
```
justvanrossum, line 256: While I agree that removing the `<features>` string is an improvement, I think we should leave the filename in if it isn't `<features>`.

khaledhosny: This is inside a block that first matches against `/^<features>:(?<lineno>\d+):(?<column>\d+)/`, or did I misread the code?

khaledhosny:
> > May be “Feature source” can better be “Source location”?
> 
> Agreed.

Done.

src-js/views-editor/src/panel-characters-glyphs.js
justvanrossum: The intention there is that we will only try to *navigate* to the source if it matches `<features>:lineno:col`.  But if that doesn't match, it should still show the entire `item.sourceLocation`, so people can still use that info in other ways.

justvanrossum: Oh nevermind, let me read the code once more. I missed that `else` there.

justvanrossum: You're right, your change is good as it is. Sorry!

PR #2567, opened 2026-05-01
Fix canvas interpolation issue

qwerasd205:
I noticed an issue in Safari (26.4) where the center column of pixels on the canvas was being repeated, causing jank right in the middle of the working area.

|the issue|highlighted|
|-|-|
|image|image|

To reproduce this in Safari, use a high DPI display and make sure the canvas is more than 2K pixels wide, then fiddle with the width of one of the sidebars. 50% of sidebar widths will cause this problem.

The problem is revealed by Safari's behavior of switching to nearest neighbor interpolation for scaling the canvas texture when it's more than 2K px on an axis, but the underlying issue is that almost always the canvas texture has to be scaled (stretched!) a tiny bit, if the sidebars are out and have been adjusted in size at all. It's not usually noticeable though it does degrade the quality a tiny bit.

The root cause is that the canvas texture size was being computed as `floor(<width|height> * scale)`, which, when `width`/`height` or `scale` are not integers will half the time result in the texture being slightly narrower than the actual pixel size the element ends up being displayed at, meaning it has to be stretched by a pixel.

Rather than trying to determine what size the element will be, the solution I think is best (and so implemented) is to resize the canvas element to explicitly be the appropriate number of CSS pixels so that it will match the texture size exactly when scaled to screen pixels and rendered by the browser.

### DIFF
--- a/src-js/fontra-core/src/canvas-controller.js
+++ b/src-js/fontra-core/src/canvas-controller.js
@@ -92,10 +92,33 @@ export class CanvasController {
     const width = this.canvasWidth;
     const height = this.canvasHeight;
     const scale = this.devicePixelRatio;
-    this.canvas.width = Math.floor(width * scale);
-    this.canvas.height = Math.floor(height * scale);
-    this.canvas.style.width = width + "px";
-    this.canvas.style.height = height + "px";
+
+    // We want to be sure that we have at least enough pixels to cover the
+    // entire area of the canvas, accounting for dpi and browser scaling.
+    //
+    // To do this, we take the ceiling of the CSS pixel width and height
+    // values multiplied by the pixel ratio, which is a number that should
+    // be guaranteed to be equal to or greater than the final number of screen
+    // pixels that the canvas element (or rather its container) takes up.
+    this.canvas.width = Math.ceil(width * scale);
+    this.canvas.height = Math.ceil(height * scale);
+
+    // We then take those numbers and divide them back down by the scale
+    // and tell the browser to display the canvas at that many CSS pixels
+    // wide and tall. This should ensure that there is a 1 to 1 mapping
+    // between pixels in the canvas's data buffer and pixels as displayed
+    // on screen for the user.
+    //
+    // If we didn't do this, and just used the container's size, then the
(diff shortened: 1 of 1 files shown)


justvanrossum:
Amazing analysis, thanks! Will test soon, and likely merge as-is.

justvanrossum (approved):

justvanrossum:
I don't have a large enough display to reproduce and test the fix, but your logic seems solid, and the code looks good.

qwerasd205:
Unfortunately it looks like this bug somehow still exists, despite this fix. I'm not sure if it's an unfortunate coincidence that something changed WRT canvas scale rounding in Safari 26.5 (as I have just updated my OS before noticing the bug again) or if the fix didn't work completely to begin with. Regardless, I'll have to do some experimentation to figure out how to fix it more definitively.

qwerasd205:
I did some more research in to the problem and unfortunately it seems like there's currently *no* good solution in Safari.

The solution created by the web standards people for ensuring a pixel-perfect match between canvas backing size and canvas element size is to use a `ResizeObserver` and access the `devicePixelContentBoxSize` property to get device pixels directly for the element with no extra calculations or rounding necessary on our part. The problem with that is that Safari is the one major browser that hasn't implemented it yet. (Every other browser did in 2020/2021 ...)

Any solution that tries to account for how things will be rounded depending on the devicePixelRatio is bound to be empirical and prone to breaking when layout/paint algorithms change even slightly. 

References:
- WebKit bug tracking the (lack of) implementation of `devicePixelContentBoxSize`: https://www2.webkit.org/show_bug.cgi?id=219005
- Webkit draft PR that implements it: https://github.com/WebKit/WebKit/pull/60534 -- if this merges some time soon this may become a non-problem for us in like... a year or so. It's been open without activity for a couple months now though.

---

I do have another idea though-- theoretically no stretching should ever happen if we just let the canvas use its intrinsic content size instead of sizing it with CSS. So if we remove the CSS size assignment and calculate a size that should theoretically *always* be at least large enough (a couple pixels of overflow is acceptable, the container can be made to crop the rendering so it won't be seen) then this will never happen.

I'll explore that idea later today.

PR #2571, opened 2026-05-04
[shaping] Make dumb shaper dumber

### LINKED ISSUE #2517: There should be an option to disable any form of text shaping
Sometimes it is desired to have the glyphs just set next to each other each taking its full advance width (e.g. when editing a sequence of marks), but unchecking “Apply text shaping and feature” still leaves emulated features enabled, and even after unchecking these, combining marks still have there advances zeroed. Maybe unchecking “Apply text shaping and feature” should do this, or may be another option is needed if the current behavior is desired.

justvanrossum:
Make dumb shaper dumber: we don't want it to do anything beyond laying out the characters sequentially

This fixes #2517.

This disables all emulated GPOS functionality when the "Apply text shaping and features" option is off.

### DIFF
--- a/src-js/fontra-core/src/shaper.js
+++ b/src-js/fontra-core/src/shaper.js
@@ -536,14 +536,13 @@ class DumbShaper extends ShaperBase {
     for (const [i, codePoint] of enumerate(codePoints)) {
       const glyphName = this.nominalGlyph(codePoint);
       const xAdvance = Math.round(glyphObjects[glyphName]?.xAdvance ?? 500);
-      const isMark = this.isGlyphMarkFunc(glyphName);
 
       glyphs.push({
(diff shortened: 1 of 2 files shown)


justvanrossum:
@khaledhosny I was doubting between this new behavior and having three behaviors:

1. no shaping, no features, don't emulate features (behavior after this PR when "Apply text shaping and features" is *off*)
2. no shaping, no features, do emulate features (behavior before this PR when "Apply text shaping and features" is *off*)
3. shaping, features, do emulate features ("Apply text shaping and features" is *on*)

But ultimately decided to only offer 1 and 3, as I can't really find a good use case for option 2 that warrants complicating the UI more.

Do you find this a reasonable solution?

khaledhosny:
I agree that there is no much use for option 2.

PR #2574, opened 2026-05-04
canvas: make min zoom limit relative to em size (UPM)

### LINKED ISSUE #2573: Minimum canvas zoom (zooming out) should be relative to UPM
Currently the minimum zoom (maximum zoom *out*) is fixed relative to font units, which means the effective point size you can zoom out to on the canvas is limited by the UPM of the font. This zoom-out limit should be relative to the em size so that zooming out all the way yields a consistent point size across UPM values.

At 1000 UPM, when zoomed all the way out 1em = 10px, but at 2000 UPM, when zoomed all the way out, 1em = 20px. It's nice for me to be able to zoom out far enough to be able to quickly proof the font I'm working on at 12px but at any more than 1200 UPM I can't do that.

This fix is probably trivial, so I'll get started on a PR for it.

---

As an aside, it would be nice if there were a display that showed the current point (or pixel) size based on the zoom level and UPM, and a way to zoom to a specific point (or pixel) size. I may look in to contributing that, see how easy it is to implement. Also... user-configurable zoom limits maybe? Just throwing ideas out there.

qwerasd205:
Closes #2573 

I'm not sure whether you'll be okay with how I chose to implement this. I went with the way that gave the smallest set of changes, but it does slightly muddle the responsibilities of the `CanvasController` by making it aware of the UPM, and the way it's provided that info is by being passed a function (rather than having other parts of the code call *in* to it to update it on the current UPM whenever that changes, or having it be completely unaware of UPM and have someone else calculate and update the magnification limits for it whenever the UPM changes).

### DIFF
--- a/src-js/fontra-core/src/canvas-controller.js
+++ b/src-js/fontra-core/src/canvas-controller.js
@@ -1,5 +1,5 @@
 import { normalizeRect, rectCenter, validateRect } from "./rectangle.js";
-import { assert, consolidateCalls, isNumber, withSavedState } from "./utils.js";
+import { assert, clamp, consolidateCalls, isNumber, withSavedState } from "./utils.js";
 
 const DEFAULT_MIN_MAGNIFICATION = 0.005;
 const DEFAULT_MAX_MAGNIFICATION = 200;
@@ -211,13 +211,14 @@ export class CanvasController {
   }
 
   _clampMagnification() {
-    const old_magnification = this.magnification;
-    this.magnification = Math.min(
-      Math.max(this.magnification, this._minMagnification),
+    const oldMagnification = this.magnification;
+    this.magnification = clamp(
+      this.magnification,
+      this._minMagnification,
       this._maxMagnification
     );
 
-    if (this.magnification !== old_magnification) {
+    if (this.magnification !== oldMagnification) {
       this._magnificationChangedCallback?.(this.magnification);
       this.requestUpdate();
       this._dispatchEvent("viewBoxChanged", "magnification");
@@ -260,8 +261,9 @@ export class CanvasController {
     const prevMagnification = this.magnification;
 
     this.magnification = this.magnification * zoomFactor;
-    this.magnification = Math.min(
-      Math.max(this.magnification, this.minMagnification),
+    this.magnification = clamp(
+      this.magnification,
+      this.minMagnification,
       this.maxMagnification
     );
     zoomFactor = this.magnification / prevMagnification;
--- a/src-js/fontra-core/src/canvas-controller.js
+++ b/src-js/fontra-core/src/canvas-controller.js
@@ -1,17 +1,36 @@
 import { normalizeRect, rectCenter, validateRect } from "./rectangle.js";
 import { assert, consolidateCalls, isNumber, withSavedState } from "./utils.js";
 
-const MIN_MAGNIFICATION = 0.005;
-const MAX_MAGNIFICATION = 200;
+// Minimum pixels per em and maximum pixels per unit for zooming out and in.
+//
+// Note that these are not _screen pixels_, they are css "pixels" which are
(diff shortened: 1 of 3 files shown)


justvanrossum:
Yeah, hmm, I'm indeed not quite comfortable with CanvasController having font knowledge.

Since you're already adding getters, perhaps we should add setters for the magnification bounds, and set them from the SceneController, which can also subscribe to changes of the unitsPerEm value:

```js
    this.fontController.addChangeListener({ unitsPerEm: null }, (change) =>
      this.setCanvasMagnificationLimits();
    );
```

(Probably best at the end of `setupSceneSettings()`)

qwerasd205:
Alright, I moved the responsibility for calculating (and recalculating) magnification limits to the `SceneController`.

src-js/fontra-core/src/canvas-controller.js
```diff
@@ -1,18 +1,21 @@
...
 import { assert, consolidateCalls, isNumber, withSavedState } from "./utils.js";
 
-const MIN_MAGNIFICATION = 0.005;
-const MAX_MAGNIFICATION = 200;
+const DEFAULT_MIN_MAGNIFICATION = 0.005;
+const DEFAULT_MAX_MAGNIFICATION = 200;
 
 export class CanvasController {
-  constructor(canvas, magnificationChangedCallback) {
+  constructor(canvas, magnificationChangedCallback, getUnitsPerEm) {
```
justvanrossum: This argument is no longer needed, right?

src-js/fontra-core/src/canvas-controller.js
```diff
@@ -207,15 +210,57 @@ export class CanvasController {
     delete this._initialMagnification;
   }
 
+  set minMagnification(newValue) {
+    // If there's no change then we don't have to do anything.
+    if (newValue == this._minMagnification) return;
```
justvanrossum: In the Fontra code base, I alwas use explicit blocks:

```suggestion
    if (newValue == this._minMagnification) {
      return;
    }
```

src-js/fontra-core/src/canvas-controller.js
```diff
@@ -207,15 +210,57 @@ export class CanvasController {
...
+
+    this._minMagnification = newValue;
+
+    // If our magnification is within the new bound then we're done.
```
justvanrossum: Please refactor this section as a method to be shared between `set minMagnification()` and `set maxMagnification()`

src-js/fontra-core/src/canvas-controller.js
qwerasd205: Ah yeah, forgot about that, will fix.

src-js/fontra-core/src/canvas-controller.js
qwerasd205: So it should just clamp to both bounds regardless of which bound is being changed?

justvanrossum: Yeah, that seems easiest.

qwerasd205:
Addressed the review feedback

src-js/fontra-core/src/canvas-controller.js
```diff
@@ -207,15 +210,59 @@ export class CanvasController {
     delete this._initialMagnification;
   }
 
+  _clampMagnification() {
+    const old_magnification = this.magnification;
```
justvanrossum: Please use `oldMagnification` camelCase.

src-js/fontra-core/src/canvas-controller.js
```diff
@@ -207,15 +210,59 @@ export class CanvasController {
...
+    this.magnification = Math.min(
```
justvanrossum: `utils.js` contains a `clamp()` function.

justvanrossum (commented):
Two small nits.

justvanrossum (approved):
Thanks!

PR #2576, opened 2026-05-04
Selectable text size

qwerasd205:
This is implemented on top of #2574 so had its commits in it-- I believe the GitHub UI will update it appropriately for better review once the other PR is merged.

This is a first pass at #105 -- it adds two inputs to the text entry panel, one for setting the text size in points and the other for setting it in pixels.

These values automatically update as the user zooms in other ways, and when modified by the user they directly change the magnification value of the canvas, which I'm sure is the wrong way to go about it, and can be a little jarring because it's not animated at all.

The UI also isn't very pretty. This is why I called it a "first pass", since it needs cleaning up, but I think the foundation of having live display of the current size in pt and px that can be updated by the user to zoom to that level is good.

https://github.com/user-attachments/assets/eabd61dd-9153-41b2-82ee-258264fe708d

### DIFF
--- a/src-js/views-editor/src/panel-text-entry.js
+++ b/src-js/views-editor/src/panel-text-entry.js
@@ -9,6 +9,7 @@ import {
   Accordion,
   makeAccordionHeaderButton,
 } from "@fontra/web-components/ui-accordion.js";
+import { Form } from "@fontra/web-components/ui-form.js";
 import Panel from "./panel.js";
 
 // These features are on by default. See hb-ot-shape.cc
@@ -179,6 +180,7 @@ export default class TextEntryPanel extends Panel {
 
     this.setupTextEntryElement();
     this.setupTextAlignElement();
+    this.setupTextSizeElement();
     this.setupAccordionElement();
     this.setupIntersectionObserver();
   }
@@ -219,6 +221,7 @@ export default class TextEntryPanel extends Panel {
                 }),
               ]
             ),
+            html.div({ id: "text-size-menu" }),
             html.div({ id: "text-settings-accordion" }),
           ]
         ),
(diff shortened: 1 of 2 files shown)


justvanrossum:
One idea is to make it a single line:

```
   Text size  [ ...input... ]  [ points / pixels popup ]
```
(The popup value would be an app level preference)

Or maybe forget about pixels, and only offer points.

qwerasd205:
How does this look? The field is in points but if you type `<n>px` it automatically converts to points for you.

https://github.com/user-attachments/assets/7e9eaa3e-7133-4f8f-aaaa-121923c3afd3

justvanrossum:
> How does this look? The field is in points but if you type `<n>px` it automatically converts to points for you.

Looks good to me!

justvanrossum (commented):
Thanks for updating this.

The UI looks good to me. I found a visual snag, and have some code feedback.

I would love `textSize` to be more ephemeral, and not be part of the `sceneSettings` object. The UI code could listen to `sceneSettings.viewBox` changes, or add another canvas event listener for "viewBoxChanged".

Perhaps the canvas controller needs a `setMagnification()` method that does the clamping.

Changing the text size with the text field seems to not make the view update according to the `onePixelUnit` value: nodes change in size, selection outlines, etc.

Updating the maginification alone centers around the canvas origin, which is rather arbitrary. It should probable take the center of the current viewBox into account.

I feel that the `Form` class is a bit too heavy for this single item. There's `labeledTextInput()`, but that requires an ObservableObject (that could be a special object, or `sceneSettings` after all, hmm). Maybe just a plain label + text input?

qwerasd205:
I'm going to close this for now since I'm focusing on making typescript conversions and this is bound to get stale. Once all the relevant files for this change are typescript-ified I'll probably pick this back up.

PR #2579, opened 2026-05-05
Add a whole bunch of JSDoc type annotations

qwerasd205:
I just sorta walked around the parts of the codebase I've been working with and checked whether my editor was resolving the types on important things (areas with a lot of indirection mainly), and whenever it wasn't resolving them I added annotations until it could.

Some annotations are incomplete- with fields typed as `any` instead of a more specific type, because I didn't want to go down the rabbit hole for that type.

These annotations do make the touched sections of the code a lot nicer to work in in editors with appropriate tooling, since going to definitions is made easier and completions for properties and methods ... exist.

If you have any questions or concerns feel free to ask.

### DIFF
--- a/src-js/fontra-core/src/canvas-controller.js
+++ b/src-js/fontra-core/src/canvas-controller.js
@@ -1,21 +1,36 @@
 import { normalizeRect, rectCenter, validateRect } from "./rectangle.js";
 import { assert, clamp, consolidateCalls, isNumber, withSavedState } from "./utils.js";
 
+/** @import { Rectangle } from "./rectangle.js" */
+/** @import { SceneView } from "./scene-view.js" */
+
 const DEFAULT_MIN_MAGNIFICATION = 0.005;
 const DEFAULT_MAX_MAGNIFICATION = 200;
 
 export class CanvasController {
+  /**
+   * @param {HTMLCanvasElement} canvas
+   * @param {undefined | (number) => void} magnificationChangedCallback
+   */
   constructor(canvas, magnificationChangedCallback) {
+    /** @type {HTMLCanvasElement} */
     this.canvas = canvas; // The HTML5 Canvas object
+    /** @type {CanvasRenderingContext2D} */
     this.context = canvas.getContext("2d");
+    /** @type {SceneView} */
     this.sceneView = undefined; // will be set later
 
+    /** @type {number} */
     this.magnification = 1;
+    /** @type {{ x: number, y: number }} */
     this.origin = { x: this.canvasWidth / 2, y: 0.85 * this.canvasHeight }; // TODO choose y based on initial canvas height
 
+    /** @type {number} */
     this._minMagnification = DEFAULT_MIN_MAGNIFICATION;
+    /** @type {number} */
     this._maxMagnification = DEFAULT_MAX_MAGNIFICATION;
 
+    /** @type {undefined | (number) => void} */
     this._magnificationChangedCallback = magnificationChangedCallback;
 
     const resizeObserver = new ResizeObserver((entries) => {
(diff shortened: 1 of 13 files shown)


justvanrossum:
Thank you. Seeing this makes me think JSDoc annotations are a stop-gap and we would benefit much more from (gradual) TS conversion. At least we'd get some type checking with that, whereas the JSDoc annotations can easily turn stale.

PR #2309 promised to convert an initial small set of core JS files to TS, but then accidentally dragged in a ton of other stuff, and the author didn't want to or couldn't persue this any further. But I'm still interested in that approach, and perhaps we can clean up some of that work.

qwerasd205:
If you'd like, I could make a small PR that mostly is just to get a foot in the door for the TS conversion (adding the relevant build dependencies, tsconfig file, etc.), and which converts like one file (perhaps `utils.js`) to give a taste of what the js->ts conversion will look like. Then open up additional PRs afterwards with conversion of more complicated files.

Out of scope for the typescript conversion but one thing is that the current inter-file dependency structure is very un-rigorous. A style that I like because it generally avoids the profusion of `import` statements is to have a main file for each module that re-exports everything that needs to be accessed from elsewhere- then instead of, e.g.
```js
import { foo } from "@fontra/core/a.js";
import { bar, baz } from "@fontra/core/b.js";
import { quux } from "@fontra/core/c.js";
```
it becomes the much cleaner
```js
import { foo, bar, baz, quux } from "@fontra/core";
```
and this also avoids js->ts changes causing a bunch of changes in files that consume it since you no longer need to refer to them directly by extension- only the main file that re-exports needs to be changed.

TBH, would you prefer a PR that cleans up the dependency structure first, before the TS conversion PR, so that when the TS conversion is done it's easier to review since it wouldn't cause a ton of files to change just for the extension change in the import statements?

justvanrossum:
> perhaps utils.js

And/or rectangle.js / transform.js?

> TBH, would you prefer a PR that cleans up the dependency structure

Hmm, `@fontra/core` is pretty large and diverse, and I'm curious and worried about the consequences. Could you whip up some small examples of what this would entail? That might help me wrap my mind around this, and possibly convince me that it's an improvement worth persuing.

justvanrossum:
Btw I’m not afraid of many import changes: those are relatively easy to review.

But I’m so far not warming to the idea of bundling all `@fontra/core` exports in a single namespace. I can see that more easily for `@fontra/web-components`.

qwerasd205:
I started working on it, but you're right that `@fontra/core` is very large and diverse and it really doesn't work well with a single main file, so I've convinced myself it's not the right direction. Also JavaScript's import syntax doesn't allow for deep destructuring, so even re-exporting namespaces for things doesn't work well on the consumer side. I feel like things could do with better namespacing but I guess it's fine enough as it is - I'm definitely not familiar enough with the whole codebase to be the one to do a large refactor to reorganize the namespaces, so I'll let that idea be.

For now I'll get started on a first pass converting a handful of small files to TypeScript then. (and if you'd prefer that to these JSDoc annotations then feel free to close this PR)

justvanrossum:
> I feel like things could do with better namespacing

You are absolutely right. `@fontra/core` being "very large and diverse" is a very kind way of putting it. I'll try to think over time how it could perhaps be split up into more logical groups.

> For now I'll get started on a first pass converting a handful of small files to TypeScript then

Thank you, that would be great.

> (and if you'd prefer that to these JSDoc annotations then feel free to close this PR)

I'm afraid I do. JSDoc annotations feel quite fine-grained, but without a mechanism to verify them against the actual code. TS is the way forward.

Thank you so much for getting this conversation moving again, and for your ongoing efforts.

justvanrossum:
But if you think it’s useful to have these JSDoc annotations in the meantime, we can also decide to merge.

qwerasd205:
I've made #2585 that begins the typescript conversion process. I'll close this PR now.

PR #2580, opened 2026-05-06
Add blog and Discord links to fontra-menus

davelab6:
To match prominent links on the homepage

Possibly this needs 1+ more commits that add the localizations of the menus


justvanrossum:
Thanks! I'll patch this up with formatting and translation logic. I'll take your branch, but it may have to become a superseding PR.

justvanrossum:
"Discord" doesn't need translation (like "GitHub"), and I'm torn about "Blog": I'll leave it untranslated, at least for now.

(I managed to fix the syntax in your branch, so no need for a new PR)

PR #2585, opened 2026-05-07
Begin typescript conversion

qwerasd205:
Adds a `tsconfig.json` so that `tsc` in the root directory will type check all `.ts` files inside `src-js`. Enabling checking of js via tsc (`"checkJs": true`) yields a ton of errors, most of which I'm sure are spurious and just present because it isn't sure about the types of everything.

Once a lot more stuff is converted to typescript it may make sense to enable js checking- and it might make sense to enable it and run `tsc` just to get a list of files that might be good candidates for conversion - but aside from that it's not really a problem that it's not enabled right now, since the first and foremost benefit of this change in my opinion is the improved interaction with tooling (editor completions mainly).

As a start, I've converted `rectangle.js` and `utils.js` to typescript. They're mostly completely typed, though some of the things that take glyph objects (or arrays thereof) in `utils.js` have them typed as `any` with a `TODO` comment about properly typing them in the future.

I've tried to separate this in to individually reviewable commits so the noise from changing all of the imports doesn't get in the way of the main changes.

If you have any questions, let me know.

### DIFF
--- a/src-js/fontra-core/src/utils.ts
+++ b/src-js/fontra-core/src/utils.ts
@@ -3,8 +3,10 @@ import type { Point } from "./rectangle.ts";
 import { Transform } from "./transform.js";
 import { addItemwise } from "./var-funcs.js";
 
+/**
+ * Shallow object compare. Arguments may be null or undefined
+ */
 export function objectsEqual(obj1: any, obj2: any) {
-  // Shallow object compare. Arguments may be null or undefined
   if (!obj1 || !obj2) {
     return obj1 === obj2;
   }
@@ -29,15 +31,19 @@ export function withSavedState(context: CanvasRenderingContext2D, func: () => vo
   }
 }
 
+/**
+ * Return a function that will request `func` to be called in the next
+ * iteration of the event loop. If it gets called again before `func` was
+ * actually called, ignore the call.
+ *
+ * This ensures that multiple calls within the same event loop cycle get
+ * consolidated into a single call.
+ *
+ * Useful for things like "request update".
+ */
 export function consolidateCalls<Fn extends (...args: any[]) => void>(
   func: Fn
 ): (...args: Parameters<Fn>) => void {
-  // Return a function that will request `func` to be called in the next
-  // iteration of the event loop. If it gets called again before `func` was
-  // actually called, ignore the call.
-  // This ensures that multiple calls within the same event loop cycle get
-  // consolidated into a single call.
-  // Useful for things like "request update".
   let didSchedule = false;
 
   return (...args) => {
@@ -54,15 +60,19 @@ export function consolidateCalls<Fn extends (...args: any[]) => void>(
 
 export type TimeoutID = ReturnType<typeof setTimeout>;
 
+/**
+ * Schedule calls to func with a timer. If a previously scheduled call
+ * has not yet run, cancel it and let the new one override it.
+ *
+ * Returns a wrapped function that should be called instead of func.
+ *
+ * This is useful for calls triggered by events that can supersede
+ * previous calls; it avoids scheduling many redundant tasks.
+ */
 export function scheduleCalls<Fn extends (...args: any[]) => void>(
   func: Fn,
   timeout = 0
 ): (...args: Parameters<Fn>) => TimeoutID {
-  // Schedule calls to func with a timer. If a previously scheduled call
-  // has not yet run, cancel it and let the new one override it.
-  // Returns a wrapped function that should be called instead of func.
-  // This is useful for calls triggered by events that can supersede
-  // previous calls; it avoids scheduling many redundant tasks.
   let timeoutID: TimeoutID | null = null;
   return (...args) => {
     if (timeoutID !== null) {
@@ -76,13 +86,15 @@ export function scheduleCalls<Fn extends (...args: any[]) => void>(
   };
 }
 
+/**
+ * Return a wrapped function. If the function gets called before
+ * minTime (in ms) has elapsed since the last call, don't call
+ * the function.
+ */
 export function throttleCalls<Fn extends (...args: any[]) => void>(
   func: Fn,
   minTime = 0
 ): (...args: Parameters<Fn>) => TimeoutID | null {
-  // Return a wrapped function. If the function gets called before
-  // minTime (in ms) has elapsed since the last call, don't call
-  // the function.
   let lastTime = 0;
   let timeoutID: TimeoutID | null = null;
   return (...args) => {
@@ -156,19 +168,26 @@ export const arrowKeyDeltas = {
   ArrowRight: [1, 0],
 };
 
+/**
+ * Modulo with Python behavior for negative values of `v`
+ *
+ * Assumes `n` to be positive
+ */
 export function modulo(v: number, n: number) {
-  // Modulo with Python behavior for negative values of `v`
-  // Assumes `n` to be positive
   return v >= 0 ? v % n : ((v % n) + n) % n;
 }
 
+/**
+ * Return 1 if `v` is true-y, 0 if `v` is false-y
+ */
 export function boolInt(v: boolean) {
-  // Return 1 if `v` is true-y, 0 if `v` is false-y
   return v ? 1 : 0;
 }
 
+/**
+ * Like Python's reversed(seq) builtin
+ */
 export function* reversed<T>(seq: T[]) {
-  // Like Python's reversed(seq) builtin
   for (let i = seq.length - 1; i >= 0; i--) {
     yield seq[i];
   }
@@ -204,8 +223,10 @@ export function* range(start: number, stop?: number, step = 1) {
   }
 }
 
+/**
+ * After Python's itertools.chain()
+ */
 export function* chain<T>(...iterables: Iterable<T>[]) {
-  // After Python's itertools.chain()
   for (const iterable of iterables) {
     for (const item of iterable) {
       yield item;
@@ -213,9 +234,12 @@ export function* chain<T>(...iterables: Iterable<T>[]) {
   }
 }
 
+/**
+ * Cartesian product of input iterables. Equivalent to nested for-loops.
+ *
+ * After Python's itertools.product()
+ */
 export function* product<T>(...args: Iterable<T>[]): Generator<T[], void, unknown> {
-  // Cartesian product of input iterables.  Equivalent to nested for-loops.
-  // After Python's itertools.product()
   if (!args.length) {
     yield [];
     return;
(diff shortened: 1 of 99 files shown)


src-js/fontra-core/src/utils.ts
```diff
@@ -221,23 +260,25 @@ export function* product(...args) {
   }
 }
 
-export function compare(a, b) {
-  // Return -1 when a < b, 1 when a > b, and 0 when a == b
-  return (a > b) - (a < b);
+/**
+ * Return -1 when a < b, 1 when a > b, and 0 when a == b
+ */
+export function compare(a: number, b: number): -1 | 0 | 1 {
```
justvanrossum: This function is also used to compare strings (for example by `ensureGlyphCompatibility()` in `glyph-controller.js`.

src-js/fontra-core/src/utils.ts
```diff
@@ -246,7 +287,9 @@ export function parseSelection(selection) {
   return result;
 }
 
-export function makeUPlusStringFromCodePoint(codePoint) {
+export type Falsey = null | undefined | false | 0 | -0 | 0n | "";
+
+export function makeUPlusStringFromCodePoint(codePoint: number | Falsey) {
```
justvanrossum: The intention is to accept a number or `undefined`, rather than the more broad "falsey". Scanning through confirms that. Maybe the explicit type check should also reflect that (it can probably be removed once all modules using this function are converted to .ts?).

src-js/fontra-core/src/utils.ts
```diff
@@ -284,19 +329,22 @@ export async function readFromClipboard(types, plainText = true) {
   return undefined;
 }
 
-export function getCharFromCodePoint(codePoint) {
-  return codePoint != undefined ? String.fromCodePoint(codePoint) : "";
+export function getCharFromCodePoint(codePoint: number | undefined) {
```
justvanrossum, line 342: General question: should the return type be declared here? Or is it clear to TS that it's a string by looking at the code?

src-js/fontra-core/src/utils.ts
```diff
@@ -562,7 +630,8 @@ export function writeObjectToURLFragment(obj, replace = false) {
   }
 }
 
-export function areGuidelinesCompatible(parents) {
+// TODO: proper typing for glyphs
```
justvanrossum: This function is used both for arrays of glyph objects as well as for arrays of font sources. The only requirement is that the array elements have a "guidelines" property, which is an array of guidelines, but in this case "an object with a `name` property" would be good enough. Is it possible to express this in TS, or would it be better to later do a union betwee glyph and font layer, once they are typed?

src-js/fontra-core/src/utils.ts
```diff
@@ -615,9 +685,17 @@ export function areCustomDatasCompatible(parents) {
   return true;
 }
 
+export type Guideline = {
```
justvanrossum, line 710: Ideally we'll end up with a more central place for these core type defs. Note there's a Python equivent for this `classes.py`. 

See for example the [`Guideline`](https://github.com/fontra/fontra/blob/b07d7fc74d0e5a6838a979aef11359ba6b298cd4/src/fontra/core/classes.py#L145-L152) class. The `classes.py` module describes the core types for Fontra font data. Ideally we'd think of a way to derive the TS classes from `classes.py`, or find a way to check that they are the same. 

There's also `classes.json`, which is derived from `classes.py`, but misses some info, such as default values.

src-js/fontra-core/src/utils.ts
```diff
@@ -866,33 +981,36 @@ export const friendlyHttpStatus = {
   505: "HTTP Version Not Supported",
 };
 
-export function stringCompare(a, b) {
+export function stringCompare(a: string, b: string) {
```
justvanrossum: Oh ah, haha, maybe this one should be deleted, *or* we should use it consistently instead of `compare()` if we're comparing strings. Doesn't need to be this PR, but maybe we should collect a list of "little things that need fixing".

justvanrossum (commented):
Wow, this is pretty great, thank you so much!

I left some comments and questions.

src-js/fontra-core/src/utils.ts
qwerasd205: In cases where I haven't added an explicit return type it's because typescript is able to infer it correctly from the function body.

src-js/fontra-core/src/utils.ts
qwerasd205: Agreed

src-js/fontra-core/src/utils.ts
qwerasd205: Yeah I noticed a handful of odd things that would fit on that list, like a lot of functions with names that I feel don't accurately reflect their function. Like `isNumber` for example should really be `isFinite` since that's what it checks, is that a given value is a finite number.

src-js/fontra-core/src/utils.ts
qwerasd205: That is certainly possible to express in typescript, I'll update it to do so. (Perhaps later it could be updated to be a union between glyph and font layer though, just for correctness)

src-js/fontra-core/src/utils.ts
qwerasd205: Alright, I'll update the type and the check.

Theoretically, runtime type checks only need to be performed at the interface between ts and js code, yeah, so once no js code is directly calling this function the runtime type check can be removed from it.

src-js/fontra-core/src/utils.ts
qwerasd205: I can update this so that it takes `string | number`, but make it generic so that the types of both arguments have to match. (And replace all uses of `stringCompare` with this.)

qwerasd205:
Added 4 commits that address much of the feedback

src-js/fontra-core/src/utils.ts
qwerasd205: I've done this and added a note to the doc string about using a more appropriate string comparison where applicable. I looked at how `compare` and `stringCompare` were being used and I think a lot of them should be replaced with a more proper comparison function, but I didn't want to dirty this PR by making those changes as a part of it.

src-js/fontra-core/src/utils.ts
qwerasd205: I added a `TODO` comment noting that this definition is temporary and should be replaced with types derived from the python classes.

qwerasd205:
Oops, forgot to update the tests for `makeUPlusStringFromCodePoint` after making it stricter about what types it takes, will fix

justvanrossum (approved):

qwerasd205:
Ah, I've just realized that in order to natively run typescript files without additional flags, node.js needs to be version 22.18.0 or later, though we may want to add the `--experimental-transform-types` to the various run commands in case we start using non-erasable TypeScript syntax.

Regardless, the requirements section of the README needs to be updated and we should add an ["engines"](https://docs.npmjs.com/cli/v10/configuring-npm/package-json?v=true#engines) field to the `package.json` to enforce the minimum node version.

Ref: https://nodejs.org/learn/typescript/run-natively

justvanrossum:
The CI uses nodejs 24, I’ll put >= 24 in the readme. _(edit: done)_

justvanrossum:
I wonder what it would take to ensure the tests run properly transpiled. I would assume that transpiling and "stripping types" (by node) may behave in subtly different ways.

qwerasd205:
Most TypeScript syntax is "erasable", such that transpiling is the same as stripping. We can enable the [`erasableSyntaxOnly`](https://www.typescriptlang.org/tsconfig/#erasableSyntaxOnly) option in the tsconfig to ensure that we don't accidentally use any non-erasable syntax.

justvanrossum:
Another question: if I on purpose change a type in rectangle.ts to a type that doesn't exist, or use a non-existing Rect property, I do not get an error from the bundler. I would have assumed that transpiling would fail in those cases?

justvanrossum:
> Most TypeScript syntax is "erasable", such that transpiling is the same as stripping. We can enable the [erasableSyntaxOnly](https://www.typescriptlang.org/tsconfig/#erasableSyntaxOnly) option in the tsconfig to ensure that we don't accidentally use any non-erasable syntax.

But how to run the tests with actual transpiled code?

qwerasd205:
> Another question: if I on purpose change a type in rectangle.ts to a type that doesn't exist, or use a non-existing Rect property, I do not get an error from the bundler. I would assume that transpiling would fail in those cases?

I don't know if webpack is set up to run type checking while bundling. You can get type checking by running `tsc` in the root directory of the project. (should probably add that to the `test` command)

> But how to run the tests with actual transpiled code?

If we want to use something other than erasable syntax then we'd have to set up the tests to transpile first and then run, or use node's "experimental transform types" flag which tries to transform non-erasable syntax (basically it tries to do what the typescript transpiler does for those cases).

justvanrossum:
> I don't know if webpack is set up to run type checking while bundling. You can get type checking by running tsc in the root directory of the project. (should probably add that to the test command)

I think this should run as part of pre-commit. Either way the CI must do type checking.

I ran `tsc` after I did `npm install -g typescript`, and am getting this error:
```
$ tsc
error TS2688: Cannot find type definition file for '@types/emscripten'.
  The file is in the program because:
    Entry point of type library '@types/emscripten' specified in compilerOptions

  tsconfig.json:14:15
    14     "types": ["@types/emscripten"]
                     ~~~~~~~~~~~~~~~~~~~
    File is entry point of type library specified here.


Found 1 error.
```

I found this pre-commit hook: https://github.com/mohsentaleb/precommit-tsc, but running that gives a ton of errors if I add that to `.pre-commit-config.yaml`:

```
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index bd578b64c..34fb3dbaf 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,3 +28,8 @@ repos:
       - id: trailing-whitespace
       - id: end-of-file-fixer
       - id: check-yaml
+
+  - repo: https://github.com/mohsentaleb/precommit-tsc
+    rev: "v1.0.0"
+    hooks:
+      - id: tsc
```

Summary: "Found 842 errors.", hmmm.

justvanrossum:
> I don't know if webpack is set up to run type checking while bundling.

But it does transpile, right? I would assume transpiling implies type checking.

justvanrossum:
Maybe I merged too soon: we really must figure out how to do type checking as part of pre-commit and/or bundling.

justvanrossum:
Or even as an explicit CI step, like we do with mypy.

qwerasd205:
> > I don't know if webpack is set up to run type checking while bundling.
> 
> But it does transpile, right? I would assume transpiling implies type checking.

It doesn't imply it- most bundlers don't check since that takes much longer than just stripping erasable syntax and transforming everything else.

qwerasd205:
The error occurred because you don't have all the dev dependencies installed (in this case `@types/emscripten` is missing). (run `npm install` to make sure everything is installed)

qwerasd205:
That hook is failing because it's running `tsc` on each file individually, not using the `tsconfig.json`. Ideally the hook should just invoke `tsc`, which returns a non-zero exit code when there are problems and nothing otherwise.

qwerasd205:
#2586

PR #2590, opened 2026-05-10
Convert observable-object.js (and its tests) to typescript

qwerasd205:
This already improves the dev experience around observable objects, and once the models those objects are wrapped around are properly typed it will be even better.

Just as an example of the gains already though: typescript is able to figure out from the js code that the type of this field on the model is a boolean, so because of these typings it knows that the `event.newValue` in this key listener is also a boolean (shown here with the language server hover hint).
image

And when used from a typescript file, it will be a statically checked type error to pass a non-existing field (that is, one which isn't specified on the model's type) as a key to any of those functions like `addKeyListener`.


justvanrossum (approved):
Thanks! While I'm not 100% up to speed with advanced TS features like "extends keyof", I think I do understand the intention.

I read the `ts-ignore` comments, and I think I understand.

qwerasd205:
> Thanks! While I'm not 100% up to speed with advanced TS features like "extends keyof", I think I do understand the intention.

When a generic type parameter (as in `whatever<T>`) is written, it's possible to constrain that parameter to only types that are compatible with a provided type by using the `extends` keyword, so for example `whatever<T extends U>` requires that whatever `T` is (either implicitly or explicitly) it must be compatible with (it must "extend") `U`.

`keyof T` produces a type that is a union of all keys of `T`, so if `T` is `{ foo: number; bar: number; }` then `keyof T` is `"foo" | "bar"`.

So `whatever<K extends keyof T>` effectively means "`whatever` will use a type `K` which must be compatible with the types of the keys of `T`", basically "`K` is a key of `T`".

justvanrossum:
Thanks for the explanation, that's really helpful.

PR #2595, opened 2026-05-15
Update prettier to v3.8.3

### LINKED ISSUE #2592: Update `prettier` (currently blocking typescript migration)
The current version of `prettier` used in the repo is `3.0.3`. This hasn't been an issue previously, but it's getting in the way of the migration from javascript to typescript, since the outdated version is lacking support for some modern syntax in TypeScript-- specifically import attributes, which is blocking the migration of `fontra-core/classes.js` since it imports a json file `with { type: "json" }`. Support for import attributes in TypeScript came in prettier version `3.1.1` ([changelog](https://github.com/prettier/prettier/blob/main/CHANGELOG.md#311)).

I made a first attempt at updating the prettier version to the latest; but unfortunately, changes to the `prettier` CLI broke the pre-commit hook we use (https://github.com/pre-commit/mirrors-prettier), and the repo for the hook has been archived and hasn't been updated past version `3.1.0`.

The stale state of the official pre-commit hook for prettier is discussed here: https://github.com/prettier/prettier/discussions/16227, and a couple people have working up-to-date forks, but since they're not from a primary source (either `pre-commit` or `prettier`) I don't know whether we would want to trust them. I'm inclined to say that the best course of action is to vendor the hook and make the fixes ourselves.

If you give me the go-ahead, I'll make a branch with a vendored copy of the hook, update prettier to the latest version (`3.8.3`) and make a PR.

### EDIT:
I have a branch with the update, I'll PR it if it looks good to you. Note that the updated version of prettier has a few formatting choice differences, which causes quite a few js files to change (almost all due to changed rules around ternary indentation) if you run prettier directly (`npx prettier . --write`), so we'd probably want to do that so that unrelated changes in a file don't cause a commit to suddenly have a format change in a totally different part. Should that (formatting all the files that are affected by the changes to prettier) be a part of the PR too?

Diff for my branch: https://github.com/fontra/fontra/compare/main...qwerasd205:fontra:update-prettier

qwerasd205:
Closes #2592

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,12 +15,16 @@ repos:
     hooks:
       - id: flake8
 
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v3.0.3
+  - repo: local
     hooks:
       - id: prettier
+        name: prettier
         exclude: "src/fontra/client/third-party"
         types_or: [css, javascript, json, xml, yaml, markdown, html, ts]
+        entry: prettier --write --list-different --ignore-unknown
+        language: node
+        require_serial: false
+        additional_dependencies: ["prettier@3.8.3"]
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v6.0.0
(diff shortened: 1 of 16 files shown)


justvanrossum (approved):
Great, thank you!!

PR #2660, opened 2026-06-22
Keep forward browser history when going back

justvanrossum:
Browser forward history was sometimes lost when using the browser back button. This PR tries harder to not push unneeded history state.

### DIFF
--- a/src-js/fontra-core/src/observable-object.ts
+++ b/src-js/fontra-core/src/observable-object.ts
@@ -7,6 +7,7 @@ export type Event<T, K extends keyof T> = {
   newValue: T[K];
   oldValue: T[K];
   senderInfo: any;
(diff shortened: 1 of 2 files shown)


qwerasd205:
This seems to makes zooming and panning the canvas not update the URL

justvanrossum:
Oh oops, I’ll investigate.

justvanrossum:
> This seems to makes zooming and panning the canvas not update the URL

Are you sure? I don't see that happening.

qwerasd205:
Weird. Must be some kind of transient bug where sometimes it's possible for the editor controller to get stuck on top of the sender stack. I can't reliably reproduce it, so it's possible it was due to some changes I had made while testing stuff.

PR #2663, opened 2026-06-24
js: switch from History API to Location API

### LINKED ISSUE #2654: Change URL behavior
Currently, view info is put in to the URL hash, compressed and base64 encoded. I assume this is to allow a user to, for example, restart their browser and pick back up from where they left off.

The relevant code is the call to `writeObjectToURLFragment` in `editor.js` (also used in `fontoverview.js`).

There are three big issues with this approach:

1. This is limited. The potential length of the encoded data is uncapped since it includes the text in the text entry panel, but browsers have various limits on URL length for different purposes. Safari, for example, cuts off URLs in the navbar at 4096 chars, which means the data will get truncated and corrupted.
2. This floods browser history. Every time I'm done with an editing session in Fontra, I have at least a few hundred different versions of the URL in my browser history.
3. This causes Safari to hitch, lag, generally experience performance issues, and even occasionally crash! There are several problems that occur in Safari. The first is that it seems like the entire context is cloned every time `history.pushState` is used, which causes large GCs as the objects from the old contexts need to be cleared. The second is that after inserting hundreds of URLs in to the history, Safari may fail to insert an entry - specifically may have trouble with inserting in to the favicon cache (based on the stacktrace I get) - and crash.

The solution I propose is to have a unique ID generated for a given session, put *that* in the URL hash, and store the relevant data in `localStorage` or `indexedDB` with, perhaps, some LRU and/or timeout based eviction logic.

If some people really have a utility for this behavior of encoding every zoom, pan, and word typed in the text entry field as a separate history entry, perhaps it could be made a config option... but to me it feels pretty useless.

The lag and hitching in Safari can be really bothersome, so if you give me the go-ahead to implement my alternative view state persistence logic, I'll get right on it.

qwerasd205:
Fixes #2654.

## Summary of Changes
This PR replaces the use of `window.history.replaceState` and `window.history.pushState` with the very similar `window.location.replace` and `window.location.assign`. The key difference between these two APIs is that the `history` methods allow associating an arbitrary object with each history entry, while the `location` methods do not; but we were passing an empty object for that anyway so that doesn't matter.

The motivation for this change is rather subtle, and I explain in more detail below, but ultimately it seems to have resolved a problem of hitching, lag, and freezing that occurred in long-running editing sessions in Safari.

However, another difference in behavior is that, while the `history` methods don't trigger the `popstate` event, the `location` methods *do*. We rely on the `popstate` event to restore view parameters as the user navigates their history with the browser's forward/back buttons or other UI. We don't want to run that logic, however, if we are the ones responsible for the URL hash (AKA fragment) changing, since it's at best redundant and at worst would cause all sorts of weird behaviors. In order to fix this, I added a flag that is set before changing the location in the `writeObjectToURLFragment` function and is cleared after any event listeners have fired. We gate our response to the `popstate` event on that flag to avoid any problems.

## Explanation
This was a painful problem to diagnose, but I'm *fairly* confident I diagnosed it correctly, since the fix (which was mercifully simple) did seem to resolve the lag and hitching I had experienced before.

Ultimately, the problem lies in Safari's "reader mode". It runs a script on the page to identify which, if any, elements represent the "article" content on the page. It does this when a page is first loaded, and, critically, in response to the `Navigation` API's `navigate` event ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/Navigation/navigate_event)).

This means that every time the URL changes, it tries to forget everything it knew about the page and then re-run its search logic. This would be fine, if slightly wasteful, but shouldn't create increasing amounts of lag. However, I believe what happens is that the external component that injects this script is injecting another copy of the script (which runs in its own execution context, it would seem, likely to try to prevent prototype pollution). This happens *every single time* that `replaceState` is called, which is up to 5 times a second. And every time a `navigate` event happens, every copy will `delete` its index of the elements on the page and make a new one... and all of those are cleaned up in the same garbage collection pass as the page's main thread uses. This means we end up with monstrous GCs that take hundreds of milliseconds, sometimes even thousands.

Now, the `navigate` event is triggered by every sort of navigation, including the `location` methods, but looking at the exact logic (thanks to the code that someone helpfully extracted from Safari), you can see that the scripts gates its response on the event's `hashChange` property:

https://github.com/NohamR/safari-internal-js/blob/8dc9429f8414533792e2fe34d1762a47369c852e/26.2/js/reader/ReaderArticleFinder.js#L5085-L5092

Despite only changing the hash, our call to `replaceState` (or `pushState`) does not result in an event with that property, which is by design, since it also - per spec - doesn't trigger the `hashchange` event. By using the `location` methods instead, the `hashChange` property will be set correctly on the `navigate` event, and the problem goes away.

I haven't checked if the change to the `location` methods also prevents the reader mode script from being duplicated in the first place, but I wouldn't be surprised if it does. Even if it doesn't, it prevents massive garbage collections and the editor will be as smooth after 6 hours of work as it is after 5 minutes, which was not the case before.

For completeness sake- I noted the existence of this problem (and tested the fix) in Safari version 26.5.

---

Ultimately, this shouldn't have been Fontra's problem to solve, it's a failure on the part of Safari, and quite a nasty bug at that. One can only hope that Apple will get their act together and fix it up at some point in the future.

> [!NOTE]
> You may want to test these changes in all the browsers just to make sure all the behavior seems correct to you. I did tests on my end, but you never know.

### DIFF
--- a/src-js/fontra-core/src/utils.ts
+++ b/src-js/fontra-core/src/utils.ts
@@ -624,6 +624,16 @@ export function readObjectFromURLFragment() {
   return url.hash ? loadURLFragment(url.hash) : {};
 }
 
+let _writingURLFragment = false;
+/**
+ * Returns true in an event listener for `popstate` or `navigate` if the event
+ * was triggered by a call to `writeObjectToURLFragment`, and not by the user
+ * using the browser's navigation (forward and back buttons).
+ */
+export function eventIsCausedByWritingURLFragment() {
+  return _writingURLFragment;
+}
+
 export function writeObjectToURLFragment(obj: any, replace = false) {
   const newFragment = dumpURLFragment(obj);
   // @ts-ignore Typescript complains that window.location is missing some
@@ -633,10 +643,23 @@ export function writeObjectToURLFragment(obj: any, replace = false) {
     return;
   }
   url.hash = newFragment;
+
+  // See explanation on `eventIsCausedByWritingURLFragment`.
+  _writingURLFragment = true;
+  // We reset the value in a freshly queued `hashchange` listener which is
+  // guaranteed to fire after any other existing listener since it's new.
+  window.addEventListener(
+    "hashchange",
+    () => {
+      _writingURLFragment = false;
+    },
+    { once: true }
+  );
+
   if (replace) {
-    window.history.replaceState({}, "", url);
+    window.location.replace(url);
   } else {
-    window.history.pushState({}, "", url);
+    window.location.assign(url);
   }
 }
 
--- a/src-js/views-editor/src/editor.js
+++ b/src-js/views-editor/src/editor.js
@@ -37,6 +37,7 @@ import { labeledCheckbox, labeledTextInput, pickFile } from "@fontra/core/ui-uti
 import {
   commandKeyProperty,
   enumerate,
+  eventIsCausedByWritingURLFragment,
   fetchJSON,
   hyphenatedToCamelCase,
   hyphenatedToLabel,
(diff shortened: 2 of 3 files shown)


justvanrossum:
Amazing, thank you! I have not tested it yet, and will come back to you after I did. On a glance it looks good to me, but I'll do a more thorough review as well.

For completeness sake, can you edit the PR description to contain the version of Safari that you are observing this with?

I wonder if a summary of the Safari issue should be included as a comment in the code.

justvanrossum (approved):
I've played with this in a few browsers, and it seems to work fine. I'll merge after #2665.

PR #2681, opened 2026-07-01
[font overview] Add "Copy Glyph Name(s) and "Copy Character(s)" context menus

### LINKED ISSUE #2676: [UX] Allow for copying character in font overview cards
Would be pretty handy to be able to copy either the glyph name (in this case `Endescender-cy`) or the character (in this case Ң)

Image

justvanrossum:
This fixes #2676.

Additionally, this changes the behavior of "Select All" to really select all, and not skip glyphs that aren't in the font (but are in the combined selected glyph sets).


Manueljlin:
thanks :)

PR #2687, opened 2026-07-06
Fix zoom in/out behavior when there is no selection

### LINKED ISSUE #2686: Zoom actions in top toolbar don't keep the center of the viewport
https://github.com/user-attachments/assets/5ae92f03-1d2e-4364-8d28-58111bfb21ca

# expected behavior
When zooming in from toolbar buttons, it should zoom relative to center of the viewport (so the visible part of the canvas)

# current behavior
Right now, it seems to center relative to the canvas center, not the viewport center. Because that includes the occluded parts, the more lines you have the further both centers drift apart.

justvanrossum:
Fix zoom in/out behavior when there is no selection at all: use the view box center in that case, instead of the scene box center

This fixes #2686.

### DIFF
--- a/src-js/views-editor/src/editor.js
+++ b/src-js/views-editor/src/editor.js
@@ -3425,7 +3425,7 @@ export class EditorController extends ViewController {
 
   _zoom(factor) {
     let viewBox = this.sceneSettings.viewBox;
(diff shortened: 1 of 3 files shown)


Manueljlin:
thanks!

